PR #4, opened 2021-02-05
Add proto messages representing all account balances

tinker-michaelj:
- Addresses [#852](https://github.com/hashgraph/hedera-services/issues/852) in `hashgraph/hedera-services`


noshmody:
@ljianghedera should the protos for the account balances file be in its own proto file?

ljianghedera:
I would think the CryptoGetAccountBalance is an appropriate location for these proto messages as they are all associated the account balances queries.

PR #6, opened 2021-02-09
Separate messages used for balances proto under hedera/streams

tinker-michaelj:


services/CryptoGetAccountBalance.proto
```diff
@@ -62,15 +62,3 @@ message CryptoGetAccountBalanceResponse {
     uint64 balance = 3; // The current balance, in tinybars.
     repeated TokenBalance tokenBalances = 4; // The token balances possessed by the target account.
 }
-
```
steven-sheehy, line 65: Should we move TokenBalance and TokenBalances to BasicTypes so stream file doesn't have to depend upon CryptoGetAccountBalance.proto? I don't think this changes the generated classes, but not sure about non-java stuff.

steven-sheehy (changes requested):

services/CryptoGetAccountBalance.proto
tinker-michaelj: Hmm yes you're right, I compiled both ways and it doesn't seem to matter for Java at least.

Moved them, can see if anyone complains.

steven-sheehy (approved):
LGTM

PR #8, opened 2021-02-10
Include decimals in queries of token/account balances and relationships

tinker-michaelj:


steven-sheehy (approved):
LGTM

ljianghedera (approved):
LGTM

PR #9, opened 2021-02-16
Clean up protobufs

danielakhterov:
I believe consistency is extremely important when it comes to describing an API, and protobufs are exactly that. So my goal with this PR is to simply make the protobuf more consistent. I've taken a little bit of time to do this myself, but there is still quite a bit that could be done. All the changes I've made are subject to change as I don't have a strong stance for any particular change as long as the change is consistent throughout the protobufs. E.g. I renamed `*.expiry` to `*.expiration_time`, but if we think `*.expiry` makes more sense, then we should rename *all* `expiration_time` fields to `expiry` since both `expiration_time` and `expiry` represent the same idea.
### Non-breaking changes
 - Added a linter `buf` https://github.com/bufbuild/buf which is run on CI
 - Set 80 character line limit. 
   * Before this was sort of arbitrary as some lines went one for 100+ characters, while others only went to ~80.
 - Removed trailing spaces.
 - Update all comments for message types to be multi-line comments
   * Before this was really inconsistent. Sometimes it would be a multi-line comment, sometimes a single line comment, and other times it'd be a multi-line comment without proper indentation on subsequent lines.
 - Update all comments on message fields to be single line comments.
   * Before the comments on fields would either be a single line comment to the side of the field, or multiple single lines comments on top of the field. I just decided to make all comments on fields be on top as this makes it consistent with the rest of the protobufs. Although, it does make the protobufs themselves larger.
 - Added `UNDOCUMENTED` doc comment to all messages/fields which were not documented.
   * To find the undocumented messages/fields simply search for that word. Hopefully this will help get all messages/fields documented.
 - Rename all files to from TitleCase to lower_snake_case.
   * This is because of Google's protobuf style guide https://developers.google.com/protocol-buffers/docs/style
 - Renamed all services' methods from camelCase to TitleCase
   * This is because of Google's protobuf style guide https://developers.google.com/protocol-buffers/docs/style
 - Renamed enum variants from TitleCase to SCREAMING_SNAKE_CASE
   * This is because of Google's protobuf style guide https://developers.google.com/protocol-buffers/docs/style
 - Updated indentation to always be 4 spaces
   * Before for some of the new files such as `Schedule*.proto` the indentation was 2 spaces, while the rest of the files had 4 spaces.

### Potentially breaking changes
 - Moved the copyright comment to the top of the file for all protobuf files, and removed unprintable characters <200b>, <200c>, and <200d> which were in the copyright comment.
   * I'm not sure of those unprintable characters had any meaning, but if they were important they can be re-added.
   * I'm not sure if this would be considered a breaking change.
 - Renamed protobuf package from `proto` to `services` for the `services/*.proto` files
   * This _should_ not be a breaking change, but haven't tested it myself.
 - Updated all imports to use full path from root of the repo.
   * E.g. `import "basic_types.proto"` -> `import "services/basic_types.proto"`
     - This affects `mirror/`, `services/`, and `streams/`
   * E.g. `Timestamp consensus_timestamp` -> `services.Timestamp consensus_timestamp`
     - This only affects `mirror/` and `streams/`
 - Renamed protobuf package from `proto` to `mirror` for `mirror/*.proto` files
   * This _should_ not be a breaking change, but haven't tested it myself.

### Breaking changes
 - Rename all fields from camelCase to lower_snake_case
   * This is because of Google's protobuf style guide https://developers.google.com/protocol-buffers/docs/style
   * Since some of the fields were named such as `accountID` the lower_snake_case version would be `account_i_d` which I can only assume is not what was intended. So after renaming `accountID` to `account_id` this becomes a breaking change.
 - Renamed `*.auto_renew_account` to `auto_renew_account_id`
   * This is because for nearly all other `*Id` types the field names were suffixed with `_id`, but `auto_renew_account` was not.
 - Renamed `*.treasury` to `*.treasury_account_id`
   * This is for the same reason above.
 - Renamed `*.schedule_ref` to `*.schedule_ref_id`
   * This is for the same reason above.
 - Renamed `*.token` and `*.tokens` to `*.token_id` and `*.token_ids` respectively.
   * This is for the same reason above.
 - Renamed `*.expiry` to `*.expiration_time`
   * This is for the same reason above.
 - Renamed `*.transfers` and `*.token_transfers` to `*.transfer_list` and `*.token_transfers_lists`.
   * This is because `CryptoTransferTransactionBody` and `TransactionRecord` had the same fields, but were named differently.
 - Renamed `*.contract` to `*.contract_id`
   * This is for the same reason above.

**NOTE**: Breaking changes don't mean the changes break the protobufs encoding/decoding, but rather the generated code for a particular language. For instance a field name such as `accountID` in the protobufs generates `[get|set]AccountID()` in Java, but a field name of `account_id` would generate `[get|set]AccountId()` in Java; notice the lowercase letter D at the end. By renaming `accountID` to `account_id` the generated code will be broken requiring updates. These changes would affect any code that works with the protobufs.


tinker-michaelj (commented):
Thanks for this work @danielakhterov, this looks like a big step forward! I will read through in more detail and experiment to better understand what is breaking/non-breaking as I have time and electricity/Internet over the next days.

The beginning of release 0.13.0 may be a good time to pursue these standardization efforts; will discuss with Nosh...adding Steven and Leemon as reviewers also.

_Note:_ I've only had internet power/Internet over the last 36 hours, and expect more of the same, so

donaldthibeau:
@danielakhterov can we separate this into a PR with breaking changes and a PR with the non-breaking changes?

danielakhterov:
@tinker-michaelj @donaldthibeau I removed all the breaking changes in #13

PR #13, opened 2021-03-03
[Non-Breaking] Clean up protobufs 

danielakhterov:
In response to #9 I removed all the breaking changes. To test if this PR breaks Java code gen I copied these proto files into Java SDK v2 and updated the imports accordingly. There might be some breaking changes I didn't notice because the SDK didn't use them.

## Changes
 - Added a linter `buf` https://github.com/bufbuild/buf which is run on CI
 - Set 80 character line limit. 
   * Before this was sort of arbitrary as some lines went one for 100+ characters, while others only went to ~80.
 - Removed trailing spaces.
 - Update all comments for message types to be multi-line comments
   * Before this was really inconsistent. Sometimes it would be a multi-line comment, sometimes a single line comment, and other times it'd be a multi-line comment without proper indentation on subsequent lines.
 - Update all comments on message fields to be single line comments.
   * Before the comments on fields would either be a single line comment to the side of the field, or multiple single lines comments on top of the field. I just decided to make all comments on fields be on top as this makes it consistent with the rest of the protobufs. Although, it does make the protobufs themselves larger.
 - Added `UNDOCUMENTED` doc comment to all messages/fields which were not documented.
   * To find the undocumented messages/fields simply search for that word. Hopefully this will help get all messages/fields documented.
 - Rename all files to from TitleCase to lower_snake_case.
   * This is because of Google's protobuf style guide https://developers.google.com/protocol-buffers/docs/style
 - Renamed all services' methods from camelCase to TitleCase
   * This is because of Google's protobuf style guide https://developers.google.com/protocol-buffers/docs/style
 - Renamed enum variants from TitleCase to SCREAMING_SNAKE_CASE
   * This is because of Google's protobuf style guide https://developers.google.com/protocol-buffers/docs/style
 - Updated indentation to always be 4 spaces
   * Before for some of the new files such as `Schedule*.proto` the indentation was 2 spaces, while the rest of the files had 4 spaces.
 - Moved the copyright comment to the top of the file for all protobuf files, and removed unprintable characters <200b>, <200c>, and <200d> which were in the copyright comment.
   * I'm not sure of those unprintable characters had any meaning, but if they were important they can be re-added.
   * I'm not sure if this would be considered a breaking change.
 - Renamed protobuf package from `proto` to `services` for the `services/*.proto` files
 - Updated all imports to use full path from root of the repo.
   * E.g. `import "basic_types.proto"` -> `import "services/basic_types.proto"`
     - This affects `mirror/`, `services/`, and `streams/`
   * E.g. `Timestamp consensus_timestamp` -> `services.Timestamp consensus_timestamp`
     - This only affects `mirror/` and `streams/`


danielakhterov:
Closing as we determined this needs to be split up.

PR #15, opened 2021-03-11
Update protobufs to satisfy revised scheduled txns spec

tinker-michaelj:
Please see the [revised scheduled transactions spec](https://github.com/hashgraph/hedera-services/blob/01155-D-ReviseScheduleProtobufs/docs/scheduled-transactions/revised-spec.md) for more context.

### DIFF
--- a/services/ResponseCode.proto
+++ b/services/ResponseCode.proto
@@ -215,7 +215,7 @@ enum ResponseCodeEnum {
   INVALID_SCHEDULE_ACCOUNT_ID = 204; // The Schedule Create Transaction TransactionID account does not exist
   NO_NEW_VALID_SIGNATURES = 205; // The provided sig map did not contain any new valid signatures from required signers of the scheduled transaction
   UNRESOLVABLE_REQUIRED_SIGNERS = 206; // The required signers for a scheduled transaction cannot be resolved, for example because they do not exist or have been deleted
-  SCHEDULED_TRANSACTION_NOT_IN_WHITELIST = 207; // ScheduleCreate and ScheduleSign transactions cannot be scheduled
+  SCHEDULED_TRANSACTION_NOT_IN_WHITELIST = 207; // Only whitelisted transaction types may be scheduled
   SOME_SIGNATURES_WERE_INVALID = 208; // At least one of the signatures in the provided sig map did not represent a valid signature for any required signer
   TRANSACTION_ID_FIELD_NOT_ALLOWED = 209; // The scheduled field in the TransactionID may not be set to true
   IDENTICAL_SCHEDULE_ALREADY_CREATED = 210; // A schedule already exists with the same identifying fields of an attempted ScheduleCreate (that is, all fields other than scheduledPayerAccountID)
--- a/services/ResponseCode.proto
+++ b/services/ResponseCode.proto
@@ -218,5 +218,8 @@ enum ResponseCodeEnum {
   UNPARSEABLE_SCHEDULED_TRANSACTION = 207; // The bytes allegedly representing a transaction to be scheduled could not be parsed
   UNSCHEDULABLE_TRANSACTION = 208; // ScheduleCreate and ScheduleSign transactions cannot be scheduled
   SOME_SIGNATURES_WERE_INVALID = 209; // At least one of the signatures in the provided sig map did not represent a valid signature for any required signer
-  TRANSACTION_ID_FIELD_NOT_ALLOWED = 210; // The <tt>scheduled</tt> and <tt>nonce</tt> fields in the <tt>TransactionID</tt> may not be set in a top-level transaction
+  TRANSACTION_ID_FIELD_NOT_ALLOWED = 210; // The scheduled field in the TransactionID may not be set in a top-level transaction; the TransactionID of a scheduled transaction must be empty
+  IDENTICAL_SCHEDULE_ALREADY_CREATED = 211; // A schedule already exists with the same identifying fields of an attempted ScheduleCreate (that is, all fields other than scheduledPayerAccountID)
+  UTF8_ENCODING_CONTAINS_ZERO_BYTE = 212; // A string field in the transaction has a UTF-8 encoding with the prohibited zero byte
+  SCHEDULE_WAS_DELETED = 213; // A schedule being signed or deleted has already been deleted.
 }
--- a/services/ResponseCode.proto
+++ b/services/ResponseCode.proto
@@ -218,7 +218,7 @@ enum ResponseCodeEnum {
   UNPARSEABLE_SCHEDULED_TRANSACTION = 207; // The bytes allegedly representing a transaction to be scheduled could not be parsed
   UNSCHEDULABLE_TRANSACTION = 208; // ScheduleCreate and ScheduleSign transactions cannot be scheduled
   SOME_SIGNATURES_WERE_INVALID = 209; // At least one of the signatures in the provided sig map did not represent a valid signature for any required signer
-  TRANSACTION_ID_FIELD_NOT_ALLOWED = 210; // The scheduled field in the TransactionID may not be set in a top-level transaction; the TransactionID of a scheduled transaction must be empty
+  TRANSACTION_ID_FIELD_NOT_ALLOWED = 210; // The scheduled field in the TransactionID may not be set to true
   IDENTICAL_SCHEDULE_ALREADY_CREATED = 211; // A schedule already exists with the same identifying fields of an attempted ScheduleCreate (that is, all fields other than scheduledPayerAccountID)
   UTF8_ENCODING_CONTAINS_ZERO_BYTE = 212; // A string field in the transaction has a UTF-8 encoding with the prohibited zero byte
   SCHEDULE_WAS_DELETED = 213; // A schedule being signed or deleted has already been deleted.
--- a/services/ResponseCode.proto
+++ b/services/ResponseCode.proto
@@ -219,7 +219,7 @@ enum ResponseCodeEnum {
   SOME_SIGNATURES_WERE_INVALID = 208; // At least one of the signatures in the provided sig map did not represent a valid signature for any required signer
   TRANSACTION_ID_FIELD_NOT_ALLOWED = 209; // The scheduled field in the TransactionID may not be set to true
   IDENTICAL_SCHEDULE_ALREADY_CREATED = 210; // A schedule already exists with the same identifying fields of an attempted ScheduleCreate (that is, all fields other than scheduledPayerAccountID)
-  UTF8_ENCODING_CONTAINS_ZERO_BYTE = 211; // A string field in the transaction has a UTF-8 encoding with the prohibited zero byte
-  SCHEDULE_WAS_DELETED = 212; // A schedule being signed or deleted has already been deleted
-  SCHEDULE_WAS_EXECUTED = 213; // A schedule being signed has already been executed
+  INVALID_ZERO_BYTE_IN_STRING = 211; // A string field in the transaction has a UTF-8 encoding with the prohibited zero byte
+  SCHEDULE_ALREADY_DELETED = 212; // A schedule being signed or deleted has already been deleted
+  SCHEDULE_ALREADY_EXECUTED = 213; // A schedule being signed or deleted has already been executed
 }
--- a/services/ResponseCode.proto
+++ b/services/ResponseCode.proto
@@ -222,4 +222,5 @@ enum ResponseCodeEnum {
   IDENTICAL_SCHEDULE_ALREADY_CREATED = 211; // A schedule already exists with the same identifying fields of an attempted ScheduleCreate (that is, all fields other than scheduledPayerAccountID)
   UTF8_ENCODING_CONTAINS_ZERO_BYTE = 212; // A string field in the transaction has a UTF-8 encoding with the prohibited zero byte
   SCHEDULE_WAS_DELETED = 213; // A schedule being signed or deleted has already been deleted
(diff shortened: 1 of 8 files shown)


bugbytesinc:
Have you considered, instead of creating an extra parallel `SchedulableTransactionBody` which is cut-and-paste copy of a `TransactionBody`, instead introduce a "SchedulingInfo" object or similar as a part of the pre-existing `TransactionBody` with the necessary information to identify the transaction as being scheduled.  Or another alternative could be to modify `TransactionBody` altering it to hold a one-of `TransactionID` or `ScheduledTransactionID` or something similar.

services/ResponseCode.proto
```diff
@@ -218,5 +218,9 @@ enum ResponseCodeEnum {
   UNPARSEABLE_SCHEDULED_TRANSACTION = 207; // The bytes allegedly representing a transaction to be scheduled could not be parsed
   UNSCHEDULABLE_TRANSACTION = 208; // ScheduleCreate and ScheduleSign transactions cannot be scheduled
   SOME_SIGNATURES_WERE_INVALID = 209; // At least one of the signatures in the provided sig map did not represent a valid signature for any required signer
-  TRANSACTION_ID_FIELD_NOT_ALLOWED = 210; // The <tt>scheduled</tt> and <tt>nonce</tt> fields in the <tt>TransactionID</tt> may not be set in a top-level transaction
+  TRANSACTION_ID_FIELD_NOT_ALLOWED = 210; // The scheduled field in the TransactionID may not be set to true
+  IDENTICAL_SCHEDULE_ALREADY_CREATED = 211; // A schedule already exists with the same identifying fields of an attempted ScheduleCreate (that is, all fields other than scheduledPayerAccountID)
+  UTF8_ENCODING_CONTAINS_ZERO_BYTE = 212; // A string field in the transaction has a UTF-8 encoding with the prohibited zero byte
```
steven-sheehy, line 223: Perhaps just `STRING_CONTAINS_ZERO_BYTE` or `INVALID_ZERO_BYTE_IN_STRING`? All proto strings are utf-8 encoding so that's assumed. What's important that it is a field of type string.

services/SchedulableTransactionBody.proto
```diff
@@ -0,0 +1,116 @@
...
+import "TokenDissociate.proto";
+
+import "ScheduleDelete.proto";
+
+/* A schedulable transaction. Note that the global/dynamic system property 
+<tt>scheduling.whitelist</tt> controls which transaction types may be scheduled. 
+In Hedera Services 0.13.0, it will include only <tt>CryptoTransfer</tt> and
+<tt>ConsensusSubmitMessage</tt> functions. */
+message SchedulableTransactionBody {
+  uint64 transactionFee = 1; // The maximum transaction fee the client is willing to pay
```
steven-sheehy, line 75: When the transaction is executed will you be constructing an empty `TransactionBody` then copy all these fields over explicitly? Or will you be reconstituting a `TransactionBody` from the `SchedulableTransactionBody` bytes e.g. `TransactionBody.parseFrom(scheduleableTransactionBodyBytes)`. If the latter, you will need to adjust these field indexes to match `TransactionBody`.

services/SchedulableTransactionBody.proto
```diff
@@ -0,0 +1,116 @@
...
+    TokenRevokeKycTransactionBody tokenRevokeKyc = 26; // Revokes KYC of an account for a token
+    TokenDeleteTransactionBody tokenDeletion = 27; // Deletes a token instance
+    TokenUpdateTransactionBody tokenUpdate = 28; // Updates a token instance
+    TokenMintTransactionBody tokenMint = 29; // Mints new tokens to a token's treasury account
+    TokenBurnTransactionBody tokenBurn = 30; // Burns tokens from a token's treasury account
+    TokenWipeAccountTransactionBody tokenWipe = 31; // Wipes amount of tokens from an account
+    TokenAssociateTransactionBody tokenAssociate = 32; // Associate tokens to an account
+    TokenDissociateTransactionBody tokenDissociate = 33; // Dissociate tokens from an account
+
+    ScheduleDeleteTransactionBody scheduleDelete = 43; // Marks a schedule in the network's action queue as deleted, preventing it from executing
```
steven-sheehy, line 114: If we're not matching the field indexes, then this shouldn't jump from 33 to 43.

services/ScheduleCreate.proto
```diff
@@ -26,28 +26,46 @@ option java_package = "com.hederahashgraph.api.proto.java";
...
-  Key adminKey = 2; // (optional) The Key which is able to delete the Scheduled Transaction (if tx is not already executed)
-  AccountID payerAccountID = 3; // (optional) The account which is going to pay for the execution of the Scheduled TX. If not populated, the scheduling account is charged
-  SignatureMap sigMap = 4; // (optional) Signatures that could be provided (similarly to how signatures are provided in ScheduleSign operation) on Scheduled Transaction creation
-  string memo = 5; // (optional) Publicly visible information about the Scheduled entity, up to 100 bytes. No guarantee of uniqueness.
+  SchedulableTransactionBody scheduledTransactionBody = 1; // The scheduled transaction
+  string memo = 2; // An optional memo with a UTF-8 encoding of no more than 100 bytes which does not contain the zero byte
+  Key adminKey = 3; // An optional Hedera key which can be used to sign a ScheduleDelete and remove the schedule
+  AccountID payerAccountID = 4; // An optional id of the account to be charged the service fee for the scheduled transaction at the consensus time that it executes (i    f ever); defaults to the ScheduleCreate payer if not given
```
steven-sheehy, line 70: Lots of space between the i and f of `if`

services/ScheduleGetInfo.proto
```diff
@@ -29,35 +29,36 @@ import "BasicTypes.proto";
...
-  string memo = 7; // Publicly visible information about the Scheduled entity, up to 100 bytes. No guarantee of uniqueness.
-  Timestamp expirationTime = 8; // The epoch second at which the schedule will expire
-  TransactionID scheduledTransactionID = 9; // TransactionID that should be used to query for the receipt or record of the scheduled transaction
-}
+  Information summarizing schedule state 
+*/
+message ScheduleInfo {  
+  ScheduleID scheduleID = 1; // The id of the schedule
+  bool deleted = 2; // Has the schedule been deleted?
+  bool executed = 3; // Has the schedule been executed?
```
steven-sheehy, line 50: What about instead of a boolean a consensus Timestamp at which it was executed? This conveys the same info as a boolean (it's executed if present) with additional information about when it was executed.

services/ScheduleGetInfo.proto
```diff
@@ -29,35 +29,36 @@ import "BasicTypes.proto";
...
-  Key adminKey = 6; // The Key which is able to delete the Scheduled Transaction if set
-  string memo = 7; // Publicly visible information about the Scheduled entity, up to 100 bytes. No guarantee of uniqueness.
-  Timestamp expirationTime = 8; // The epoch second at which the schedule will expire
-  TransactionID scheduledTransactionID = 9; // TransactionID that should be used to query for the receipt or record of the scheduled transaction
-}
+  Information summarizing schedule state 
+*/
+message ScheduleInfo {  
+  ScheduleID scheduleID = 1; // The id of the schedule
+  bool deleted = 2; // Has the schedule been deleted?
```
steven-sheehy, line 49: What about instead of a boolean a consensus Timestamp at which it was deleted? This conveys the same info as a boolean (it's deleted if present) with additional information about when it was deleted.

steven-sheehy (changes requested):

services/SchedulableTransactionBody.proto
tinker-michaelj: I've done [the former](https://github.com/hashgraph/hedera-services/blob/2cae66315472f6c844d532b59b0201d195dcb697/hedera-node/src/main/java/com/hedera/services/state/merkle/MerkleSchedule.java#L377) in my Services PR. 👍

services/ResponseCode.proto
tinker-michaelj: Used `INVALID_ZERO_BYTE_IN_STRING`.

services/SchedulableTransactionBody.proto
tinker-michaelj: Fixed.

services/ScheduleCreate.proto
tinker-michaelj: Fixed.

services/SchedulableTransactionBody.proto
```diff
@@ -0,0 +1,114 @@
...
+import "ScheduleDelete.proto";
+
+/* A schedulable transaction. Note that the global/dynamic system property 
+<tt>scheduling.whitelist</tt> controls which transaction types may be scheduled. 
+In Hedera Services 0.13.0, it will include only <tt>CryptoTransfer</tt> and
+<tt>ConsensusSubmitMessage</tt> functions. */
+message SchedulableTransactionBody {
+  uint64 transactionFee = 1; // The maximum transaction fee the client is willing to pay
+  string memo = 2; // A memo to include the execution record; the UTF-8 encoding may be up to 100 bytes and must not include the zero byte
+  oneof data {
```
danielakhterov, line 75: There is no inherit problem with this, but I wanted to mention that by naming fields inside of a `oneof` in camelCase instead of lower_snake_case it makes it harder to update these fields later since it breaks the generated Java protobuf code. Java generates a `DataCase` enum with values `CONTRACTCALL` for `contractCall` and `CONTRACT_CALL` for `contract_call`. This is why in https://github.com/hashgraph/hedera-protobufs/pull/13 I wasn't able to rename fields inside a `oneof`. If that PR is to ever be merged (which from my understand it is), then it'd easier to update if these fields were renamed to lower_snake_case now, rather than later. If this cannot be done, then that is fine, but just wanted to make sure you're aware of  this issue.

tinker-michaelj: Ah I didn't know this; renamed to lower_snake_case!

services/ScheduleSign.proto
```diff
@@ -28,14 +28,24 @@ option java_multiple_files = true;
...
 
 /*
-Provides signatures for a given Scheduled Transaction. The Scheduled Transaction executes if all of the required signatures are witnessed.
-INVALID_SCHEDULE_ID is returned if the specified ScheduleID does not exist; or has now expired, been deleted, or been executed
-UNRESOLVABLE_REQUIRED_SIGNERS is returned if any of the required keys were updated/deleted and cannot be resolved
-INVALID_ACCOUNT_ID is returned if the payerAccountID has been deleted or cannot be resolved
-NO_NEW_VALID_SIGNATURES is returned if there are no new signatures provided in the sigMap to be witnessed
-SOME_SIGNATURES_WERE_INVALID is returned if one of the signatures provided does not represent a valid signature for any required signer
+Adds signing keys to a scheduled transaction. If the new keys satisfy the scheduled transaction's signing
+requirements, and it was scheduled with <tt>executeImmediatelyOnValidSignatures=true</tt>, it will become
```
steven-sheehy, line 32: executeImmediatelyOnValidSignatures should be removed since it's no longer applicable.

services/ScheduleSign.proto
```diff
@@ -28,14 +28,24 @@ option java_multiple_files = true;
...
 /*
-Provides signatures for a given Scheduled Transaction. The Scheduled Transaction executes if all of the required signatures are witnessed.
-INVALID_SCHEDULE_ID is returned if the specified ScheduleID does not exist; or has now expired, been deleted, or been executed
-UNRESOLVABLE_REQUIRED_SIGNERS is returned if any of the required keys were updated/deleted and cannot be resolved
-INVALID_ACCOUNT_ID is returned if the payerAccountID has been deleted or cannot be resolved
-NO_NEW_VALID_SIGNATURES is returned if there are no new signatures provided in the sigMap to be witnessed
-SOME_SIGNATURES_WERE_INVALID is returned if one of the signatures provided does not represent a valid signature for any required signer
+Adds signing keys to a scheduled transaction. If the new keys satisfy the scheduled transaction's signing
+requirements, and it was scheduled with <tt>executeImmediatelyOnValidSignatures=true</tt>, it will become 
+eligible for execution. In Hedera Services 0.13.0, this will in fact cause its execution immediately
```
steven-sheehy, line 33: Should not mention specific versions in case they change.

services/ScheduleSign.proto
```diff
@@ -28,14 +28,24 @@ option java_multiple_files = true;
...
+after the triggering <tt>ScheduleSign</tt>; but future releases may change this sequencing.
+
+Upon <tt>SUCCESS</tt>, the receipt includes the <tt>scheduledTransactionID</tt> to use to query for the 
+record of the scheduled transaction's execution (if it occurs). 
+
+Other notable response codes include,
+<ul>
+  <li> <tt>INVALID_SCHEDULE_ID</tt> if there is no schedule with the given <tt>ScheduleID</tt>.
```
steven-sheehy, line 41: Perhaps to stay DRY, we should list only the response codes without a description and to mention to see responsecode proto for their description?

steven-sheehy (changes requested):

services/ScheduleSign.proto
tinker-michaelj: Done.

services/ScheduleSign.proto
tinker-michaelj: Done.

services/ScheduleSign.proto
tinker-michaelj: Done.

steven-sheehy (approved):
LGTM

PR #19, opened 2021-03-18
00017 d hcs message size

ljianghedera:
Added a response enum code for the hedera services issue https://github.com/hashgraph/hedera-services/issues/1145


tinker-michaelj (approved):
LGTM

PR #20, opened 2021-03-19
Update README 

tinker-michaelj:
- Add more info on tagging conventions.
- Demote scheduled txns from a top-level service.


SimiHunjan (approved):
Looks great!

Issue #21, opened 2021-03-22
0.0.101 NodeAddressBook contains duplicate entries for each distinct IP


anighanta:
**Summary**
As a Hedera Services client, I need the address book to contain one entry per node so it's easier to consume and doesn't duplicate information. Currently, for the 0.0.101 address book that contains IP address information it duplicates each `NodeAddress` for each distinct IP that that node has. This is confusing since the node count shows as 39 in mainnet (via `NodeAddressBook.getNodeAddressCount()`) despite there only being 13ish nodes. It's also wasteful to duplicate information as every client who wants to communicate with the network will be pulling this file to get up to date IP information.

The proto currently looks like this:

```proto
message NodeAddress {
    bytes ipAddress = 1; // The ip address of the Node with separator & octets
    int32 portno = 2; // The port number of the grpc server for the node
    bytes memo = 3; // The memo field of the node (usage to store account ID is deprecated)
    string RSA_PubKey = 4; // The RSA public key of the node
    int64 nodeId = 5; // A non-sequential identifier for the node
    AccountID nodeAccountId = 6; // The account to be paid for queries and transactions sent to this node
    bytes nodeCertHash = 7; // A hash of the X509 cert used for gRPC traffic to this node
}

message NodeAddressBook {
    repeated NodeAddress nodeAddress = 1; // Contains multiple Node Address for the network
}
```

**Possible resolution**
File 0.0.101 will be NodeAddressBookAbbreviated
File 0.0.102 will be NodeAddressBook
```proto
message NodeEndpoint {
    string ipAddress = 1; // The IP address of the node
    string port = 2; // The port of the node
}

message NodeAddressAbbreviated {
    bytes ipAddress = 1 [deprecated=true]; // The IP address of the Node with separator & octets
    int32 portno = 2 [deprecated=true];// The port number of the grpc server for the node
    bytes memo = 3; [deprecated=true]; // The memo field of the node (usage to store account ID is deprecated)
    string RSA_PubKey = 4 [deprecated=true]; // The RSA public key of the node
    int64 nodeId = 5; // A non-sequential identifier for the node. This value is the key between the full and abbreviated address books 
    AccountID nodeAccountId = 6; // The account to be paid for queries and transactions sent to this node
    bytes nodeCertHash = 7 [deprecated=true]; // A hash of the X509 cert used for gRPC traffic to this node
    repeated NodeEndpoint = 8; // A node's IP address and port
}

message NodeAddressBookAbbreviated {
[...]

PR #22, opened 2021-03-22
Fix Duplicated entries in AddressBook 0.0.101

anighanta:
Resolves issue: #21 
Parent Issue : https://github.com/hashgraph/hedera-services/issues/750

Signed-off-by: anighanta <anirudh.ghanta@hedera.com>

### DIFF
--- a/services/BasicTypes.proto
+++ b/services/BasicTypes.proto
@@ -349,7 +349,7 @@ message NodeEndpoint {
 /*
 The metadata for a Node – including IP Address, and the crypto account associated with the Node.
 
-Note :  This structure is intended to be used to parse 0.0.102, use NodeAddressAbbreviated to parse 0.0.101.
+Note :  This structure is intended to be used to parse 0.0.102, use NodeAddressForClients to parse 0.0.101.
         Usage of NodeAddress to parse 0.0.101 is deprecated.
 */
 message NodeAddress {
@@ -368,12 +368,12 @@ message NodeAddress {
 /*
 A list of nodes and their metadata.
 */
-message NodeAddressBook {
+message AddressBook {
     repeated NodeAddress nodeAddress = 1; // Contains all details of the nodes for the network
 }
 
 /* All the information required to connect to the network.*/
-message NodeAddressAbbreviated {
+message NodeAddressForClients {
     int64 nodeId = 5; // A non-sequential identifier for the node. This value is the key between the full and abbreviated address books
     AccountID nodeAccountId = 6; // The account to be paid for queries and transactions sent to this node
     bytes nodeCertHash = 7; // A hash of the X509 cert used for gRPC traffic to this node
@@ -381,8 +381,8 @@ message NodeAddressAbbreviated {
 }
 
 /* A List of all nodes and just their information needed to connect to the network.*/
-message NodeAddressBookAbbreviated {
-    repeated NodeAddressAbbreviated nodeAddressAbbreviated = 1; // Contains minimal node details for the network
+message AddressBookForClients {
+    repeated NodeAddressForClients nodeAddressForClients = 1; // Contains minimal node details for the network
 }
 
 message SemanticVersion {
--- a/services/BasicTypes.proto
+++ b/services/BasicTypes.proto
@@ -353,8 +353,8 @@ Note :  This structure is intended to be used to parse 0.0.102, use NodeAddressF
         Usage of NodeAddress to parse 0.0.101 is deprecated.
 */
 message NodeAddress {
-    bytes ipAddress = 1 [deprecated=true]; // The ip address of the Node with separator & octets
-    int32 portno = 2 [deprecated=true]; // The port number of the grpc server for the node
+    bytes ipAddress = 1 [deprecated=true]; // The ip address of the Node with separator & octets. Usage is deprecated, NodeEndPoint is preferred to retrieve a node's list of IP addresses and ports
+    int32 portno = 2 [deprecated=true]; // The port number of the grpc server for the node.  Usage is deprecated, NodeEndPoint is preferred to retrieve a node's list of IP addresses and ports
     bytes memo = 3 [deprecated=true]; // The memo field of the node (usage to store account ID is deprecated)
     string RSA_PubKey = 4; // The RSA public key of the node
     int64 nodeId = 5; // A non-sequential identifier for the node
@@ -366,10 +366,10 @@ message NodeAddress {
 }
 
 /*
-A list of nodes and their metadata.
+A list of nodes and their metadata that contains all details of the nodes for the network
 */
 message AddressBook {
-    repeated NodeAddress nodeAddress = 1; // Contains all details of the nodes for the network
+    repeated NodeAddress nodeAddress = 1; // Metadata of nodes on the network
 }
 
 /* All the information required to connect to the network.*/
--- a/services/BasicTypes.proto
+++ b/services/BasicTypes.proto
@@ -383,7 +383,7 @@ message ClientNodeAddress {
 
 /* A List of all nodes and just their information needed to connect to the network.*/
 message ClientNodeAddressBook {
-    repeated ClientNodeAddress nodeAddressAbbreviated = 1; // Contains minimal node details for the network
+    repeated ClientNodeAddress clientNodeAddress = 1; // Contains minimal node details for the network
 }
 
 message SemanticVersion {
--- a/services/BasicTypes.proto
+++ b/services/BasicTypes.proto
@@ -348,6 +348,9 @@ message NodeEndpoint {
 
 /*
 The metadata for a Node – including IP Address, and the crypto account associated with the Node.
+
+Note :  This structure is intended to be used to parse 0.0.102, use ClientNodeAddress to parse 0.0.101.
+        Usage of NodeAddress to parse 0.0.101 is deprecated.
 */
 message NodeAddress {
     bytes ipAddress = 1 [deprecated=true]; // The ip address of the Node with separator & octets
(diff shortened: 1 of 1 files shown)


steven-sheehy:
I have outstanding questions in the services ticket that I would like to see answered.

anighanta:
> I have outstanding questions in the services ticket that I would like to see answered.

https://github.com/hashgraph/hedera-services/issues/750#issuecomment-804425932

services/BasicTypes.proto
```diff
@@ -340,24 +340,50 @@ message CurrentAndNextFeeSchedule {
...
 /*
 A list of nodes and their metadata.
 */
 message NodeAddressBook {
-    repeated NodeAddress nodeAddress = 1; // Contains multiple Node Address for the network
+    repeated NodeAddress nodeAddress = 1; // Contains all details of the nodes for the network
+}
+
+/* A version of NodeAddress to allow backward compatibility.*/
+message NodeAddressAbbreviated {
```
steven-sheehy, line 373: Talked with Donald and we preferred the name `ClientNodeAddress`.

services/BasicTypes.proto
```diff
@@ -340,24 +340,50 @@ message CurrentAndNextFeeSchedule {
...
+/* This contains the IP Address and the port representing an endpoint of a Node in a network */
+message NodeEndpoint {
+    string ipAddress = 1; // The IP address of the node
+    string port = 2; // The port of the node
+}
+
 /*
 The metadata for a Node – including IP Address, and the crypto account associated with the Node.
 */
 message NodeAddress {
```
steven-sheehy, line 352: Need to add a comment that this structure is intended to be used to parse `0.0.102` and that its use to parse `0.0.101` is deprecated and to use the new message type.

services/BasicTypes.proto
```diff
@@ -340,24 +340,50 @@ message CurrentAndNextFeeSchedule {
...
+    bytes memo = 3; // The memo field of the node (usage to store account ID is deprecated)
+    string RSA_PubKey = 4; // The RSA public key of the node
+    int64 nodeId = 5; // A non-sequential identifier for the node. This value is the key between the full and abbreviated address books
+    AccountID nodeAccountId = 6; // The account to be paid for queries and transactions sent to this node
+    bytes nodeCertHash = 7; // A hash of the X509 cert used for gRPC traffic to this node
+    repeated NodeEndpoint nodeEndpoint = 8; // A node's IP address and port
+}
+
+/* A version of NodeAddressBook to allow backward compatibility.*/
+message NodeAddressBookAbbreviated {
```
steven-sheehy, line 385: Talked with Donald and we preferred the name `ClientNodeAddressBook`.

services/BasicTypes.proto
```diff
@@ -340,24 +340,50 @@ message CurrentAndNextFeeSchedule {
...
+
+/* A version of NodeAddress to allow backward compatibility.*/
+message NodeAddressAbbreviated {
+    bytes ipAddress = 1; // The IP address of the Node with separator & octets
+    int32 portno = 2; // The port number of the grpc server for the node
+    bytes memo = 3; // The memo field of the node (usage to store account ID is deprecated)
+    string RSA_PubKey = 4; // The RSA public key of the node
+    int64 nodeId = 5; // A non-sequential identifier for the node. This value is the key between the full and abbreviated address books
+    AccountID nodeAccountId = 6; // The account to be paid for queries and transactions sent to this node
+    bytes nodeCertHash = 7; // A hash of the X509 cert used for gRPC traffic to this node
```
steven-sheehy, line 380: In Nosh's writeup in the original ticket, he indicated `nodeCertHash` is deprecated in the abbreviated message. I don't think that is the case since it is currently in the 0.0.101 file and is useful for clients to connect. Can you confirm if it should be present here or not?

services/BasicTypes.proto
```diff
@@ -340,24 +340,50 @@ message CurrentAndNextFeeSchedule {
...
+    repeated NodeEndpoint nodeEndpoint = 8; // A node's IP address and port
+}
+
+/* A version of NodeAddressBook to allow backward compatibility.*/
+message NodeAddressBookAbbreviated {
+    repeated NodeAddressAbbreviated nodeAddressAbbreviated = 1; // Contains minimal node details for the network
```
steven-sheehy, line 386: Can you test that the bytes from a current `0.0.101` file in production generated from the previous `NodeAddressBook` message can be read without issue by `NodeAddressBookAbbreviated.parseFrom()` and all fields are present? There was some concern that the approach we're taking here may not be backwards compatible.

steven-sheehy (changes requested):

services/BasicTypes.proto
anighanta: Removed `deprecated=true` after yesterday's discussion.

services/BasicTypes.proto
anighanta: Sure, will test and let you know.

services/BasicTypes.proto
Neeharika-Sompalli: @donaldthibeau Could you please re-confirm the change of name in protobuf. Thanks!

services/BasicTypes.proto
Neeharika-Sompalli: @donaldthibeau Could you please re-confirm the change of name in protobuf. Thanks!

services/BasicTypes.proto
Neeharika-Sompalli: Changed the name to `NodeAddressForClients` as per https://github.com/hashgraph/hedera-services/issues/750#issuecomment-806970447

services/BasicTypes.proto
Neeharika-Sompalli: Changed the name to `AddressBookForClients` as per hashgraph/hedera-services#750 (comment)

services/BasicTypes.proto
```diff
@@ -340,24 +340,49 @@ message CurrentAndNextFeeSchedule {
...
 The metadata for a Node – including IP Address, and the crypto account associated with the Node.
+
+Note :  This structure is intended to be used to parse 0.0.102, use NodeAddressForClients to parse 0.0.101.
+        Usage of NodeAddress to parse 0.0.101 is deprecated.
 */
 message NodeAddress {
-    bytes ipAddress = 1; // The ip address of the Node with separator & octets
-    int32 portno = 2; // The port number of the grpc server for the node
-    bytes memo = 3; // The memo field of the node (usage to store account ID is deprecated)
+    bytes ipAddress = 1 [deprecated=true]; // The ip address of the Node with separator & octets
```
steven-sheehy, line 356: Should add a comment that it's deprecated since `nodeEndpoint` is now preferred to retrieve a node's list of IP addresses.

services/BasicTypes.proto
```diff
@@ -340,24 +340,49 @@ message CurrentAndNextFeeSchedule {
...
+    int32 portno = 2 [deprecated=true]; // The port number of the grpc server for the node
```
steven-sheehy, line 357: Should add a comment that it's deprecated since `nodeEndpoint` is now preferred to retrieve a node's list of IP addresses and ports.

services/BasicTypes.proto
```diff
@@ -340,24 +340,49 @@ message CurrentAndNextFeeSchedule {
...
+    bytes memo = 3 [deprecated=true]; // The memo field of the node (usage to store account ID is deprecated)
```
steven-sheehy, line 358: Should add a comment that it's deprecated since`nodeAccountId` and `description` should be used instead.

steven-sheehy (changes requested):

services/BasicTypes.proto
```diff
@@ -340,24 +340,49 @@ message CurrentAndNextFeeSchedule {
...
 */
-message NodeAddressBook {
-    repeated NodeAddress nodeAddress = 1; // Contains multiple Node Address for the network
+message AddressBook {
+    repeated NodeAddress nodeAddress = 1; // Contains all details of the nodes for the network
+}
+
+/* All the information required to connect to the network.*/
+message NodeAddressForClients {
+    int64 nodeId = 5; // A non-sequential identifier for the node. This value is the key between the full and abbreviated address books
```
danielakhterov, line 377: Can you elaborate on this more please; reading the comments I'm not entirely sure what `nodeId` is; what are these `key`'s between full and abbreviated address books? Also, what is an abbreviated address book? I understand this field is a `int64` and `nodeAccountId` is an `AccountID`, but what is the difference between these two fields.

anighanta: The Idea behind this change is that, as we have multiple IPs for each node, when we get the addressBook 0.0.101 of mainnet for example.. we get a list with repeated entries for each node. To resolve this we want to create a new abbreviated version of the addressBook which will have a repeated `Endpoint`. So in future, when we get the addressbook we will get just enough information to connect to the network using this abbreviated message. 

Note: this  abbreviated message is `AddressBookForClients`.

A nodeId is just a sequential identifier of a node and the nodeAccountID is the `AccountID` associated with the node. Right now although they represent different things, their value is the same as in if nodeID is `N`, its account id is `0.0.N`

steven-sheehy: > if nodeID is N, its account id is 0.0.N

Correction: if nodeID is N, its account id is 0.0.(N + 2). Example: node 1 is 0.0.3

steven-sheehy: This is just the case now, but won't always be the case. Nodes can change their account ID in the future to something else while the node ID has to stay the same. Neither will necessarily stay sequential in the future either.

services/BasicTypes.proto
steven-sheehy: Although in practice we never populated the memo with a description. So maybe we can leave off the latter part.

services/BasicTypes.proto
anighanta: We have tested this scenario and it works

steven-sheehy (approved):
LGTM

PR #27, opened 2021-04-09
Revise address book types to 0.13.0 standard

tinker-michaelj:
**Summary of the change**

1. File `0.0.101` to be populated with just connectivity and payment info for all nodes.
2. File `0.0.102` to be populated with a full metadata for all nodes.


steven-sheehy (approved):
LGTM

PR #29, opened 2021-04-21
re-add deprecated fields to Transaction

mehcode:
previously deprecated fields are still needed to be able to historically parse valid transactions

### DIFF
--- a/services/Transaction.proto
+++ b/services/Transaction.proto
@@ -28,12 +28,19 @@ option java_multiple_files = true;
 
 import "Duration.proto";
 import "BasicTypes.proto";
+import "TransactionBody.proto";
 
 /* A single signed transaction, including all its signatures. The SignatureList will have a Signature for each Key in the transaction, either explicit or implicit, in the order that they appear in the transaction. For example, a CryptoTransfer will first have a Signature corresponding to the Key for the paying account, followed by a Signature corresponding to the Key for each account that is sending or receiving cryptocurrency in the transfer. Each Transaction should not have more than 50 levels. 
  * The SignatureList field is deprecated and succeeded by SignatureMap.
  */
 message Transaction {
-    bytes signedTransactionBytes = 5; // SignedTransaction serialized into bytes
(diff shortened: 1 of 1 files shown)


danielakhterov:
Yea, this is quite important for the mirror nodes; not sure how Hedera's official mirror node is working without these fields.

danielakhterov (approved):
LGTM

steven-sheehy:
The mirror node just looks at the protobufs unknown fields by its old field identifier. It can still parse old transactions using this approach.

mehcode:
That makes sense. The Rust protobuf implementations don't support that. It's a more opinionated deserialization.

steven-sheehy:
Maybe not prost. What about https://docs.rs/protobuf/2.8.1/protobuf/struct.UnknownFields.html ?

mehcode:
That lib is not well supported by the community. For example the prominent gRPC lib - `tonic` - requires `prost`. 

---

Neither here nor there though. Are you objecting to this PR? These are the only deprecated fields that were actually removed.

steven-sheehy:
It might be fine to re add them for this one off. My concern is that sets a precedent that we can never evolve our API by removing fields due to the limitations of one particular proto library. This negates one of the advantages of using protobuf in that it maintains backwards compatibility and doesn't require us to implement V2 models and APIs whenever we need to make bigger changes. We will need to discuss with internal stakeholders.

mehcode:
Well, no. The API can evolve as much as it wants. As long as the deprecated fields remain forever present. The `one of` constructs work fine to hide a v1 type inside of a v2 structure. This is very similar to the (very well received) stability guarantees that Rust provides. The API can evolve as much as it wants, and deprecate as much as it needs, as long as nothing is ever _removed_.

tinker-michaelj (approved):
Would be good to emphasize in comment that even if valid signatures are provided via `sigs`, they will be completely ignored by Services.

qnswirlds (approved):

Neeharika-Sompalli (approved):

PR #33, opened 2021-04-28
Add full semver meta to SemanticVersion

tinker-michaelj:
Allow the Services `getVersionInfo` query to [automatically return full info](https://github.com/hashgraph/hedera-services/pull/1339) on a deployed version via Maven property filtering.
- Add the `pre` and `build` fields from the [semver spec](https://semver.org/#spec-item-9) to our `SemanticVersion` proto type.

### DIFF
--- a/services/BasicTypes.proto
+++ b/services/BasicTypes.proto
@@ -378,9 +378,11 @@ message NodeAddressBook {
 }
 
 message SemanticVersion {
-   int32 major = 1; // Increases with incompatible API changes
-   int32 minor = 2; // Increases with backwards-compatible new functionality
-   int32 patch = 3; // Increases with backwards-compatible bug fixes
+    int32 major = 1; // Increases with incompatible API changes
+    int32 minor = 2; // Increases with backwards-compatible new functionality
+    int32 patch = 3; // Increases with backwards-compatible bug fixes
+    string pre_release_version = 4; // A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers (https://semver.org/#spec-item-9); so given a semver 0.14.0-alpha.1+21AF26D3, this field would contain 'alpha.1'
+    string build_metadata = 5; // Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version (https://semver.org/#spec-item-10); so given a semver 0.14.0-alpha.1+21AF26D3, this field would contain '21AF26D3'
 }
 
 message Setting {
--- a/services/BasicTypes.proto
+++ b/services/BasicTypes.proto
@@ -377,6 +377,9 @@ message NodeAddressBook {
     repeated NodeAddress nodeAddress = 1; // Metadata of all nodes in the network
 }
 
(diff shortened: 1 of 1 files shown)


rbair23 (dismissed):

SimiHunjan (approved):
Looks great!

services/BasicTypes.proto
```diff
@@ -378,9 +378,11 @@ message NodeAddressBook {
 }
 
 message SemanticVersion {
-   int32 major = 1; // Increases with incompatible API changes
-   int32 minor = 2; // Increases with backwards-compatible new functionality
-   int32 patch = 3; // Increases with backwards-compatible bug fixes
+    int32 major = 1; // Increases with incompatible API changes
+    int32 minor = 2; // Increases with backwards-compatible new functionality
+    int32 patch = 3; // Increases with backwards-compatible bug fixes
+    string pre_release_version = 4; // A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers (https://semver.org/#spec-item-9); so given a semver 0.14.0-alpha.1+21AF26D3, this field would contain 'alpha.1'
```
steven-sheehy, line 384: Appending `_version` is redundant as this is within the `SemanticVersion` context. Also none of the others have that appended. Would prefer just `pre_release`.

services/BasicTypes.proto
```diff
@@ -378,9 +378,11 @@ message NodeAddressBook {
 }
 
 message SemanticVersion {
```
steven-sheehy, line 380: Perhaps add a top level comment that states that Hedera follows the semantic versioning spec and provide a link.

steven-sheehy (changes requested):

services/BasicTypes.proto
tinker-michaelj: The [semver spec](https://semver.org/#spec-item-9) refers to it as a "pre-release version", and that's within the context of semver also. :D

services/BasicTypes.proto
tinker-michaelj: Done.

services/BasicTypes.proto
mehcode: I recommend following something like this: https://docs.rs/semver/0.11.0/semver/struct.Version.html#fields

 - `pre`
 - `build`

To keep things concise. It's obvious what those are, we don't need the `_release_version` suffix.

steven-sheehy: > The [semver spec](https://semver.org/#spec-item-9) refers to it as a "pre-release version", and that's within the context of semver also. 😁

In the paragraph before they also say `Major version` yet we don't have that field named as `major_version`. That's inconsistent. As that is a paragraph in English it is okay to add the version context. For a field name, it looks weird to have version twice in `semanticVersion.getPreReleaseVersion()`.

tinker-michaelj: Cool, shortened field names to `pre` and `build`!

steven-sheehy (approved):
LGTM

Issue #40, opened 2021-06-14
HIP 18 : Add Messages required for customFees in HBAR and TokenUnits


anighanta:
Resolves issue : https://github.com/hashgraph/hedera-services/issues/1586

Add Messages:

- [x] `Fraction` to denote the percentage and `FractionalFee` to denote the fraction with its minimum value to collect.
- [x] `FixedFee` To specify custom fixed Fee to pay in Token Units/ HBAR.
- [x] `CustomFee` to specify one of FractionalFee / FixedFee along with`receivingAccountID`
- [x] `CustomFees` to encapsulate multiple `CustomFee` of a token.
- [x] `CustomTokenUnitFees` To specify the tansferList generated because of the associated `TokenUnitFee `
- [x] `CustomHbarFees` to Specify the  tansferList generated because of the associated `HbarFee`
- [x] Add `CustomFees` to `TokenCreate`, `TokenUpdate` transaction bodies and to `TokenInfo`
- [x] Add `CustomTokenUnitFees` and `CustomHbarFees` to TransactionRecord

PR #42, opened 2021-06-14
HIP 18

anighanta:
Closes issue : #40 and #41

### DIFF
--- a/services/BasicTypes.proto
+++ b/services/BasicTypes.proto
@@ -85,10 +85,18 @@ message TransferList {
     repeated AccountAmount accountAmounts = 1; // Multiple list of AccountAmount pairs, each of which has an account and an amount to transfer into it (positive) or out of it (negative)
 }
 
+/* A sender account, a receiver account, and the serial number of an NFT of a Token with NON_FUNGIBLE_UNIQUE type. */
+message NftTransfer {
+    AccountID senderAccountID = 1; // The accountID of the sender
+    AccountID receiverAccountID = 2; // The accountID of the receiver
+    int64 serialNumber = 3; // The serial number of the NFT
+}
+
 /* A list of token IDs and amounts representing the transferred out (negative) or into (positive) amounts, represented in the lowest denomination of the token */
 message TokenTransferList {
     TokenID token = 1; // The ID of the token
-    repeated AccountAmount transfers = 2; // Multiple list of AccountAmounts, each of which has an account and amount
+    repeated AccountAmount transfers = 2; // Applicable to tokens of type FUNGIBLE_COMMON. Multiple list of AccountAmounts, each of which has an account and amount
+    repeated NftTransfer nftTransfers = 3; // Applicable to tokens of type NON_FUNGIBLE_UNIQUE. Multiple list of NftTransfers, each of which has a sender and receiver account, including the serial number of the NFT
 }
 
 /* Unique identifier for a topic (used by the consensus service) */
@@ -110,6 +118,44 @@ message ScheduleID {
     int64 shardNum = 1; // A nonnegative shard number
     int64 realmNum = 2; // A nonnegative realm number
     int64 scheduleNum = 3; // A nonnegative schedule number
+}
+
+/**
+ * Possible Token Types (IWA Compatibility).
+ * Apart from fungible and non-fungible, Tokens can have either a common or unique representation. This distinction might seem subtle, but it is important when considering
+ * how tokens can be traced and if they can have isolated and unique properties.
+ */
+enum TokenType {
+    /**
+     * Interchangeable value with one another, where any quantity of them has the same value as another equal quantity if they are in the same class.
+     * Share a single set of properties, not distinct from one another. Simply represented as a balance or quantity to a given Hedera account.
+     */
+    FUNGIBLE_COMMON = 0;
+    /**
+     * Unique, not interchangeable with other tokens of the same type as they typically have different values.
+     * Individually traced and can carry unique properties (e.g. serial number).
+     */
+    NON_FUNGIBLE_UNIQUE = 1;
+}
+
+
+/**
+ * Possible FeeData Object SubTypes. Supplementary to the main HederaFunctionality Type.
+ * When not explicitly specified, DEFAULT is used.
+ */
+enum SubType {
+    DEFAULT = 0;
+    TOKEN_FUNGIBLE_COMMON = 1;
+    TOKEN_NON_FUNGIBLE_UNIQUE = 2;
+}
+
+/**
+ * Possible Token Supply Types (IWA Compatibility).
+ * Indicates how many tokens can have during its lifetime.
+ */
+enum TokenSupplyType {
+    INFINITE = 0; // Indicates that tokens of that type have an upper bound of Long.MAX_VALUE.
+    FINITE = 1; // Indicates that tokens of that type have an upper bound of maxSupply, provided on token creation.
 }
 
 /* Possible Freeze statuses returned on TokenGetInfoQuery or CryptoGetInfoResponse in TokenRelationship */
@@ -162,7 +208,7 @@ message KeyList {
 
 /* A Signature corresponding to a Key. It is a sequence of bytes holding a public key signature from one of the three supported systems (ed25519, RSA-3072,  ECDSA with p384). Or, it can be a list of signatures corresponding to a single threshold key. Or, it can be the ID of a smart contract instance, which is authorized to act as if it had a key. If an account has an ed25519 key associated with it, then the corresponding private key must sign any transaction to transfer cryptocurrency out of it. If it has a smart contract ID associated with it, then that smart contract is allowed to transfer cryptocurrency out of it. The smart contract doesn't actually have a key, and  doesn't actually sign a transaction. But it's as if a virtual transaction were created, and the smart contract signed it with a private key. A key can also be a "threshold key", which means a list of M keys, any N of which must sign in order for the threshold signature to be considered valid. The keys within a threshold signature may themselves be threshold signatures, to allow complex signature requirements (this nesting is not supported in the currently, but will be supported in a future version of API). If a Signature message is missing the "signature" field, then this is considered to be a null signature. That is useful in cases such as threshold signatures, where some of the signatures can be null.
  * The definition of Key uses mutual recursion, so it allows nesting that is arbitrarily deep. But the current API only accepts Key messages up to 3 levels deep, such as a list of threshold keys, each of which is a list of primitive keys. Therefore, the matching Signature will have the same limitation. This restriction may be relaxed in future versions of the API, to allow deeper nesting.
- * This message is deprecated and succeeded by SignaturePair and SignatureMap messages. 
+ * This message is deprecated and succeeded by SignaturePair and SignatureMap messages.
  */
 message Signature {
     option deprecated = true;
@@ -282,6 +328,9 @@ enum HederaFunctionality {
     ScheduleDelete = 71; // Delete Scheduled Transaction
     ScheduleSign = 72; // Sign Scheduled Transaction
     ScheduleGetInfo = 73; // Get Scheduled Transaction Information
+    TokenGetAccountNftInfo = 74; // Get Token Account Nft Information
+    TokenGetNftInfo = 75; // Get Token Nft Information
+    TokenGetNftInfos = 76; // Get Token Nft List Information
 }
 
 /*
@@ -307,7 +356,9 @@ message TransactionFeeSchedule {
     // A particular transaction or query
     HederaFunctionality hederaFunctionality = 1;
     // Resource price coefficients
-    FeeData feeData = 2;
+    FeeData feeData = 2 [deprecated=true];
+    // Resource price coefficients. Supports subtype price definition.
+    repeated FeeData fees = 3;
 }
 
 /*
@@ -320,6 +371,8 @@ message FeeData {
     FeeComponents networkdata = 2;
     // Fee paid to the network for providing the service associated with the transaction; for instance, storing a file
     FeeComponents servicedata = 3;
+    // SubType distinguishing between different types of FeeData, correlating to the same HederaFunctionality
+    SubType subType = 4;
 }
 
 /*
@@ -338,7 +391,7 @@ message CurrentAndNextFeeSchedule {
     FeeSchedule nextFeeSchedule = 2; // Contains next Fee Schedule
 }
 
-/* 
+/*
 Contains the IP address and the port representing a service endpoint of a Node in a network. Used to reach the Hedera API and submit transactions to the network.
 */
 message ServiceEndpoint {
@@ -346,12 +399,12 @@ message ServiceEndpoint {
     int32 port = 2; // The port of the node
 }
 
-/* 
-The data about a node, including its service endpoints and the Hedera account to be paid for services 
-provided by the node (that is, queries answered and transactions submitted.) 
+/*
+The data about a node, including its service endpoints and the Hedera account to be paid for services
+provided by the node (that is, queries answered and transactions submitted.)
 
 If the <tt>serviceEndpoint</tt> list is not set, or empty, then the endpoint given by the (deprecated)
-<tt>ipAddress</tt> and <tt>portno</tt> fields should be used. 
+<tt>ipAddress</tt> and <tt>portno</tt> fields should be used.
 
 All fields are populated in the 0.0.102 address book file while only fields that start with # are populated in the 0.0.101 address book file.
 */
@@ -377,10 +430,15 @@ message NodeAddressBook {
     repeated NodeAddress nodeAddress = 1; // Metadata of all nodes in the network
 }
 
+/* Hedera follows semantic versioning (https://semver.org/) for both the HAPI protobufs and the Services software.
+This type allows the <tt>getVersionInfo</tt> query in the <tt>NetworkService</tt> to return the deployed versions
+of both protobufs and software on the node answering the query. */
 message SemanticVersion {
-   int32 major = 1; // Increases with incompatible API changes
-   int32 minor = 2; // Increases with backwards-compatible new functionality
-   int32 patch = 3; // Increases with backwards-compatible bug fixes
+    int32 major = 1; // Increases with incompatible API changes
+    int32 minor = 2; // Increases with backwards-compatible new functionality
+    int32 patch = 3; // Increases with backwards-compatible bug fixes
+    string pre = 4; // A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers (https://semver.org/#spec-item-9); so given a semver 0.14.0-alpha.1+21AF26D3, this field would contain 'alpha.1'
+    string build = 5; // Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version (https://semver.org/#spec-item-10); so given a semver 0.14.0-alpha.1+21AF26D3, this field would contain '21AF26D3'
 }
 
 message Setting {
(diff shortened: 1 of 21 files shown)


services/BasicTypes.proto
```diff
@@ -99,6 +99,53 @@ message TokenTransferList {
     repeated NftTransfer nftTransfers = 3; // Applicable to tokens of type NON_FUNGIBLE_UNIQUE. Multiple list of NftTransfers, each of which has a sender and receiver account, including the serial number of the NFT
 }
 
+/* Representing the percentage of tokens to be transfers according to the TokenUnitFee*/
```
ljianghedera, line 102: Should it be `to be transferred` or `to transfer`?

qnswirlds (changes requested):
Discussed via Slack. These protobufs changes should be ceased here. There have been misinterpretations of details in the HIP-18.

services/BasicTypes.proto
```diff
@@ -99,6 +99,53 @@ message TokenTransferList {
...
+    uint64 minimumValueToCollect = 2;
+}
+
+/* A list of accounts and amounts that are involved in a Token's HbarFee*/
+message CustomHbarFees {
+    TransferList customHbarFees = 1;
+}
+
+/* A list of tokenIds and amounts representing the transferred out (negative) or into (positive) amounts according to a Token's TokenUnitFee*/
+message CustomTokenUnitFees {
```
ljianghedera, line 120: Is this message just for fees or for multiple token transfers?

anighanta: This is for adding the + / - amounts in token units which resulted because of custom fees, in the transaction records.

services/TokenCreate.proto
```diff
@@ -60,4 +60,5 @@ message TokenCreateTransactionBody {
     TokenType tokenType = 17; // IWA compatibility. Specifies the token type. Defaults to FUNGIBLE_COMMON
     TokenSupplyType supplyType = 18; // IWA compatibility. Specified the token supply type. Defaults to INFINITE
     int64 maxSupply = 19; // IWA Compatibility. Depends on TokenSupplyType. For tokens of type FUNGIBLE_COMMON - the maximum number of tokens that can be in circulation. For tokens of type NON_FUNGIBLE_UNIQUE - the maximum number of NFTs (serial numbers) that can be minted. This field can never be changed!
+    CustomFees customFees = 20; // Custom fee in Token Units/hbar that needs to be paid to the specified receiver on a TokenTransfer
```
ljianghedera, line 63: Based on this morning's discussion, there should be a new key for custom fees enhancement.

services/TokenGetInfo.proto
```diff
@@ -60,6 +60,7 @@ message TokenInfo {
     TokenType tokenType = 19; // The token type
     TokenSupplyType supplyType = 20; // The token supply type
     int64 maxSupply = 21; // For tokens of type FUNGIBLE_COMMON - The Maximum number of fungible tokens that can be in circulation. For tokens of type NON_FUNGIBLE_UNIQUE - the maximum number of NFTs (serial numbers) that can be in circulation
+    CustomFees customFees = 22; // Custom fee in Token Units/hbar that needs to be paid to the specified receiver on a TokenTransfer
```
ljianghedera, line 63: This also need to return the new key for token custom fees.

services/BasicTypes.proto
```diff
@@ -99,6 +100,50 @@ message TokenTransferList {
...
+message Fraction {
+    uint64 numerator = 1; // numerator of the fraction
+    uint64 denominator = 2; // denominator of the fraction
+}
+
+/* Fraction of the Token to be collected along with its floor if the transfer amount is low*/
+message FractionalFee {
+    Fraction fractionOfUnitsToCollect = 1; // percentage of the Tokens involved in the transfer to be charged as fee
+    uint64 minimumUnitsToCollect = 2; // minimum token units to be charged as fee
+    google.protobuf.UInt64Value maximumUnitsToCollect = 3; // maximum token units to be charged as fee
```
abhishek-hedera, line 113: HIP 18 doesn't define a maximum unit for collection yet. If this is enabled, it would result in less token units in royalty percentage. For eg, a 10% royalty on 100 token transfer would lead to 10 tokens as custom fees, whereas transferring an amount of 100000 with a maximum limit of 1000 would lead to only 1% of total transfer as custom fees.

services/BasicTypes.proto
```diff
@@ -99,6 +100,50 @@ message TokenTransferList {
...
+
+/* A list of accounts and amounts that are involved in a Token's HbarFee */
+message CustomHbarFees {
+    TransferList customHbarFees = 1;
+}
+
+/* A list of tokenIds and amounts representing the transferred out (negative) or into (positive) amounts according to a Token's TokenUnitFee */
+message CustomTokenUnitFees {
+    repeated TokenTransferList customTokenUnitFees = 1;
+}
```
abhishek-hedera, line 145: Could add comments to the fields similar to other messages.

qnswirlds:
@rbair23, @SimiHunjan, @donaldthibeau, @steven-sheehy please review the protobufs changes for HIP-18. Thanks!

services/CryptoTransfer.proto
```diff
@@ -31,6 +31,8 @@ import "BasicTypes.proto";
 If any sender account fails to have sufficient hbars, then the entire transaction fails, and none of those transfers occur, though the transaction fee is still charged. This transaction must be signed by the keys for all the sending accounts, and for any receiving accounts that have receiverSigRequired == true. The signatures are in the same order as the accounts, skipping those accounts that don't need a signature.
 */
 message CryptoTransferTransactionBody {
-    TransferList transfers = 1;
-    repeated TokenTransferList tokenTransfers = 2;
+    TransferList transfers = 1;  // For hbar transfer;
```
SimiHunjan, line 34: Not sure if the ";" was intended after the word transfer

qnswirlds: No, @SimiHunjan, we will fix it with some other typos that we noted already. Thanks!

services/BasicTypes.proto
```diff
@@ -99,6 +99,12 @@ message TokenTransferList {
     repeated NftTransfer nftTransfers = 3; // Applicable to tokens of type NON_FUNGIBLE_UNIQUE. Multiple list of NftTransfers, each of which has a sender and receiver account, including the serial number of the NFT
 }
 
+/* Representing the percentage of tokens to be transferred as part of the CustomFee */
+message Fraction {
+    uint64 numerator = 1; // numerator of the fraction
+    uint64 denominator = 2; // denominator of the fraction
+}
```
SimiHunjan, line 106: The first letter of the word in the comment should be in uppercase to match the rest

services/CustomFees.proto
```diff
@@ -0,0 +1,71 @@
...
+
+/* Custom fees that are charged for a CryptoTransfer. */
+message CustomFeesCharged {
+  repeated CustomFeeCharged customFeesCharged = 1; // all custom fees charged 
+}
+
+/* Custom fees associated to a token */
+message CustomFees {
+  repeated CustomFee customFees = 1; // all custom fees for a token 
+}
```
SimiHunjan, line 71: The first letter of the word in the comments should be in uppercase to match the rest

services/CustomFees.proto
```diff
@@ -0,0 +1,71 @@
...
+
+import "BasicTypes.proto";
+import "google/protobuf/wrappers.proto";
+
+/* A fraction of the units to be collected from the payer of a CryptoTransfer 
+that transfers units of a token. The amount charged will never be
+less than the stated minimumUnitsToCollect, and never greater than 
+the stated maximumUnitsToCollect (if given). */
+message FractionalFee {
+  Fraction fractionOfUnitsToCollect = 1; // percentage of the Tokens involved in the transfer to be charged as fee
```
steven-sheehy, line 33: This is inefficient to transfer and store, including for the mirror node where we have to store it indefinitely. Why can't we just use `double` to store the fraction? Internally, I'm sure you'll calculate `fractionOfUnitsToCollect. numerator / fractionOfUnitsToCollect.denominator` to actually utilize it, so you're not losing any granularity and you might as well push that calculation to the client.

services/ResponseCode.proto
```diff
@@ -190,7 +192,7 @@ enum ResponseCodeEnum {
   TOKEN_HAS_NO_WIPE_KEY = 181; // Wipe key is not set on token
   INVALID_TOKEN_MINT_AMOUNT = 182; // The requested token mint amount would cause an invalid total supply
   INVALID_TOKEN_BURN_AMOUNT = 183; // The requested token burn amount would cause an invalid total supply
-  TOKEN_NOT_ASSOCIATED_TO_ACCOUNT = 184; // A required token-account relationship is missing
+  TOKEN_NOT_ASSOCIATED_TO_ACCOUNT = 184;  // A required token-account relationship is missing
```
steven-sheehy, line 195: nit: Extra space added that's not there for the other rows.

services/TokenCreate.proto
```diff
@@ -60,4 +61,6 @@ message TokenCreateTransactionBody {
...
+    Key customFeesKey = 21; // The key with which to charge customFees on token transfers. If empty, customFees are not required. If the customFee is added, this tokenCreate transaction needs to be signed by the customFeesKey.
```
steven-sheehy, line 65: Perhaps it makes sense to push `customFeesKey` into `CustomFees` since everywhere you have one you have the other and since their functionality is interdependent. This would be more encapsulated.

services/TokenUpdate.proto
```diff
@@ -53,4 +54,6 @@ message TokenUpdateTransactionBody {
     Duration autoRenewPeriod = 11; // The new interval at which the auto-renew account will be charged to extend the token's expiry.
     Timestamp expiry = 12; // The new expiry time of the token. Expiry can be updated even if admin key is not set. If the provided expiry is earlier than the current token expiry, transaction wil resolve to INVALID_EXPIRATION_TIME
     google.protobuf.StringValue memo = 13; // If set, the new memo to be associated with the token (UTF-8 encoding max 100 bytes)
+    CustomFees customFees = 14; // Custom fee in Token Units/hbar that needs to be paid to the specified receiver on a TokenTransfer
```
ijungmann, line 57: Can a custom fee be deleted or updated, or only created, and if so how will that look in this object?

services/CustomFees.proto
```diff
@@ -0,0 +1,71 @@
...
+  uint64 minimumUnitsToCollect = 2; // minimum token units to be charged as fee
+  google.protobuf.UInt64Value maximumUnitsToCollect = 3; // maximum token units to be charged as fee
+}
+
+/* A fixed custom fee to be collected during a CryptoTransfer that transfers units of a token */
+message FixedFee {
+  uint64 unitsToCollect = 1; // units of either hbar/token to be collected
```
steven-sheehy, line 40: Isn't `amount` the more succinct and generally used term?

services/CustomFees.proto
```diff
@@ -0,0 +1,71 @@
...
+import "google/protobuf/wrappers.proto";
+
+/* A fraction of the units to be collected from the payer of a CryptoTransfer 
+that transfers units of a token. The amount charged will never be
+less than the stated minimumUnitsToCollect, and never greater than 
+the stated maximumUnitsToCollect (if given). */
+message FractionalFee {
+  Fraction fractionOfUnitsToCollect = 1; // percentage of the Tokens involved in the transfer to be charged as fee
+  uint64 minimumUnitsToCollect = 2; // minimum token units to be charged as fee
+  google.protobuf.UInt64Value maximumUnitsToCollect = 3; // maximum token units to be charged as fee
```
steven-sheehy, line 35: Having `OfUnitsToCollect` appended to every field seems redundant. Fees by their nature are always something to be collected. And I think `amount` is a more appropriate term. Or you can just drop the repeated suffix since it's already within the context of `FractionalFee` wrapper to make them `fraction`, `minimum` and `maximum`.

services/CustomFees.proto
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+import "BasicTypes.proto";
+import "google/protobuf/wrappers.proto";
+
+/* A fraction of the units to be collected from the payer of a CryptoTransfer 
+that transfers units of a token. The amount charged will never be
+less than the stated minimumUnitsToCollect, and never greater than 
+the stated maximumUnitsToCollect (if given). */
+message FractionalFee {
```
steven-sheehy, line 32: q: Why doesn't FractionalFee need a `tokenID` to differentiate between units of hbar vs token?

steven-sheehy (changes requested):

services/CustomFees.proto
```diff
@@ -0,0 +1,71 @@
...
+  AccountID feeCollector = 3; // Fee collector of this custom fee.
+}
+
+/* A custom transfer fee in token units or hbar that was computed, based
+on a CustomFee type, to be charged for a CryptoTransfer. */
+message CustomFeeCharged {
+  uint64 unitsCharged = 1; // units of either hbar/token that were charged for this CryptoTransfer 
+  TokenID tokenId = 2; // denominating token type. If this is empty the denomination type is hbar.
+  AccountID feeCollector = 3; // Fee collector of this custom fee.
+}
```
ijungmann, line 61: The payer/sender account id would always be the accountId in the transactionID, correct?

services/TransactionRecord.proto
```diff
@@ -46,4 +47,5 @@ message TransactionRecord {
     TransferList transferList = 10; // All hbar transfers as a result of this transaction, such as fees, or transfers performed by the transaction, or by a smart contract it calls, or by the creation of threshold records that it triggers.
     repeated TokenTransferList tokenTransferLists = 11; // All Token transfers as a result of this transaction
     ScheduleID scheduleRef = 12; // Reference to the scheduled transaction ID that this transaction record represent
+    CustomFeesCharged customFeesCharged = 13; // The custom fees that were charged, and must have been paid if the transaction status resolved to SUCCESS
```
ijungmann, line 50: A little confused by the comment, is this saying that the field could be filled with "potential charges that were not paid" if the transaction was not successful?

tinker-michaelj: Yes, these are just the required fees that are charged.

If the payer cannot afford them, instead of `SUCCESS` we would get `INSUFFICIENT_PAYER_BALANCE_FOR_CUSTOM_FEE`.

Let me think about how to word this more clearly.

services/CustomFees.proto
tinker-michaelj: @ijungmann That's correct....but as below, this list just specifies the fees which must be paid for the `CryptoTransfer` to resolve to `SUCCESS`. 

**Only** the `transferList` and `tokenTransfersList` include actual balance changes.

services/CustomFees.proto
tinker-michaelj: True, will change.

services/CustomFees.proto
ijungmann: Ah, okay, I misunderstood this, so fee charges will appear in the actual transfer lists.  That makes life a lot easier for the mirror node, cool.

tinker-michaelj: @ijungmann What do you think about instead,
```
/* A transfer fee assessed to a CryptoTransfer that moves units of a token with a custom fee schedule. Unless the payer can afford to pay all assessed fees, the transfer will not succeed.  */
message AssessedCustomFee {
  uint64 amount = 1; // units of ℏ or a denominating HTS token 
  TokenID tokenId = 2; // the denominating token; units are ℏ if empty
  AccountID feeCollectorAccountId = 3; // the account that will will receive the fee 
}
```

services/TransactionRecord.proto
xin-hedera: Is it better to use 
```suggestion
    repeated CustomFeeCharged customFeesCharged = 13; // The custom fees that were charged, and must have been paid if the transaction status resolved to SUCCESS
```

with the wrapper, the generated java code to get the charged fees will be:

```java
  transactionRecord.getCustomFeesCharged().getCustomFeesChargedList()
```

services/CustomFees.proto
xin-hedera: Can we just use `uint64` for `maximumUnitsToCollect` too? value 0 means no max set.

xin-hedera (changes requested):

services/CustomFees.proto
xin-hedera: looks like `tokenID` should be moved to `CustomFee`

services/CustomFees.proto
tinker-michaelj: Leemon has set a strict ban on all floating point calculations because different JVMs can do floating point arithmetic differently.

services/CustomFees.proto
tinker-michaelj: @qnswirlds has cleared with the HIP-18 authors that a fractional fee will inherit the denomination of its parent token. (At least in the initial implementation.)

Am clarifying this in the comments.

services/CustomFees.proto
tinker-michaelj: Since we require fully replacing the `CustomFees` field on an update, this makes sense. 👍

services/BasicTypes.proto
tinker-michaelj: Done.

services/CustomFees.proto
tinker-michaelj: Done.

services/BasicTypes.proto
tinker-michaelj: Fixed.

ljianghedera (approved):
LGTM

services/CustomFees.proto
```diff
@@ -0,0 +1,65 @@
...
+message FractionalFee {
+  Fraction fractional_amount = 1; // The fraction of the transferred units to assess as a fee
+  sint64 minimum_amount = 2; // The minimum amount to assess
+  sint64 maximum_amount = 3; // The maximum amount to assess (zero implies no maximum)
+}
+
+/* A fixed number of units (hbar or token) to asssess as a fee during a CryptoTransfer 
+that transfers units of the token to which this fixed fee is attached. */
+message FixedFee {
+  sint64 amount = 1; // The number of units to assess as a fee
```
xin-hedera, line 39: Q: can `amount` be negative? If not, `sint64` does not provide benefit and `int64` is more appropriate. The same applies to most other `sint64` fields.

tinker-michaelj: Since Java does not have an unsigned integral type, we can't actually support the full range of values `[0, 2^64 - 1]`, so this is a more honest contract.

xin-hedera: understood. `int64` and `sint64` are both signed, the difference as in protobuf afaik is `sint64` is encoded in a zigzag way which reduces encoded message size for negative value. so if `amount` is always non-negative, `int64` suites better v.s. `sint64`.

tinker-michaelj: Ah I see, changed all `sint64` to `int64`. 👍

steven-sheehy (approved):
LGTM

qnswirlds (commented):
> We've been following camel case for variables. Please change all of those variables in snake_case to camelCase. Thanks!

There was a concern that camel case is non-standard for protobuf and doesn't generate the desired names in non-Java languages. So moving forward, we will standardize to snake case.

services/CustomFees.proto
```diff
@@ -0,0 +1,65 @@
...
+/* A custom transfer fee that was assessed during handling of a CryptoTransfer. */
+message AssessedCustomFee {
+  sint64 amount = 1; // The number of units assessed for the fee
+  TokenID token_id = 2; // The denomination of the fee; taken as hbar if left unset
+  AccountID fee_collector_account_id = 3; // The account to receive the assessed fee
+}
+
+/* The custom fees attached to a token; may be marked as mutable or immutable. */
+message CustomFees {
+  bool can_update_with_admin_key = 1; // Whether the fees can be updated with the token's admin key; once this is false, it can never be made true again
```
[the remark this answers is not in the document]
tinker-michaelj: A token without an admin key is completely immutable.

But even a mutable token may want to "commit" to never changing/increasing its custom fees.

The token can do this by setting `can_update_with_admin_key=false`; while the rest of its fields can still be updated with the admin key.

services/CustomFees.proto
ijungmann: That works!

services/TokenUpdate.proto
ijungmann: @tinker-michaelj I think this question got passed over, still relevant.

xin-hedera (approved):
LGTM

services/TokenUpdate.proto
ijungmann: I do see the comments for updating the fees for a token , to clarify can one actually update a particular fee, or is it just "on an update, all existing fees are cleared and the new fees replace them."

qnswirlds (approved):
LGTM!

tinker-michaelj (approved):
LGTM

ljianghedera (approved):
LGTM

PR #60, opened 2021-07-28
Update ResponseCode.proto

SimiHunjan:
**Description**:

In the original implementation, the transaction fee payer account was responsible for paying the custom token fees. If the transaction fee-paying account did not have enough of the token to pay for the fee then INSUFFICIENT_PAYER_BALANCE_FOR_CUSTOM_FEE response was returned.

With the new implementation, the sender account is now responsible for paying the custom fees. When I see 'payer' in a response code I am thinking about the transaction fee payer account and not the sending account.

**Related issue(s)**:

Fixes #

https://github.com/hashgraph/hedera-services/issues/1825
 
**Notes for reviewer**:

**Checklist**


SimiHunjan:
submitting a new one

PR #62, opened 2021-07-28
Update response code

SimiHunjan:
**Description**:

In the original implementation, the transaction fee payer account was responsible for paying the custom token fees. If the transaction fee-paying account did not have enough of the token to pay for the fee then INSUFFICIENT_PAYER_BALANCE_FOR_CUSTOM_FEE response was returned.

With the new implementation, the sender account is now responsible for paying the custom fees. When I see 'payer' in a response code I am thinking about the transaction fee payer account and not the sending account.

Solution
Update the response code to specify the sender account does not have enough token balance to pay for the custom fee.
INSUFFICIENT_SENDER_ACCOUNT_BALANCE_FOR_CUSTOM_FEE

**Related issue(s)**:

https://github.com/hashgraph/hedera-services/issues/1825

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/services/ResponseCode.proto
+++ b/services/ResponseCode.proto
@@ -240,7 +240,7 @@ enum ResponseCodeEnum {
   BATCH_SIZE_LIMIT_EXCEEDED = 228; // Repeated operations count exceeds the limit
   INVALID_QUERY_RANGE = 229; // The range of data to be gathered is out of the set boundaries
   FRACTION_DIVIDES_BY_ZERO = 230; // A custom fractional fee set a denominator of zero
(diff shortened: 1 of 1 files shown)


SimiHunjan:
@tinker-michaelj let me know if this is correct with your suggestion of marking response code 231 as deprecated and adding a new one

tinker-michaelj:
> @tinker-michaelj let me know if this is correct with your suggestion of marking response code 231 as deprecated and adding a new one

Yep, this makes sense to me!

tinker-michaelj (approved):
LGTM

steven-sheehy (approved):

PR #69, opened 2021-08-04
068 d allow fractional fee charged to sender or receiver

### LINKED ISSUE #68: Allow fractional fees to be charged to _either_ sender or receiver
### Problem

Please refer to description in services issue #1926: https://github.com/hashgraph/hedera-services/issues/1926

### Solution

```
message FractionalFee {
  Fraction fractional_amount = 1; // The fraction of the transferred units to assess as a fee
  int64 minimum_amount = 2; // The minimum amount to assess
  int64 maximum_amount = 3; // The maximum amount to assess (zero implies no maximum)
  bool netOfTransfers = 4; // If true, assesses the fee to the sender, so the receiver gets the full amount from the token transfer list, and the sender is charged an additional fee; if false, the receiver does NOT get the full amount, but only what is left over after paying the fractional fee
}
```

### Alternatives

_No response_

ljianghedera:
**Description**:

This PR adds one bool field to allow the user to be charged to sender or receiver.

**Related issue(s)**:

Fixes #68

**Checklist**


anighanta (dismissed):

services/CustomFees.proto
```diff
@@ -34,6 +34,7 @@ message FractionalFee {
   Fraction fractional_amount = 1; // The fraction of the transferred units to assess as a fee
   int64 minimum_amount = 2; // The minimum amount to assess
   int64 maximum_amount = 3; // The maximum amount to assess (zero implies no maximum)
+  bool netOfTransfers = 4; // If true, assesses the fee to the sender, so the receiver gets the full amount from the token transfer list, and the sender is charged an additional fee; if false, the receiver does NOT get the full amount, but only what is left over after paying the fractional fee
```
abhishek-hedera, line 37: This should be in snake case.

qnswirlds (commented):
I think we should merge to `develop` instead of to `main`.

PR #71, opened 2021-08-08
Add RoyaltyFee custom fee type

tinker-michaelj:
**Description**:
- Define the `RoyaltyFee` message needed by https://github.com/hashgraph/hedera-services/issues/1935; please see that issue for details.

### DIFF
--- a/services/ResponseCode.proto
+++ b/services/ResponseCode.proto
@@ -251,7 +251,7 @@ enum ResponseCodeEnum {
   CUSTOM_FEE_MUST_BE_POSITIVE = 239; // Only positive fees may be assessed at this time
   TOKEN_HAS_NO_FEE_SCHEDULE_KEY = 240; // Fee schedule key is not set on token
   CUSTOM_FEE_OUTSIDE_NUMERIC_RANGE = 241; // A fractional custom fee exceeded the range of a 64-bit signed integer
-  INVALID_CUSTOM_FRACTIONAL_FEES_SUM = 242; // The sum of all custom fractional fees must be strictly less than 1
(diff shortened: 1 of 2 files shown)


xin-hedera:
it's possible when the fungible value in exchange for the `nft` is low and multiplying it with the fraction will result in 0 royalty fee charged, is it desired or should we add a `minimum`?

services/ResponseCode.proto
```diff
@@ -251,7 +251,7 @@ enum ResponseCodeEnum {
   CUSTOM_FEE_MUST_BE_POSITIVE = 239; // Only positive fees may be assessed at this time
   TOKEN_HAS_NO_FEE_SCHEDULE_KEY = 240; // Fee schedule key is not set on token
   CUSTOM_FEE_OUTSIDE_NUMERIC_RANGE = 241; // A fractional custom fee exceeded the range of a 64-bit signed integer
-  INVALID_CUSTOM_FRACTIONAL_FEES_SUM = 242; // The sum of all custom fractional fees must be strictly less than 1
```
qnswirlds, line 254: If we still keep the `FractionalFee`, I think we should also keep this `INVALID_CUSTOM_FRACTIONAL_FEES_SUM` response code.

tinker-michaelj:
> it's possible when the fungible value in exchange for the `nft` is low and multiplying it with the fraction will result in 0 royalty fee charged, is it desired or should we add a `minimum`?

Interesting question. 🤔 It seems this minimum would need to be specific to each of the fungible tokens the fee collector accepts, which could make the feature somewhat elaborate.

PR #73, opened 2021-08-14
Fix/comments

danielakhterov:
**Description**:

Update comments to follow the same style; that style being Javadoc style:
```java
/**
 * <comment>
 */
```

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**

- [X] Documented (Code comments, README, etc.)
- [X] Tested (unit, integration, etc.)


steven-sheehy (commented):
Glanced through and it looked fine. Just clarifying if this change was done using an automated process? If so, that gives me high confidence that there wouldn't be any copy and mistake errors.

danielakhterov:
@steven-sheehy the process I used was to use VIM's built in comment formatter and macros. I would simply select the region to format and then hit a macro which would convert `//` into `*`, put `/**` before the selected region, and ` */` at the end of the selected region. Of course not all comments were formatted originally to start with `//` so I had to properly choose which macro to use. I've also use a _little_ regex to format one line comments such as `/* .* */`. So this process wasn't completely automated, but it wasn't completely manual with copy and pasting either.

danielakhterov:
Seems like I got a bit of rebasing to do.

steven-sheehy (approved):
LGTM

PR #90, opened 2021-09-14
Remove deprecated freeze transaction fields and add `FreezeType` enum

tinker-michaelj:
**Description**:
- Remove deprecated fields from `FreezeTransactionBody`.
- Add `FreezeType` enum to control handling of `Freeze` transactions.
- Add two response codes for freeze-specific failure modes.

### DIFF
--- a/services/response_code.proto
+++ b/services/response_code.proto
@@ -1005,7 +1005,18 @@ enum ResponseCodeEnum {
   EXISTING_AUTOMATIC_ASSOCIATIONS_EXCEED_GIVEN_LIMIT = 263;
 
   /**
-   * Cannot set the number of automatic associations for an account more than the maximum allowed token associations <tt>tokens.maxPerAccount</tt>
+   * Cannot set the number of automatic associations for an account more than the maximum allowed 
+   * token associations <tt>tokens.maxPerAccount</tt>.
    */
   REQUESTED_NUM_AUTOMATIC_ASSOCIATIONS_EXCEEDS_ASSOCIATION_LIMIT = 264;
+
+  /** 
+   * The update file in a freeze transaction body must exist.
+   */
+  FREEZE_UPDATE_FILE_DOES_NOT_EXIST = 265;
+
+  /** 
+   * The hash of the update file in a freeze transaction body must match the in-memory hash.
+   */
+  FREEZE_UPDATE_FILE_HASH_DOES_NOT_MATCH = 266;
 }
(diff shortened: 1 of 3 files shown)


nathanklick (approved):

mehcode:
Please do not remove any fields from protobufs.

tinker-michaelj:
> Please do not remove any fields from protobufs.
> 
> If a field is removed, mirror nodes need to resort to tricks to pull out data from old transactions.

Checked with Nathan and Steven that `FreezeTransactionBody`s aren't parsed by Mirror Nodes, as they are very unlikely to be of any general interest.

So will just remove and reserve the fields to eliminate risk of accidental use by DevOps when freezing the network.

mehcode:
Kabuto does parse them as I believe it's adamant to the spirit of decentralization to make **any** transaction available for viewing, including its properties. 

If this PR is merged as-is, I'll need to add a patch file in the Rust SDK to re-add them. I'd really rather not do that.

tinker-michaelj:
> Kabuto does parse them as I believe it's adamant to the spirit of decentralization to make **any** transaction available for viewing, including its properties.
> 
> If this PR is merged as-is, I'll need to add a patch file in the Rust SDK to re-add them. I'd really rather not do that.

Re-restored fields, will just add mainnet code to reject any `Freeze` using these fields.

nathanklick (dismissed):

qnswirlds (approved):

services/response_code.proto
```diff
@@ -1025,4 +1026,65 @@ enum ResponseCodeEnum {
...
+   */
+  UPDATE_FILE_HASH_CHANGED_SINCE_PREPARE_UPGRADE = 272;
+
+  /** 
+   * The given freeze start time was in the (consensus) past.
+   */
+  FREEZE_START_TIME_MUST_BE_FUTURE = 273;
+
+  /** 
+   * The prepared update file cannot be updated or appended util either the upgrade has
```
SimiHunjan, line 1062: util -> until

services/response_code.proto
```diff
@@ -1025,4 +1026,65 @@ enum ResponseCodeEnum {
...
+
+  /** 
+   * If an NMT upgrade has been prepared, the following operation must be a FREEZE_UPGRADE.
+   * (To issue a FREEZE_ONLY, submit a FREEZE_ABORT first.)
+   */
+  FREEZE_UPGRADE_IN_PROGRESS = 276;
+
+  /** 
+   * If an NMT upgrade has been prepared, the subsequent FREEZE_UPGRADE transaction must 
+   * confirm the id of the to be used in the upgrade.
```
SimiHunjan, line 1081: the id of the what?

qnswirlds (approved):

PR #94, opened 2021-09-24
Fix/clarify some comments

tinker-michaelj:
**Description**:
- Clarify how to interpret the hashes in `NodeAddress`.
- Fix the actual value of [`hedera.transaction.maxValidDuration=180`](https://github.com/hashgraph/hedera-services/blob/master/hedera-node/src/main/resources/bootstrap.properties#L57) in use on production networks.

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1180,8 +1180,8 @@ message NodeAddress {
 
     /**
      * Hex encoding of the node's X509 RSA public key used to sign stream files (e.g., record 
-     * stream files). That is, this field is a string of hexadecimal characters which, translated 
-     * to binary, are the public key's DER encoding.  
+     * stream files). Specifically, this field is a string of hexadecimal characters which, 
+     * translated to binary, are the public key's DER encoding.  
      */
     string RSA_PubKey = 4;
 
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1179,7 +1179,9 @@ message NodeAddress {
     bytes memo = 3 [deprecated=true];
 
     /**
-     * The node's hex-encoded X509 RSA public key
+     * Hex encoding of the node's X509 RSA public key used to sign stream files (e.g., record 
+     * stream files). That is, this field is a string of hexadecimal characters which, translated 
+     * to binary, are the public key's DER encoding.  
      */
     string RSA_PubKey = 4;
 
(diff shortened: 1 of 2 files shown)


services/basic_types.proto
```diff
@@ -1179,7 +1179,9 @@ message NodeAddress {
     bytes memo = 3 [deprecated=true];
 
     /**
-     * The node's hex-encoded X509 RSA public key
+     * Hex encoding of the node's X509 RSA public key used to sign stream files (e.g., record 
+     * stream files). That is, this field is a string of hexadecimal characters which, translated
```
qnswirlds, line 1183: Remove `That is, ` or use `Specifically, ` as in the other change.

tinker-michaelj: Used "Specifically, "...

qnswirlds (approved):

services/basic_types.proto
```diff
@@ -1194,7 +1196,10 @@ message NodeAddress {
     AccountID nodeAccountId = 6;
 
     /**
-     * # The hex-encoded SHA-384 hash of the X509 cert used to encrypt gRPC traffic to the node
+     * # Hex encoding of the SHA-384 hash of the TLS cert the node uses to encrypt gRPC
```
steven-sheehy, line 1199: This description is bit obtuse and the second sentence duplicates much of the first. Would suggest replacing the first sentence with something that just talks about the cert, keeping the second sentence, and adding a note about what this field can be used for.

```
Every node has a TLS certificate in PEM format that it uses to encrypt gRPC traffic. This field is a string of hexadecimal characters which, translated to binary, are the SHA-384 hash of the UTF-8 NFKD representation of the node's certificate. Its value can be used to verify the node's certificate it presents during TLS negotiations.
```

steven-sheehy (changes requested):

services/basic_types.proto
```diff
@@ -1194,7 +1196,10 @@ message NodeAddress {
     AccountID nodeAccountId = 6;
 
     /**
-     * # The hex-encoded SHA-384 hash of the X509 cert used to encrypt gRPC traffic to the node
+     * # Hex encoding of the SHA-384 hash of the TLS cert the node uses to encrypt gRPC 
+     * traffic (in UTF-8 NKFD). Specifically, this field is a string of hexadecimal characters
```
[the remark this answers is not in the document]
tinker-michaelj: Oops! Fixed.

services/response_code.proto
```diff
@@ -62,7 +62,9 @@ enum ResponseCodeEnum {
   INVALID_TRANSACTION_START = 5;
 
   /**
-   * valid transaction duration is a positive non zero number that does not exceed 120 seconds
+   * The given transactionValidDuration was either non-positive, or greater than the maximum 
+   * valid duration of 180 secs (set by Services property hedera.transaction.maxValidDuration).
```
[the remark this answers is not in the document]
tinker-michaelj: Removed property name.

services/basic_types.proto
tinker-michaelj: Revised.

steven-sheehy (approved):

qnswirlds (approved):

PR #107, opened 2021-10-26
Added totalAllowedGasPerSec to throttle_definitions.proto

stoqnkpL:
Signed-off-by: Stoyan Panayotov <stoyan.panayotov@limechain.tech>

**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/services/throttle_definitions.proto
+++ b/services/throttle_definitions.proto
@@ -82,4 +82,5 @@ message ThrottleBucket {
  */
 message ThrottleDefinitions {
   repeated ThrottleBucket throttleBuckets = 1;
+  uint64 totalAllowedGasPerSec = 2;
 }
--- a/services/throttle_definitions.proto
+++ b/services/throttle_definitions.proto
@@ -82,5 +82,6 @@ message ThrottleBucket {
  */
 message ThrottleDefinitions {
   repeated ThrottleBucket throttleBuckets = 1;
-  uint64 totalAllowedGasPerSec = 2;
+  uint64 totalAllowedGasPerSecFrontend = 2;
+  uint64 totalAllowedGasPerSecConsensus = 3;
 }
--- a/services/throttle_definitions.proto
+++ b/services/throttle_definitions.proto
@@ -82,6 +82,4 @@ message ThrottleBucket {
  */
 message ThrottleDefinitions {
   repeated ThrottleBucket throttleBuckets = 1;
-  uint64 totalAllowedGasPerSecFrontend = 2;
-  uint64 totalAllowedGasPerSecConsensus = 3;
 }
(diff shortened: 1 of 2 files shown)


SimiHunjan (approved):

shemnon (approved):

services/throttle_definitions.proto
```diff
@@ -82,4 +82,6 @@ message ThrottleBucket {
  */
 message ThrottleDefinitions {
   repeated ThrottleBucket throttleBuckets = 1;
+  uint64 totalAllowedGasPerSecFrontend = 2;
+  uint64 totalAllowedGasPerSecConsensus = 3;
```
shemnon, line 86: This should be snake_case
```suggestion
  uint64 total_allowed_gas_per_sec_frontend = 2;
  uint64 total_allowed_gas_per_sec_consensus = 3;
```

shemnon:
LGTM pending change to snake_case

services/throttle_definitions.proto
```diff
@@ -82,4 +82,6 @@ message ThrottleBucket {
  */
 message ThrottleDefinitions {
   repeated ThrottleBucket throttleBuckets = 1;
+  uint64 totalAllowedGasPerSecFrontend = 2;
```
tinker-michaelj, line 85: SDK team has requested snake case for all new fields, e.g. 
```
uint64 total_allowed_gas_per_sec_frontend = 2;
```

tinker-michaelj (changes requested):
Just a stylistic change for field naming, otherwise looks good!

services/throttle_definitions.proto
tinker-michaelj, line 85: It's also worth noting that we have an (informal) contract with mirror node operators to **never** remove proto fields, no matter how long they have been deprecated. 😲

So if we have even a 10% doubt that maybe we'll end up following @shemnon's suggestion [here](https://github.com/hashgraph/hedera-services/pull/2373#discussion_r741285719), and using a "normal" throttle bucket for gas throttling, then maybe for release 0.20 we just make these two numbers global/dynamic properties.

But if we're pretty certain about splitting these numbers out in file `0.0.123`, then seems fine.

services/throttle_definitions.proto
stoqnkpL: Moved these to global dynamic properties.

services/throttle_definitions.proto
stoqnkpL: Moved the gas limit configs to global dynamic properties.

tinker-michaelj (approved):
LGTM

PR #109, opened 2021-11-02
access list (accounts or smart contracts and storage keys) for warmin…

tdermendzhievv:
Extend protobufs for EVM access list.

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1420,23 +1420,15 @@ message TokenAssociation {
  * Related to EIP-2929 and EIP-2930 addresses can be provided for "warm up" in EVM before the execution of the
  * transaction in order to save gas.
  */
-message AccessListEntry {
+message ContractAccessEntry {
 
-    oneof entity {
+  /**
+   * The address of the smart contract that should be "warmed up"/loaded in the EVM
+   */
+  ContractID contractID = 1;
 
-        /**
-         * The address of the account that should be "warmed up"/loaded in the EVM
-         */
-        AccountID accountID = 1;
-
-        /**
-         * The address of the smart contract that should be "warmed up"/loaded in the EVM
-         */
-        ContractID contractID = 2;
-    }
-
-    /**
-     * The storage keys of the smart contract that should be "warmed up"/loaded in the EVM
-     */
-    repeated bytes storageKeys = 3;
+  /**
+   * The storage keys of the smart contract that should be "warmed up"/loaded in the EVM
+   */
+  repeated bytes storageKeys = 2;
 }
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1425,10 +1425,10 @@ message ContractAccessEntry {
   /**
    * The address of the smart contract that should be "warmed up"/loaded in the EVM
    */
-  ContractID contractID = 1;
+  ContractID contract_ID = 1;
 
   /**
    * The storage keys of the smart contract that should be "warmed up"/loaded in the EVM
    */
-  repeated bytes storageKeys = 2;
+  repeated bytes storage_keys = 2;
 }
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1415,3 +1415,28 @@ message TokenAssociation {
     TokenID token_id = 1;  // The token involved in the association
     AccountID account_id = 2; // The account involved in the association
 }
+
+/**
+ * Related to EIP-2929 and EIP-2930 addresses can be provided for "warm up" in EVM before the execution of the
+ * transaction in order to save gas.
+ */
+message AccessListEntry {
+
+    oneof entity {
+
+        /**
+         * The address of the account that should be "warmed up"/loaded in the EVM
+         */
+        AccountID accountID = 1;
+
+        /**
+         * The address of the smart contract that should be "warmed up"/loaded in the EVM
+         */
+        ContractID contractID = 2;
+    }
+
+    /**
+     * The storage keys of the smart contract that should be "warmed up"/loaded in the EVM
+     */
+    repeated bytes storageKeys = 3;
+}
--- a/services/contract_call.proto
+++ b/services/contract_call.proto
@@ -63,10 +63,10 @@ message ContractCallTransactionBody {
     /**
      * List with accounts that should be "warmed up"/loaded in the EVM
      */
-    repeated AccountID accountAccess = 5;
+    repeated AccountID account_access = 5;
 
     /**
      * List with smart contracts and storage keys that should be "warmed up"/loaded in the EVM
      */
-    repeated ContractAccessEntry contractAccess = 6;
+    repeated ContractAccessEntry contract_access = 6;
 }
--- a/services/contract_call.proto
+++ b/services/contract_call.proto
@@ -61,7 +61,12 @@ message ContractCallTransactionBody {
     bytes functionParameters = 4;
 
     /**
-     * List with accounts or smart contracts and storage keys that should be "warmed up"/loaded in the EVM
+     * List with smart contracts and storage keys that should be "warmed up"/loaded in the EVM
      */
-    repeated AccessListEntry accessList = 5;
+    repeated AccountID accountAccess = 5;
+
+    /**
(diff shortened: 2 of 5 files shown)


services/basic_types.proto
```diff
@@ -1415,3 +1415,28 @@ message TokenAssociation {
...
+        /**
+         * The address of the smart contract that should be "warmed up"/loaded in the EVM
+         */
+        ContractID contractID = 2;
+    }
+
+    /**
+     * The storage keys of the smart contract that should be "warmed up"/loaded in the EVM
+     */
+    repeated bytes storageKeys = 3;
```
shemnon, line 1441: Only contracts have storage.  If we need to separate account and storage IDs I think we should explore only having storage keys when it's a contract, such as creating a separate message and having he oneof being an AccountId or the new combined message.

services/contract_call.proto
```diff
@@ -59,4 +59,9 @@ message ContractCallTransactionBody {
      * which function to call, and the parameters to pass to the function
      */
     bytes functionParameters = 4;
+
+    /**
+     * List with accounts or smart contracts and storage keys that should be "warmed up"/loaded in the EVM
+     */
+    repeated AccessListEntry accessList = 5;
```
shemnon, line 66: One possibility is instead of having an AccessListEntry that maps directly to the Ethereum model if we have two fields, `repeated AccountId accountAccesses = 5;` and `repeated ContractAccecssEntry contractAccesses = 6;` and the `AccessListEntry` only has `ContractID` and the repeated storageKeys entry.  This would also tidy up the logic in the implementation with less if clauses.

Daniel-K-Ivanov: I was thinking about the same problem and wondered whether if its an overkill to add 2 message types, but since you are suggesting the same, I think that we should do it that way and add separate messages for accounts and contracts 👍

Daniel-K-Ivanov (approved):

services/contract_call.proto
```diff
@@ -59,4 +59,14 @@ message ContractCallTransactionBody {
      * which function to call, and the parameters to pass to the function
      */
     bytes functionParameters = 4;
+
+    /**
+     * List with smart contracts and storage keys that should be "warmed up"/loaded in the EVM
```
shemnon, line 64:
```suggestion
  * List with accounts that should be "warmed up"/loaded in the EVM
```

services/contract_call_local.proto
```diff
@@ -134,6 +134,16 @@ message ContractCallLocalQuery {
      * max number of bytes that the result might include. The run will fail if it would have returned more than this number of bytes.
      */
     int64 maxResultSize = 5 [deprecated=true];
+
+    /**
+     * List with smart contracts and storage keys that should be "warmed up"/loaded in the EVM
```
shemnon, line 139:
```suggestion
  * List with accounts that should be "warmed up"/loaded in the EVM
```

services/contract_create.proto
```diff
@@ -147,4 +147,14 @@ message ContractCreateTransactionBody {
      * the memo that was submitted as part of the contract (max 100 bytes)
      */
     string memo = 13;
+
+    /**
+     * List with smart contracts and storage keys that should be "warmed up"/loaded in the EVM
```
shemnon, line 152:
```suggestion
     * List with accounts that should be "warmed up"/loaded in the EVM
```

Daniel-K-Ivanov:
I think that new fields should be snake case

tdermendzhievv:
> I think that new fields should be snake case

Field names are changed to snake case.

services/basic_types.proto
```diff
@@ -1415,3 +1415,20 @@ message TokenAssociation {
...
+/**
+ * Related to EIP-2929 and EIP-2930 addresses can be provided for "warm up" in EVM before the execution of the
+ * transaction in order to save gas.
+ */
+message ContractAccessEntry {
+
+  /**
+   * The address of the smart contract that should be "warmed up"/loaded in the EVM
+   */
+  ContractID contract_ID = 1;
```
steven-sheehy, line 1428:
```suggestion
  ContractID contract_id = 1;
```

services/basic_types.proto
```diff
@@ -1415,3 +1415,20 @@ message TokenAssociation {
...
+
+  /**
+   * The storage keys of the smart contract that should be "warmed up"/loaded in the EVM
+   */
+  repeated bytes storage_keys = 2;
```
steven-sheehy, line 1433: What is the format of the keys on Hedera? Can the `Key` object be used here? If not, should document the format expected.

services/contract_create.proto
```diff
@@ -147,4 +147,14 @@ message ContractCreateTransactionBody {
...
+
+    /**
+     * List with accounts that should be "warmed up"/loaded in the EVM
+     */
+    repeated AccountID account_access = 14;
+
+    /**
+     * List with smart contracts and storage keys that should be "warmed up"/loaded in the EVM
+     */
+    repeated ContractAccessEntry contract_access = 15;
```
steven-sheehy, line 159: We should also support these on a contract update and show them on contract info query.

services/contract_call_local.proto
```diff
@@ -134,6 +134,16 @@ message ContractCallLocalQuery {
      * max number of bytes that the result might include. The run will fail if it would have returned more than this number of bytes.
      */
     int64 maxResultSize = 5 [deprecated=true];
+
+    /**
+     * List with accounts that should be "warmed up"/loaded in the EVM
```
steven-sheehy, line 139: If I understand correctly, it sounds like this feature is used to pre-cache certain data in the EVM so contract calls are faster for people that pay for the privilege. This makes sense to have on a contract create or update.

So I'm confused on how these can be utilized on a contract call or contract call local query. When you're already executing a contract there's no chance to warm it up. If you're saying that the next call will see the warm data that might make sense for contract call but not contract call local since it should not modify state (outside payment).

Even for contract call I think the access list is something that's more associated with the contract entity and not for every execution of the bytecode. It seems like these access list modifications would be better suited to only be on contract create or update.

steven-sheehy (changes requested):

services/contract_call_local.proto
shemnon: I think the subtlety that is being missed is that the contract calls and slot accesses can be highly variable based on  function call parameters.

The warm up is to bulk query the slot data and account data so that when it is read it is in memory instead of on disk. So when a sub-call, or account read, or slot read happens we don't have to stall for disk access.  Especially querying the code for CALL operations. This is what's killing us the most on the benchmarks pre-virtual-merkle: getting contract code for calls off of the postgres DB.

Another piece of the puzzle is in q2-q3 next year I expect the gas schedules will be changed to make pre-cahced calls always cheaper and in most cases much cheaper than not pre-caching.

The final piece of the puzzle is that this is a path to parallel execution of calls.  Though not as useful in queries since there is no write conflict we can use this data, if we lock in the requirement that access lists limit reads and writes then we can run multiple threads and guarantee there will be no write conflicts with parallel threads.

For the Ethereum perspective see [EIP-2930](https://eips.ethereum.org/EIPS/eip-2930), this is a fairly direct mapping of that spec to hederea.

services/contract_create.proto
shemnon: The access list is an aspect of a particular transaction, not of a contract as a whole.

services/basic_types.proto
shemnon: 32 bytes or less.  Strictly speaking it is a UInt256 but for space it is right packed and zero filled from the left (not sign extended). There is no fixed arrays or 256 byte values in protobuf or we would have used those.

shemnon:
DCO is missing, please add sign-offs to the commits.

services/basic_types.proto
steven-sheehy: Ok, would be good to have the protobuf doc reflect something along the lines of your comment.

services/contract_call_local.proto
steven-sheehy: Ah, okay that makes more sense. Thanks for explaining.

PR #119, opened 2021-11-19
Changes to support  auto account creation

### LINKED ISSUE #118: Update messages to support auto account creation
### Problem

Hedera Services would like to support new way to create accounts using an alias (public key)

### Solution

Make the following changes in our protobufs:
1. Add an `alias` public key to the existing AccountID. Any account should have unique `alias` or `accountNum` set.
```
message AccountID {
    int64 shardNum = 1;
    int64 realmNum = 2;
 oneof account {
    int64 accountNum = 3;
    Key alias = 4;
    }
}
```
2. The account info should have information about alias
```
bytes alias;
```

### Alternatives

_No response_

Neeharika-Sompalli:
**Description**:
Protobuf changes to support auto account creation. 
1. Adds an `alias` Key field for AccountID. Any account can have either accountNum or an alias
2. Modified `AccountInfo` to have `bytes alias` 
3. Added a map of alias and account number pairs in `transaction_record`

**Related issue(s)**:
Fixes #118


services/basic_types.proto
```diff
@@ -79,9 +79,16 @@ message AccountID {
     int64 realmNum = 2;
 
     /**
-     * A nonnegative account number unique within its realm
-     */
-    int64 accountNum = 3;
```
danielakhterov, line 84: Please don't remove fields from protobufs; only deprecate them.

steven-sheehy: It's not removed, it's moved to a oneof

danielakhterov: I assumed moving a field into a `oneof` would break parsing code, but it seems it doesn't.

services/basic_types.proto
```diff
@@ -79,9 +79,16 @@ message AccountID {
...
+     * An account number unique within its realm which can be either a non-negative integer or an alias public key.
+     * ThresholdKey and KeyList are not supported for alias and only a primitive key type is supported.
+     * Alias public key is used for auto account creation using a crypto transfer. Once an alias key is set for an
+     * account it is immutable.
+     */
+    oneof account {
+        int64 accountNum = 3;
```
shemnon, line 88: As a style consideration, can each field in the protobuf get a documentation comment rather than the oneof container?  

I don't see anywhere else where the documentation is on the oneof in lieu of being on the fields themselves, and in all such cases each field in the oneof has documentation. In the single case where oneof has a doc it is because we changed the underlying type of a data field from a protobuf string to a StringValue.  The fields both had the same contextual meaning.

Neeharika-Sompalli: Thanks @shemnon . Updated documentation for each field.

services/crypto_update.proto
```diff
@@ -129,4 +129,22 @@ message CryptoUpdateTransactionBody {
      * including implicit and explicit associations.
      */
     google.protobuf.Int32Value max_automatic_token_associations = 15;
+
+    /**
+     * The new alias public key that can be set for an AccountID. For existing accounts alias key can
```
steven-sheehy, line 134: Not sure how an alias can be updated if it's only supposed to be useful for the creation of a new account.

steven-sheehy (changes requested):

services/transaction_receipt.proto
```diff
@@ -163,4 +163,10 @@ message TransactionReceipt {
      * the newly created NFTs
      */
     repeated int64 serialNumbers = 14;
+
+    /**
+     * In the receipt of a CryptoTransfer transaction used to auto create accounts by transferring to alias
+     * public keys, the ids of newly created accounts
+     */
+    repeated AccountID autoCreatedAccounts = 15;
```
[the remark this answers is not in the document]
danielakhterov: _iirc_ You'd need to really spell it out `auto_created_account_i_ds` which does not look good in snake case at all, unfortunately. That is if you'd want `IDs` specifically. If you're ok with `Ids` a simple `_ids` should work I think.

services/basic_types.proto
```diff
@@ -79,9 +79,30 @@ message AccountID {
     int64 realmNum = 2;
 
     /**
-     * A nonnegative account number unique within its realm
+     * The account number unique within its realm which can be either a non-negative integer or an alias public key.
      */
-    int64 accountNum = 3;
+    oneof account {
```
danielakhterov, line 84: I'm wondering, would it be better to make a wrapper type instead of changing `AccountID` directly?
```protobuf
message Account {
    oneof data {
        AccountID account_id = 1;
        Key key = 2;
    }
}
```
We can then deprecate `TranscationReceipt.account_id` and implement `TransactionReceipt.account` which would return that type.

services/crypto_update.proto
```diff
@@ -129,4 +129,22 @@ message CryptoUpdateTransactionBody {
...
+     * be updated by providing a primitive key or a file ID. If a fileID is provided, its hash is used as alias.
+     * Once an alias is set on an AccountID it is immutable.
+     */
+    oneof new_permanent_alias {
+
+        /**
+         * The new base64 primitive public key that will be updated for alias
+         */
+        Key key_alias = 16;
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Changed to `public_key`

services/crypto_update.proto
```diff
@@ -129,4 +129,22 @@ message CryptoUpdateTransactionBody {
      * including implicit and explicit associations.
      */
     google.protobuf.Int32Value max_automatic_token_associations = 15;
+
+    /**
+     * The new alias public key that can be set for an AccountID. For existing accounts alias key can
+     * be updated by providing a primitive key or a file ID. If a fileID is provided, its hash is used as alias.
+     * Once an alias is set on an AccountID it is immutable.
+     */
+    oneof new_permanent_alias {
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Changed to `alias`. Thanks

services/transaction_receipt.proto
```diff
@@ -163,4 +163,10 @@ message TransactionReceipt {
      * the newly created NFTs
      */
     repeated int64 serialNumbers = 14;
+
+    /**
+     * In the receipt of a CryptoTransfer transaction used to auto create accounts by transferring to alias
+     * public keys, the ids of newly created accounts
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Modified the documentation.

services/transaction_receipt.proto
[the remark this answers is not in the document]
Neeharika-Sompalli: @steven-sheehy @danielakhterov Modified to `auto_created_account_ids`

services/crypto_update.proto
Neeharika-Sompalli: As discussed removed the crypto update changes from this release.

services/crypto_update.proto
```diff
@@ -129,4 +129,22 @@ message CryptoUpdateTransactionBody {
...
+
+        /**
+         * The new base64 primitive public key that will be updated for alias
+         */
+        Key key_alias = 16;
+
+        /**
+         * The FileID of the file whose hash will be updated for alias
```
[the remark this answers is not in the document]
Neeharika-Sompalli: As discussed removed the crypto update changes from this release.

services/crypto_update.proto
```diff
@@ -129,4 +129,22 @@ message CryptoUpdateTransactionBody {
...
+         */
+        FileID hash_alias_file_id = 17;
```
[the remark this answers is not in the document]
Neeharika-Sompalli: As discussed removed the crypto update changes from this release.

services/basic_types.proto
```diff
@@ -79,9 +79,30 @@ message AccountID {
...
+
+        /**
+         * A non-negative account number unique within its realm
+         */
+        int64 accountNum = 3;
+
+        /**
+         * An alias base64 single public key. Only a primitive key is supported as an alias Key (ThresholdKey and KeyList
```
steven-sheehy, line 92: Since it's stored as bytes inside the `Key` it can't be base64.
```suggestion
         * A public key to be used as the account's alias. Only a primitive key is supported as an alias Key (ThresholdKey and KeyList
```

services/basic_types.proto
```diff
@@ -79,9 +79,30 @@ message AccountID {
     int64 realmNum = 2;
 
     /**
-     * A nonnegative account number unique within its realm
+     * The account number unique within its realm which can be either a non-negative integer or an alias public key.
```
steven-sheehy, line 82: Perhaps mention that for any `AccountID` object that appears on a query response, transaction record, or transaction receipt the `accountNum` will be populated while the `alias` will be empty.

services/basic_types.proto
Neeharika-Sompalli: Fixed. Thanks @steven-sheehy

services/basic_types.proto
Neeharika-Sompalli: Added as suggested

services/basic_types.proto
Neeharika-Sompalli: @danielakhterov Thank you ! 
Since we are using `alias` synonymous to ` accountNum`, (like `shard.realm.alias`) it need not be unique across different shard and realm. So, it is added inside `AccountID`.

steven-sheehy (approved):
LGTM. The only other thing I was discussing with Michael was if the record of the transfer should contain some alias to account num mapping for accounts that get created. The mirror node can probably maintain some global cache for `CryptoCreateTransactionBody.key` to `AccountID` mappings to translate aliases though. Just wondering if we should be more explicit and make it easier for other operators.

Neeharika-Sompalli:
> LGTM. The only other thing I was discussing with Michael was if the record of the transfer should contain some alias to account num mapping for accounts that get created. The mirror node can probably maintain some global cache for `CryptoCreateTransactionBody.key` to `AccountID` mappings to translate aliases though. Just wondering if we should be more explicit and make it easier for other operators.

@steven-sheehy  Yes agreed. Added a mapping of a newly created accounts ids and their corresponding alias.

services/basic_types.proto
```diff
@@ -79,9 +79,31 @@ message AccountID {
...
+    oneof account {
+
+        /**
+         * A non-negative account number unique within its realm
+         */
+        int64 accountNum = 3;
+
+        /**
+         * A public key to be used as the account's alias. Only a primitive key is supported as an alias Key
+         * (ThresholdKey and KeyList are not supported).
```
shemnon, line 95: We will need to be sure to test this explicitly in unit and integration tests.

May be worth enumerating the keys as "this currently only includes [list of key fields]" or add contractID and delegatable_contract_id to the list in parentheses.

shemnon (approved):
LGTM

services/basic_types.proto
```diff
@@ -1443,3 +1443,18 @@ message TokenAssociation {
...
+message AccountAlias {
+    /**
+     * Alias for given account
+     */
+    Key alias = 1;
+
+    /**
+     * Account created by transferring to the alias
+     */
+    AccountID account = 2;
```
steven-sheehy, line 1459: `AccountID` to be consistent. We always append `ID`.

The other option is to add the alias to `AccountNum` if it was present in body. But this mapping is more explicit.

Neeharika-Sompalli: Fixed name to `accountID` . Thanks @steven-sheehy

services/basic_types.proto
Neeharika-Sompalli: @shemnon  Added unit tests in the test plan to verify this.

shemnon (approved):

steven-sheehy (approved):
LGTM

PR #120, opened 2021-11-30
Protobuf changes for HIP-206

tinker-michaelj:
**Description**:
- Adds response codes `REVERTED_SUCCESS` and `MAX_STORAGE_IN_PRICE_REGIME_HAS_BEEN_USED`.
- Supports child records via `TransactionID#nonce` and `TransactionRecord#parent_consensus_timestamp`.
- Permits querying for child receipts and records by parent `TransactionID`.
- Adds `delegatable_contract_id` key type.

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -158,13 +158,14 @@ message TransactionID {
     bool scheduled = 3;
 
     /**
-     * The unique id of a "child" transaction that was executed on behalf of 
-     * a parent transaction. (Children share the transactionValidStart and
-     * accountID of their parent.) 
+     * The identifier for an internal transaction that was spawned as part 
+     * of handling a user transaction. (These internal transactions share the 
+     * transactionValidStart and accountID of the user transaction, so a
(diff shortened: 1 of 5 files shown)


SimiHunjan (dismissed):

services/basic_types.proto
```diff
@@ -156,6 +156,17 @@ message TransactionID {
      * Whether the Transaction is of type Scheduled or no
      */
     bool scheduled = 3;
+
+    /**
+     * The unique id of a "child" transaction that was executed on behalf of
```
steven-sheehy, line 161: I don't think we should mention parent/child when talking about nonce, except maybe in detailing a particular example of a nonce. Instead we should just state that this is an opaque ID used to track internal transactions that are spawned from user submitted transactions.

steven-sheehy (dismissed):
LGTM, just a small nit about a comment

steven-sheehy (approved):

SimiHunjan (approved):

shemnon (approved):

PR #121, opened 2021-12-02
Modifications for the auto account create

Neeharika-Sompalli:
Related to #118 
1. As per latest discussion [here](https://github.com/hashgraph/hedera-improvement-proposal/discussions/187#discussioncomment-1737239) , the `alias` is changed to `bytes` instead of `Key`.
2. Because mirror node need to maintain a map with all accounts and aliases as even after create , any transfer can be sent to an `alias`, the `alias <-> account` mapping is removed from `transaction_record`.

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -91,8 +91,8 @@ message AccountID {
         int64 accountNum = 3;
 
         /**
-         * A public key to be used as the account's alias. Only a primitive key is supported as an alias Key
-         * (ThresholdKey, KeyList, ContractID, and delegatable_contract_id are not supported).
+         * The public key bytes to be used as the account's alias. Currently only primitive key bytes are supported as
+         * an alias (Bytes of ThresholdKey, KeyList, ContractID, and delegatable_contract_id are not supported).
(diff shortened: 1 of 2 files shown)


SimiHunjan (approved):

mehcode:
Just passing through @Neeharika-Sompalli, but what key type? Are we fixing this to ed25519?

steven-sheehy (approved):

lbaird:
It would be useful to add to the comment on the alias after line 95: `The public key bytes are the result of serializing a protobuf Key message for any primitive key type.`

Neeharika-Sompalli:
> It would be useful to add to the comment on the alias after line 95: `The public key bytes are the result of serializing a protobuf Key message for any primitive key type.`

Thanks @lbaird . Added as suggested.

Neeharika-Sompalli:
> 

@mehcode Any primitive key type from protobuf `Key` message are supported.

lbaird (approved):
LGTM

steven-sheehy (approved):

PR #127, opened 2021-12-15
Support HIP-31

tinker-michaelj:
**Description**:
- Adds a `TokenTransferList#expected_decimals` field so clients can stipulate a fungible token transfer will only succeed if the actual decimals match the field's value.
- Adds a `UNEXPECTED_TOKEN_DECIMALS` for the failure case.

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -23,6 +23,7 @@ package proto;
  */
 
 import "timestamp.proto";
+import "google/protobuf/wrappers.proto";
 
 option java_package = "com.hederahashgraph.api.proto.java";
 option java_multiple_files = true;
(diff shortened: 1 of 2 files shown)


qnswirlds (approved):

services/basic_types.proto
```diff
@@ -264,6 +265,12 @@ message TokenTransferList {
      * which has a sender and receiver account, including the serial number of the NFT
      */
     repeated NftTransfer nftTransfers = 3;
+
+    /**
+     * If present, the number of decimals this fungible token type is expected to have. The transfer
+     * will fail with UNEXPECTED_TOKEN_DECIMALS if the actual decimals differ.
+     */
+    google.protobuf.Int32Value expected_decimals = 4;
```
steven-sheehy, line 273: Even though I usually prefer signed since Java doesn't have unsigned, I'm wondering if this should be `UInt32Value` to match the `TokenCreate.decimals` being `uint32`?

SimiHunjan (dismissed):

qnswirlds (approved):

steven-sheehy (approved):

danielakhterov:
Perhaps I'm misunderstanding, but these protobuf changes look different that the ones proposed in [HIP-31](https://github.com/hashgraph/hedera-improvement-proposal/blob/master/HIP/hip-31.md)

PR #129, opened 2021-12-15
Support HIP-21

tinker-michaelj:
**Description**:
- Adds a `NetworkGetInfo` query to the `NetworkService` to return the address book and ledger ID.

❓ &nbsp;_Question for reviewers_: Would it make sense to deprecate the `NetworkGetVersionInfo` query and include that information in the `NetworkGetInfo` query? Note that `NetworkGetVersionInfo` is currently **not** free, and `NetworkGetInfo` is to be free.

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1068,6 +1068,11 @@ enum HederaFunctionality {
      * Unpause the Token
      */
     TokenUnpause = 80;
+
+    /**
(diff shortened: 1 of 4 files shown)


qnswirlds (approved):

tinker-michaelj:
Per @SimiHunjan, this query will only be added to the mirror node API at this time.

PR #135, opened 2022-01-14
Additional response codes for extending alias support

### LINKED ISSUE #134: HIP-32 Add additional response codes for extending alias support
For HIP-32, extending alias support to all transactions and queries, we need proper response codes to be returned when
1. Invalid proxy account ID ->  `INVALID_PROXY_ACCOUNT_ID`
2. Invalid transfer account ID ->  `INVALID_TRANSFER_ACCOUNT_ID`
3. Invalid Fee collector account ID  -> `INVALID_FEE_COLLECTOR_ACCOUNT_ID`
4. Invalid auto renew account ID -> `INVALID_AUTORENEW_ACCOUNT_ID`

Neeharika-Sompalli:
Closes #134 

For extending alias support to all transactions and Queries, added additional response codes to be returned if account ID given is invalid or does not exist.

### DIFF
--- a/services/response_code.proto
+++ b/services/response_code.proto
@@ -1131,4 +1131,9 @@ enum ResponseCodeEnum {
    * The fee collector account id in TokenFeeScheduleUpdate is invalid or does not exist.
    */
   INVALID_FEE_COLLECTOR_ACCOUNT_ID = 286;
+
+  /**
+   * The alias already set on an account cannot be updated using CryptoUpdate transaction.
+   */
+  ALIAS_IS_IMMUTABLE = 287;
 }
--- a/services/response_code.proto
+++ b/services/response_code.proto
@@ -1118,7 +1118,7 @@ enum ResponseCodeEnum {
   UNEXPECTED_TOKEN_DECIMALS = 283;
 
   /**
-   * The proxy account id is invalid or does not exist,
+   * The proxy account id is invalid or does not exist.
    */
   INVALID_PROXY_ACCOUNT_ID = 284;
 
--- a/services/response_code.proto
+++ b/services/response_code.proto
@@ -1131,9 +1131,4 @@ enum ResponseCodeEnum {
    * The fee collector account id in TokenFeeScheduleUpdate is invalid or does not exist.
    */
   INVALID_FEE_COLLECTOR_ACCOUNT_ID = 286;
-
-  /**
-   * The auto renew account id is invalid or does not exist.
-   */
-  INVALID_AUTORENEW_ACCOUNT_ID = 287;
 }
--- a/services/response_code.proto
+++ b/services/response_code.proto
@@ -476,12 +476,12 @@ enum ResponseCodeEnum {
   /**
    * [Deprecated]. attempt to set negative receive record threshold
    */
-  INVALID_RECEIVE_RECORD_THRESHOLD = 86 [deprecated=true];
+  INVALID_RECEIVE_RECORD_THRESHOLD = 86 [deprecated = true];
 
   /**
    * [Deprecated]. attempt to set negative send record threshold
    */
-  INVALID_SEND_RECORD_THRESHOLD = 87 [deprecated=true];
+  INVALID_SEND_RECORD_THRESHOLD = 87 [deprecated = true];
 
   /**
    * Special Account Operations should be performed by only Genesis account, return this code if it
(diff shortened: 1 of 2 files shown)


services/response_code.proto
```diff
@@ -1116,4 +1116,19 @@ enum ResponseCodeEnum {
    * type actually has.
    */
   UNEXPECTED_TOKEN_DECIMALS = 283;
+
+  /**
+   * The proxy account id is invalid or does not exist.
+   */
+  INVALID_PROXY_ACCOUNT_ID = 284;
```
steven-sheehy, line 1123: Seems inefficient to create a new response code for every account ID field. We have hundreds of such fields. Maybe we need another field that outputs the path at which the error takes place? Perhaps consider for the future.

services/response_code.proto
```diff
@@ -1116,4 +1116,19 @@ enum ResponseCodeEnum {
...
+
+  /**
+   * The transfer account id in CryptoDelete transaction is invalid or does not exist.
+   */
+  INVALID_TRANSFER_ACCOUNT_ID = 285;
+
+  /**
+   * The fee collector account id in TokenFeeScheduleUpdate is invalid or does not exist.
+   */
+  INVALID_FEE_COLLECTOR_ACCOUNT_ID = 286;
```
steven-sheehy, line 1133: We should update other places in the protobuf like INVALID_ALIAS_KEY, etc. that mention that alias only applies to crypto transfer.

steven-sheehy (changes requested):

services/response_code.proto
Neeharika-Sompalli: As discussed, since we need specific response codes only for `auto renew account, transfer account, proxy account,  scheduling account, fee collector account`, current approach can be followed. Other places the account id being exercised is the subject that is being created/updated/deleted and can use `INVALID_ACCOUNT_ID` if it is not valid.

services/response_code.proto
Neeharika-Sompalli: @steven-sheehy   `INVALID_ALIAS_KEY` , is still reserved only for CryptoTransfer that auto creates an account , if the `alias` provided for auto account creation fails to be parsed into `Key` .

SimiHunjan (approved):

qnswirlds (approved):

services/response_code.proto
steven-sheehy: Can't all transaction types that take an `AccountID` somewhere in the transaction body now provide an alias key and thus fail if the key is invalid?

Doesn't the `alias` in the `TransactionRecord` now apply to any transaction type and need its documentation updated to reflect that?

Neeharika-Sompalli: @steven-sheehy , Every transaction with an `AccountID` can use it in `alias` form. 
1. But the auto account creation only happens with `CryptoTransfer` to `alias`. During that transaction if the given `alias` is not a valid public key , the response is `INVALID_ALIAS_KEY` .
2. Once account is created and it is used in `alias` form in other transactions, its corresponding `accountNum` from the rebuilt data structure is looked up. If the `alias` is invalid in this scenario proper response codes like `INVALID_PAYER_ACCOUNT_ID` etc., are returned based on where the alias form is used. Because `alias` can be used in multiple places in single Transaction and `INVALID_ALIAS_KEY` doesn't provide enough info. 
 Added response codes in this PR are for the same. 
3. Afaik `alias` in child `TransactionRecord` is populated when auto account is created from `CryptoTransfer`. This doesn't change.

steven-sheehy (dismissed):
LGTM

services/response_code.proto
Neeharika-Sompalli: As discussed, Every transaction with an AccountID can use it in `alias` form.

1. But the auto account creation only happens with `CryptoTransfer` to alias. During that transaction if the given `alias` is not a valid public key , the response is `INVALID_ALIAS_KEY` .
2. Once account is created and it is used in `alias` form in other transactions, its corresponding `accountNum` from the rebuilt data structure is looked up. If the `alias` given is invalid in this scenario proper response codes like `INVALID_PAYER_ACCOUNT_ID` etc., are returned based on where the alias form is used. Because alias can be used in multiple places in single Transaction and `INVALID_ALIAS_KEY` doesn’t provide enough info.
 Added response codes in this PR are for the same.
3. Afaik alias in `child TransactionRecord` is populated when auto account is created from `CryptoTransfer`. This doesn’t change.

steven-sheehy (approved):

tinker-michaelj (approved):
LGTM

PR #138, opened 2022-01-14
Populate `evm_address` in contract create record and support use in HAPI 

tinker-michaelj:
**Description**:
Prepares for `CREATE2` implementation.
1. Allows HAPI clients to use a `ContractID` to target a contract either by its number **or** its EVM address.
2. Deprecates the `ContractFunctionResult#createdContractIDs` field, which will be superseded by `ContractID`s externalized in the record stream via "child transactions".
3. Adds a `bytes evm_address` to the `ContractFunctionResult`, to be populated in the record of each contract creation (whether top-level or internal).

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -142,7 +142,7 @@ message ContractID {
 
     oneof contract {
         /**
-        * A nonnegative number unique within its realm
+        * A nonnegative number unique within a given shard and realm
         */
         int64 contractNum = 3;
 
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -147,10 +147,19 @@ message ContractID {
         int64 contractNum = 3;
 
         /**
-        * The 20-byte EVM address of the contract to call. Note that if the 
-        * target contract was created via CREATE2, its address will have 
-        * been derived from the <a href="https://eips.ethereum.org/EIPS/eip-1014">EIP-1014</a>
-        * specification, and will not have a simple relation to a 0.0.X id.
+        * The 20-byte EVM address of the contract to call. 
+        * 
+        * Every contract has an EVM address determined by its <tt>shard.realm.num</tt> id.
+        * This address is as follows:
+        *   <ol>
+        *     <li>The first 4 bytes are the big-endian representation of the shard.</li>
+        *     <li>The next 8 bytes are the big-endian representation of the realm.</li>
+        *     <li>The final 8 bytes are the big-endian representation of the number.</li>
+        *   </ol>  
+        * 
+        * Contracts created via CREATE2 have an additional, primary address that is derived 
+        * from the <a href="https://eips.ethereum.org/EIPS/eip-1014">EIP-1014</a> specification, 
+        * and does not have a simple relation to a <tt>shard.realm.num</tt> id.
         */
         bytes evm_address = 4;
     }
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -85,7 +85,6 @@ message AccountID {
      * be populated.
      */
     oneof account {
-
         /**
          * A non-negative account number unique within its realm
          */
@@ -105,7 +104,6 @@ message AccountID {
          */
         bytes alias = 4;
     }
-
 }
 
 /**
@@ -142,10 +140,20 @@ message ContractID {
      */
     int64 realmNum = 2;
 
-    /**
-     * A nonnegative number unique within its realm
-     */
-    int64 contractNum = 3;
+    oneof contract {
+        /**
+        * A nonnegative number unique within its realm
+        */
+        int64 contractNum = 3;
+
+        /**
+        * The 20-byte EVM address of the contract to call. Note that if the 
+        * target contract was created via CREATE2, its address will have 
+        * been derived from the <a href="https://eips.ethereum.org/EIPS/eip-1014">EIP-1014</a>
+        * specification, and will not have a simple relation to a 0.0.X id.
+        */
+        bytes evm_address = 4;
+    }
 }
 
 /**
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -157,9 +157,12 @@ message ContractID {
         *     <li>The final 8 bytes are the big-endian representation of the number.</li>
         *   </ol>  
         * 
-        * Contracts created via CREATE2 have an additional, primary address that is derived 
-        * from the <a href="https://eips.ethereum.org/EIPS/eip-1014">EIP-1014</a> specification, 
-        * and does not have a simple relation to a <tt>shard.realm.num</tt> id.
+        * Contracts created via CREATE2 have an <b>additional, primary address</b> that is 
+        * derived from the <a href="https://eips.ethereum.org/EIPS/eip-1014">EIP-1014</a> 
+        * specification, and does not have a simple relation to a <tt>shard.realm.num</tt> id. 
+        * 
+        * (Please do note that CREATE2 contracts can also be referenced by the three-part 
+        * EVM address described above.)
         */
         bytes evm_address = 4;
     }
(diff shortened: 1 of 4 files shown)


SimiHunjan (approved):

services/transaction_record.proto
```diff
@@ -120,6 +120,9 @@ message TransactionRecord {
     /**
      * In the record of an internal CryptoCreate transaction triggered by a user 
      * transaction with a (previously unused) alias, the new account's alias. 
+     * 
+     * In the record of an internal ContractCreate transaction triggered by a user
```
steven-sheehy, line 124: Should we support the non-entity ID solidity address on a user-submitted contract create as well?

services/transaction_record.proto
```diff
@@ -120,6 +120,9 @@ message TransactionRecord {
...
+     * transaction, the new contract's Solidity address.
      */
      bytes alias = 16;
```
steven-sheehy, line 127: I think it might make more sense to populate a `solidity_address` on the `ContractFunctionResult` instead of overloading alias. On the result is useful for the mirror node so that the mapping between contract ID and solidity address is explicit and we don't have to assume it's the shard/realm/num format for some and not others.

steven-sheehy (changes requested):

services/contract_call.proto
```diff
@@ -41,9 +41,19 @@ import "basic_types.proto";
...
+    oneof contract {
+        /**
+        * The <shard>.<realm>.<num> identifier of the contract to call
+        */
+        ContractID contractID = 1;
+
+        /**
+        * The 20-byte Solidity address of the contract to call
+        */
+        bytes solidity_address = 5;
```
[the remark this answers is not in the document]
tinker-michaelj: The `solidity_address` of a contract created via `CREATE2` must obey the [EIP-1014](https://eips.ethereum.org/EIPS/eip-1014) specification for standard contracts to work out-of-the-box. [In Solidity](https://solidity-by-example.org/app/create2/), this is:
```
        bytes32 hash = keccak256(
            abi.encodePacked(bytes1(0xff), address(this), _salt, keccak256(bytecode))
        );
        return address(uint160(uint(hash)));
```
Updated this comment with a link.

services/contract_call.proto
```diff
@@ -41,9 +41,19 @@ import "basic_types.proto";
  */
 message ContractCallTransactionBody {
     /**
-     * the contract instance to call, in the format used in transactions
+     * The contract to call
      */
-    ContractID contractID = 1;
+    oneof contract {
```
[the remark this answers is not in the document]
tinker-michaelj: Afaik, it will not be possible for a contract created on shard `0` to interact with a contract created in shard `1`.

Let me double-check with Leemon.

services/transaction_record.proto
tinker-michaelj: You mean, populate this `alias` field with the Solidity address of a "top-level" contract?

services/transaction_record.proto
tinker-michaelj: Added a `solidity_address` field. 👍

steven-sheehy: Wouldn't it belong on the `ContractFunctionResult`?

services/transaction_record.proto
steven-sheehy: Yes, but also allow it as input to the contract create transaction body. Not saying we have to do that here but keep the design open for such future use.

services/transaction_record.proto
tinker-michaelj: Good point, moved to `ContractFunctionResult`.

services/contract_call_local.proto
```diff
@@ -93,9 +93,21 @@ message ContractFunctionResult {
...
+     * and its Solidity address.)
      */
-    repeated ContractID createdContractIDs = 7;
+    repeated ContractID createdContractIDs = 7 [deprecated=true];
+
+    /**
+     * The new contract's Solidity address. Only populated after release 0.23, where 
+     * each created contract will have its own record.
+     */
+     bytes solidity_address = 8;
```
shemnon, line 110: So this is only populated in sub contracts, not the parent contract?  Hence the lack of a `repeated`?  May be worth noting in the comment as this subtlety may be missed and is very important.

services/contract_call.proto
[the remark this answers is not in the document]
shemnon: So this is only populated in child transactions, not the parent transaction? Hence the lack of a repeated? May be worth noting in the comment as this subtlety may be missed and is very important.

services/basic_types.proto
```diff
@@ -142,10 +140,20 @@ message ContractID {
...
+        /**
+        * A nonnegative number unique within its realm
+        */
+        int64 contractNum = 3;
+
+        /**
+        * The 20-byte EVM address of the contract to call. Note that if the 
+        * target contract was created via CREATE2, its address will have 
+        * been derived from the <a href="https://eips.ethereum.org/EIPS/eip-1014">EIP-1014</a>
+        * specification, and will not have a simple relation to a 0.0.X id.
```
steven-sheehy, line 153: Can we also document the `shard.realm.num` format of the `evm_address`?

services/transaction_record.proto
steven-sheehy, line 124: Think this change can be reverted now?

steven-sheehy (changes requested):

services/contract_call_local.proto
```diff
@@ -93,9 +93,21 @@ message ContractFunctionResult {
...
+     * and its Solidity address.)
      */
-    repeated ContractID createdContractIDs = 7;
+    repeated ContractID createdContractIDs = 7 [deprecated=true];
+
+    /**
+     * The new contract's EVM address. Only populated after release 0.23, where 
+     * each created contract will have its own record.
+     */
+     bytes evm_address = 8;
```
steven-sheehy, line 110: Just to confirm, this value will be populated regardless of CREATE vs CREATE2 or root contract or internal, right?

services/transaction_record.proto
tinker-michaelj: Reverted.

services/basic_types.proto
tinker-michaelj: Added these details.

services/contract_call_local.proto
tinker-michaelj: Correct.

steven-sheehy (dismissed):
LGTM

services/contract_call.proto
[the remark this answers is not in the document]
tinker-michaelj: Updated the comment of the `bytes evm_address` field in _contract_call_local.proto_ to clarify this.

steven-sheehy (dismissed):
SLGTM

services/basic_types.proto
```diff
@@ -142,10 +140,29 @@ message ContractID {
...
+        *     <li>The first 4 bytes are the big-endian representation of the shard.</li>
+        *     <li>The next 8 bytes are the big-endian representation of the realm.</li>
+        *     <li>The final 8 bytes are the big-endian representation of the number.</li>
+        *   </ol>  
+        * 
+        * Contracts created via CREATE2 have an additional, primary address that is derived 
+        * from the <a href="https://eips.ethereum.org/EIPS/eip-1014">EIP-1014</a> specification, 
+        * and does not have a simple relation to a <tt>shard.realm.num</tt> id.
+        */
+        bytes evm_address = 4;
```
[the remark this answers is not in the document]
tinker-michaelj: Added detail to comments on both `evm_address` fields.

> A related question: is it possible for an evm_address created via CREATE2 to collide in the way that if we decode it into shard.realm.num form, shard and realm match what's in this ContractID object?

Logically possible, but inconceivably improbable. 😁

steven-sheehy (approved):

xin-hedera (approved):
LGTM. Thanks for the clarifaction.

PR #142, opened 2022-01-21
HIP-21 Add a mirror API to get node addresses

steven-sheehy:
**Description**:
[HIP-21](https://hips.hedera.com/hip/hip-21) Add a mirror node gRPC API to get node addresses from an address book.

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**


Nana-EC (dismissed):

Nana-EC:
> There's no issue with the file itself.
> 
> however, I'm wondering whether we should rename it since it conflicts with the same file in `services` folder. According to the protobuf spec, iirc at least the latest `protoc-gen-go` compiler will report sth like:
> 
> ```
> panic: proto: file "network_service.proto" is already registered
>         previously from: "github.com/hashgraph/hedera-protobufs-go/services"
>         currently from:  "github.com/hashgraph/hedera-protobufs-go/mirror"
> See https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict
> ```
> 
> As a note, due to the issue, whoever depends on our go sdk has to pin to `google.golang.org/protobuf v1.26.1-0.20210525005349-febffdd88e85`. There is a long discussion here: [golang/protobuf#1122](https://github.com/golang/protobuf/issues/1122)
> 
> The author who fixed the issue in version `protobuf v1.26.1-0.20210525005349-febffdd88e85` reverted the change later as `The decision for whether to permit or reject registration conflicts needs to be decided at the protobuf project level and we (i.e., the Go-specific maintainers) are not the right people to make a judgement call on an issue of this nature.`

Also fine with the file contents. However, if this will create a problem as highlighted we should rename to avoid conflicts.

steven-sheehy:
I've renamed `network_service.proto` to `mirror_network_service.proto` after confirming that the mirror node works with either and that the proto objects inside didn't require the filename to match.

xin-hedera (dismissed):
LGTM

Nana-EC (dismissed):
LGTM

steven-sheehy:
Closing to re-target `develop`

PR #144, opened 2022-01-24
Add `starting_at` option to CryptoGetAccountRecordsQuery

tinker-michaelj:
**Description**:
- Adds an option to `CryptoGetAccountRecordsQuery` to restrict the first payer record returned.
- The user can restrict either:
    1. The earliest consensus time returned; or,
    2. The `TransactionID` of the first record returned.

### DIFF
--- a/services/crypto_get_account_records.proto
+++ b/services/crypto_get_account_records.proto
@@ -43,6 +43,25 @@ message CryptoGetAccountRecordsQuery {
      * The account ID for which the records should be retrieved
      */
     AccountID accountID = 2;
+
+    /**
+     * If set, a restriction on the first payer record to be returned. The queried node will 
+     * return this record (assuming one qualifies), and up to to the next 179 records, in 
+     * chronological order of consensus time.
+     *
+     * Without a restriction, the node will return the most recent payer records (in
+     * chronological order, up to a maximum of 180 records).
+     */
+    oneof starting_at {
+        /**
+        * The transaction id of the first record to return.
+        */
+        TransactionID transaction_id = 3;
+        /**
+        * The earliest consensus time of the first record to return.
+        */
+        Timestamp earliest_consensus_time = 4;
+    }
 }
 
 /**
--- a/services/crypto_get_account_records.proto
+++ b/services/crypto_get_account_records.proto
@@ -47,11 +47,15 @@ message CryptoGetAccountRecordsQuery {
 
     /**
      * If set, a restriction on the first payer record to be returned. The queried node will 
-     * return this record (assuming one qualifies), and up to to the next 179 records, in 
-     * chronological order of consensus time.
+     * return the first qualifying record and its successors---up to a total of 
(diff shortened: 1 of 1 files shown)


services/crypto_get_account_records.proto
```diff
@@ -43,6 +44,25 @@ message CryptoGetAccountRecordsQuery {
      * The account ID for which the records should be retrieved
      */
     AccountID accountID = 2;
+
+    /**
+     * If set, a restriction on the first payer record to be returned. The queried node will 
+     * return this record (assuming one qualifies), and up to to the next 179 records, in
```
rbair23, line 50: Isn't the limit (180) configurable? In which case we shouldn't document 179 here, we should just say we will return all records up to some maximum amount (put it in better words, but that's the idea). Same with the 180 on line 54.

services/crypto_get_account_records.proto
```diff
@@ -43,6 +44,25 @@ message CryptoGetAccountRecordsQuery {
...
+     */
+    oneof starting_at {
+        /**
+        * The transaction id of the first record to return.
+        */
+        TransactionID transaction_id = 3;
+        /**
+        * The earliest consensus time of the first record to return.
+        */
+        Timestamp earliest_consensus_time = 4;
```
rbair23, line 64: Is there any difference in performance depending on which of these two is used? In both cases are we going to iterate through the records looking for the first that meets the criteria?

services/crypto_get_account_records.proto
tinker-michaelj: Updated comment with link to the Services property.

services/crypto_get_account_records.proto
tinker-michaelj: Yes, with current FCQ implementation, we need to iterate through all records no matter what. (If random access was supported, we could do binary search for earliest consensus time, but it isn't.)

steven-sheehy: Not sure we should add this feature until the data structure has a performant way to allow it...

tinker-michaelj: The current implementation already iterates through all records. 😬

steven-sheehy: Yeah, but we're adding a limit now so it will only iterate through the first K records where K is the limit and N is the total number of records. But with this feature they could pick the last transaction id so it will be O(N).

steven-sheehy: Oh, unless it has to iterate through all to filter down by account ID. I'm guessing it's not in a map by account id.

tinker-michaelj: Right, these records are in an `FCQueue` which doesn't even support random access or reverse traversal.

No matter how you restrict (or don't restrict) the query, we must iterate through the entire FCQ.

jsync-swirlds:
Stale.  Please resubmit if needed.

PR #146, opened 2022-01-25
Address reviewer comments

tinker-michaelj:
**Description**:
- Add approve and adjust allowance bodies to the `TransactionBody` and `SchedulableTransactionBody` choices.
- Add approve/adjust transactions to the `CryptoService`.
- Clarify in comments that spender allowances are against _the payer account_ of the `approveAllowances` or `adjustAllowance` transaction.

### DIFF
--- a/services/transaction_body.proto
+++ b/services/transaction_body.proto
@@ -40,6 +40,7 @@ import "crypto_delete_live_hash.proto";
 import "crypto_transfer.proto";
 import "crypto_update.proto";
 import "crypto_approve_allowance.proto";
+import "crypto_adjust_allowance.proto";
 
 import "file_append.proto";
 import "file_create.proto";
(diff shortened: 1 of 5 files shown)


superboo (approved):
Thanks Michael!

services/transaction_body.proto
```diff
@@ -317,8 +318,13 @@ message TransactionBody {
     ScheduleSignTransactionBody scheduleSign = 44;
 
     /**
-     * Adds an approved allowance for spender to transfer owner's hbar.
+     * Adds one or more approved allowances for spenders to transfer the paying account's hbar or tokens.
      */
-    CryptoApproveAllowanceTransactionBody cryptoApproveAllowance = 45;
+    CryptoApproveAllowanceTransactionBody cryptoApproveAllowance = 48;
```
steven-sheehy, line 323: It seems like this file is in lexicographic order, not ordinal order. We should move these to be sorted next to the other Crypto bodies. Would also be nice to add a note at the bottom about ordering and how to choose the next ordinal when adding new bodies.

tinker-michaelj: Moved these to the `Crypto*` group and added a comment.

steven-sheehy (approved):
LGTM

PR #151, opened 2022-01-28
Breakout approval allowance for NFTs

### LINKED ISSUE #132: Approval & Allowance feature
### Problem

This feature does not currently exist.

The Approval/Allowance feature allows an account owner (user A) to delegate another account (user B) to spend tokens from account-A on behalf of user-A. Tokens include the native hbar token, fungible and non-fungible HTS tokens. Token owners can also increase or decrease the number of tokens the delegated user (user-B) is allowed to spend on behalf of the token owner (user-A).

Reference HIP: TBD

### Solution

Add support for this feature

### Alternatives

_No response_

superboo:
**Description**:
Breakout NFT allowances from fungible tokens to allow for ERC721 compatibility. We were previously allowing x transfers of NFT instances when in fact we should be very explicit in the NFT transfers to allow since one NFT is not interchangeable for another.

Create NftAllowance nested message in CryptoGetInfoResponse to return NFT specific info.

Create NftApproval message to be used in CryptoApproveAllowanceTransactionBody, CryptoAdjustAllowanceTransactionBody and TransactionRecord.

**Related issue(s)**:

Fixes #132

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1495,9 +1495,9 @@ message TokenAssociation {
 }
 
 /**
- * An approved allowance for hbar transfers for a spender.
+ * An approved allowance of hbar transfers for a spender.
  */
-message CryptoApproval {
+message CryptoAllowance {
     /**
      * The account ID of the spender of the hbar allowance.
      */
@@ -1510,9 +1510,36 @@ message CryptoApproval {
 }
 
 /**
- * An approved allowance for fungible token transfers for a spender.
+ * An approved allowance of non-fungible token transfers for a spender.
  */
-message TokenApproval {
+message NftAllowance {
+    /**
+     * The token that the allowance pertains to.
+     */
+    TokenID tokenId = 1;
+
+    /**
+     * The account ID of the token allowance spender.
+     */
+    AccountID spender = 2;
+
+    /**
+     * The list of serial numbers that the spender is permitted to transfer.
+     */
+    repeated int64 serialNumbers = 3;
+
+    /**
+     * If true, the spender has access to all of the account owner's NFT instances (currently
+     * owned and any in the future). If this field is set to true the serialNumbers field
+     * should be empty.
+     */
+    google.protobuf.BoolValue approvedForAll = 4;
+}
+
+/**
+ * An approved allowance of fungible token transfers for a spender.
+ */
+message TokenAllowance {
     /**
      * The token that the allowance pertains to.
      */
@@ -1529,31 +1556,4 @@ message TokenApproval {
     int64 amount = 3;
 }
 
-/**
- * An approved allowance for non-fungible token transfers for a spender.
- */
-message NftApproval {
-    /**
-     * The token that the allowance pertains to.
-     */
-    TokenID tokenId = 1;
-
-    /**
-     * The account ID of the token allowance spender.
-     */
-    AccountID spender = 2;
-
-    oneof allowance {
-        /**
-         * The list of serial numbers that the spender is permitted to transfer.
-         */
-        repeated int64 serialNumbers = 3 [packed = true];
-
-        /**
-         * If true, the spender has access to all of the account owner's NFT instances (currently
-         * owned and any in the future).
-         */
-        bool approvedForAll = 4;
-    }
-}
 
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1493,3 +1493,67 @@ message TokenAssociation {
     TokenID token_id = 1;  // The token involved in the association
     AccountID account_id = 2; // The account involved in the association
 }
+
+/**
+ * An approved allowance for hbar transfers for a spender.
+ */
+message CryptoApproval {
+    /**
+     * The account ID of the spender of the hbar allowance.
+     */
+    AccountID spender = 1;
+
+    /**
+     * The amount of the spender's allowance in tinybars.
+     */
+    int64 amount = 2;
+}
+
+/**
+ * An approved allowance for fungible token transfers for a spender.
+ */
+message TokenApproval {
+    /**
+     * The token that the allowance pertains to.
+     */
+    TokenID tokenId = 1;
+
+    /**
+     * The account ID of the token allowance spender.
+     */
+    AccountID spender = 2;
+
+    /**
+     * The amount of the spender's token allowance.
+     */
+    int64 amount = 3;
+}
+
+/**
+ * An approved allowance for non-fungible token transfers for a spender.
+ */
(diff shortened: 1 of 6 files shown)


services/basic_types.proto
```diff
@@ -1493,3 +1493,67 @@ message TokenAssociation {
...
+    /**
+     * The account ID of the token allowance spender.
+     */
+    AccountID spender = 2;
+
+    oneof allowance {
+        /**
+         * The list of serial numbers that the spender is permitted to transfer.
+         */
+        repeated int64 serialNumbers = 3 [packed = true];
```
steven-sheehy, line 1550: `packed=true` is the default in proto3 and unnecessary to specify

steven-sheehy, line 1550: For adjust, I assume all positive numbers are added to the list while discarding already existing serials?

And conversely for negative numbers they are removed from the list, but what happens if they don't exist?

What happens if `serialNumbers` contains duplicates?

services/crypto_get_info.proto
```diff
@@ -59,6 +59,64 @@ message CryptoGetInfoResponse {
      */
     ResponseHeader header = 1;
 
+    message CryptoAllowance {
```
steven-sheehy, line 62: This can be replaced with `CryptoApproval`. We might consider renaming `CryptoApproval` with `CryptoAllowance` and just letting the `approve` action be in the service name and field names.

services/crypto_get_info.proto
```diff
@@ -59,6 +59,64 @@ message CryptoGetInfoResponse {
...
+         */
+        int64 amount = 3;
+
+        /**
+         * The number of decimal places the token allowance value is divisible by.
+         */
+        int32 decimals = 4;
+    }
+
+    message NftAllowance {
```
steven-sheehy, line 96: Same thing, can consolidate with `NftApproval`.

services/crypto_approve_allowance.proto
```diff
@@ -0,0 +1,53 @@
...
+message CryptoApproveAllowanceTransactionBody {
+    /**
+     * List of hbar allowances approved by the account owner.
+     */
+    repeated CryptoApproval cryptoApprovals = 1;
+
+    /**
+     * List of token fungible allowances approved by the account owner.
+     */
+    repeated TokenApproval tokenApprovals = 2;
```
steven-sheehy, line 47: nit: Sort nft in between crypto and token here and elsewhere.

services/transaction_record.proto
```diff
@@ -122,4 +122,22 @@ message TransactionRecord {
      * transaction with a (previously unused) alias, the new account's alias. 
      */
      bytes alias = 16;
+
+     /**
+      * The current balances of any adjusted crypto allowances as a result of this
+      * transaction.
```
steven-sheehy, line 128: Maybe specify only populated in the case of a `CryptoAdjustAllowanceTransaction`

services/crypto_get_info.proto
```diff
@@ -59,6 +59,64 @@ message CryptoGetInfoResponse {
...
+        /**
+         * The account ID of the token allowance spender.
+         */
+        AccountID spender = 2;
+
+        oneof allowance {
+            /**
+             * List of serial numbers that the spender is permitted to transfer.
+             */
+            repeated int64 serialNumbers = 3 [packed = true];
```
steven-sheehy, line 111: packed unnecessary

steven-sheehy (changes requested):

services/crypto_get_info.proto
Neeharika-Sompalli: Agreed, can reuse `basic_types`

steven-sheehy: Can potentially use composition to consolidate `TokenApproval` and `TokenAllowance`.

services/basic_types.proto
superboo: Fixed

superboo: Yes, positive will add + negative will remove. It will be an error if the user tries to remove an NFT that the owner doesn't own. Before this change, duplicate token allowances in the list would be treated as an error since it would be confusing to know which one to pick if the amounts were different (do we take the last in the list?). Now that we have NFTs broken out, it's actually not harmful nor confusing if dups are in the list. It's really a waste of bytes and CPU to parse the protobuf. If it's easier for mirror node to not have an NFT list with dups, we'll return an error if we catch any.

services/crypto_get_info.proto
superboo: Elected to re-use the basic types + rename them to `*Allowances`. The oneof in the `NftAllowances` is also being removed since oneof doesn't support repeated (funny, only saw the error after the packed was removed).

services/crypto_get_info.proto
superboo: Fixed

services/crypto_approve_allowance.proto
superboo: Done

services/crypto_get_info.proto
superboo: Move to `NftAllowance` everywhere.

services/transaction_record.proto
superboo: Done

services/basic_types.proto
steven-sheehy: Would probably be best to error if duplicates.

anighanta:
@superboo @rbair23 

so the only way now to distinguish the cryptoTransfer if it is an allowance transfer is that the signatures on that transactions will not have the owner’s key ? I think we need more information in the cryptoTransfer that use allowances that specify this.

`owner`  has 1000 hbar, granted allowance of 500 hbar to `spender`  and a `receiver` with receiverSigRequired as false

`Scenario 1` : `spender` submits a cryptoTransfer of `-100 owner` , `+100 receiver` signed by just the `spender`’s key
`Scenario 2` : `spender` submits a cryptoTransfer of `-100 owner` , `+100 receiver` signed by both `spender` and `owner`’s key

`Scenario 1` should be using `spender`’s allowance and `Scenario 2` shouldn’t. To differentiate and implement this in the CryptoTransferTransitionLogic we have to expand signatures again and compare. Would a change to protobufs like adding a new field to AccountAmounts/NftTransfer that denote to use allowance or a new txn similar to cryptoTranfer be better ?

superboo:
Closing as this one targeted to develop, https://github.com/hashgraph/hedera-protobufs/pull/152 based off main

PR #155, opened 2022-01-31
Additional response codes for Approval/Allowance

Neeharika-Sompalli:
Based on latest HIP changes, added additional response codes.

### DIFF
--- a/services/response_code.proto
+++ b/services/response_code.proto
@@ -1194,7 +1194,7 @@ enum ResponseCodeEnum {
   /**
    * Fungible common token used in NFT allowances
    */
-  FUNGIBLE_COMMON_TOKEN_IN_NFT_ALLOWANCES = 298;
+  FUNGIBLE_TOKEN_IN_NFT_ALLOWANCES = 298;
 
   /**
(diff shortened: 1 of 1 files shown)


services/response_code.proto
```diff
@@ -1179,4 +1179,26 @@ enum ResponseCodeEnum {
...
+
+  /**
+   * Serial numbers are repeated in nft allowance for a single spender account
+   */
+  REPEATED_SERIAL_NUMS_IN_NFT_ALLOWANCES = 297;
+
+  /**
+   * Fungible common token used in NFT allowances
+   */
+  FUNGIBLE_COMMON_TOKEN_IN_NFT_ALLOWANCES = 298;
```
anighanta, line 1197: nit : could be named without out `COMMON`

anighanta (dismissed):

services/response_code.proto
Neeharika-Sompalli: Done.

steven-sheehy (approved):

anighanta (approved):

PR #163, opened 2022-03-15
Add precompile call data to ContractFunctionResults

### LINKED ISSUE #157: Expand ContractFunctionResult to include FunctionParameters, Amount and GasLimit
### Problem

For precompiled function calls the applicable `FunctionParameters`, `Amount` and `GasLimit` properties stated on call are missing in explicit specification.
This leaves readers of the record stream with missing info or with the problem of being forced to implement heavy logic to recreate the data in an identical manner as the client side.

These values are usually found in the `ContractCreate`/`ContractCall` transaction bodies.
However, in the precompile case the transactions are non contract transactions and are therefore missing as easy properties to read.

### Solution

- Add `FunctionParameters` for the precompile function to the `ContractFunctionResult` proto
- Add `Amount` sent to contract to the `ContractFunctionResult` proto. If not payable set to 0
- Add `GasLimit` to the `ContractFunctionResult` proto. Can match the parent transaction gasLimit

### Alternatives

_No response_

shemnon:
**Description**:

Add precompiled call data for parameters, gas, and amount to the
ContractFunctionResults. These would be in the TransactionBody
if it was a normal call but the body for precompiles is a
different type,

Signed-off-by: Danno Ferrin <danno.ferrin@hedera.com>

**Related issue(s)**:

Fixes #157

**Checklist**

### DIFF
--- a/services/contract_call_local.proto
+++ b/services/contract_call_local.proto
@@ -132,6 +132,31 @@ message ContractFunctionResult {
      * EVM address described above.)
      */
      google.protobuf.BytesValue evm_address = 9;
+
+    /**
+     * The amount of gas that was used for this call. 
+     *
+     * This field should only be populated when the paired TransactionBody in the record stream is not a
+     * ContractCreateTransactionBody or a ContractCallTransactionBody.     
+     */
+    int64 gas = 10;
+
+    /**
+     * Number of tinybars sent (the function must be payable if this is nonzero).
+     *
+     * This field should only be populated when the paired TransactionBody in the record stream is not a
+     * ContractCreateTransactionBody or a ContractCallTransactionBody.     
+     */
+    int64 amount = 11;
+
+    /**
+     * The parameters passed into the contract call.
+     *
+     * This field should only be populated when the paired TransactionBody in the record stream is not a
+     * ContractCreateTransactionBody or a ContractCallTransactionBody.     
+     */
+    bytes functionParameters = 12;
+     
 }
 
 /**
--- a/services/contract_call_local.proto
+++ b/services/contract_call_local.proto
@@ -134,7 +134,7 @@ message ContractFunctionResult {
      google.protobuf.BytesValue evm_address = 9;
 
     /**
-     * The amount of gas that was used for this call. 
+     * The amount of gas available for the call, aka the gasLimit. 
      *
      * This field should only be populated when the paired TransactionBody in the record stream is not a
      * ContractCreateTransactionBody or a ContractCallTransactionBody.     


services/contract_call_local.proto
```diff
@@ -132,6 +132,31 @@ message ContractFunctionResult {
      * EVM address described above.)
      */
      google.protobuf.BytesValue evm_address = 9;
+
+    /**
+     * The amount of gas that was used for this call.
```
steven-sheehy, line 137: There's already a `gasUsed` that shows the gas used. This would be the gas limit that shows the maximum amount of gas remaining still to use, right?

services/contract_call_local.proto
```diff
@@ -132,6 +132,31 @@ message ContractFunctionResult {
...
      */
      google.protobuf.BytesValue evm_address = 9;
+
+    /**
+     * The amount of gas that was used for this call. 
+     *
+     * This field should only be populated when the paired TransactionBody in the record stream is not a
+     * ContractCreateTransactionBody or a ContractCallTransactionBody.     
+     */
+    int64 gas = 10;
```
steven-sheehy, line 142: Should we use wrapper types for these fields to clearly indicate not set or is primitive default values sufficient?

services/contract_call_local.proto
SimiHunjan: Is it supposed to represent actual gas used maybe? Related to this issue https://github.com/hashgraph/hedera-services/issues/2949.

Nana-EC: Steven is right this should represent the `gasLimit`.
`gasUsed` is already captured in item 5 of this proto object

services/contract_call_local.proto
```diff
@@ -132,6 +132,31 @@ message ContractFunctionResult {
...
+     */
+    int64 amount = 11;
+
+    /**
+     * The parameters passed into the contract call.
+     *
+     * This field should only be populated when the paired TransactionBody in the record stream is not a
+     * ContractCreateTransactionBody or a ContractCallTransactionBody.     
+     */
+    bytes functionParameters = 12;
```
Nana-EC, line 158: Is there a max size on the bytes length that's worthy of note here?

Nana-EC (commented):
Some suggestions.
Commenting rather than requesting changes so as not to block whiles i'm OOF
Please wait for 1 mirror node team member sign off before merging.

services/contract_call_local.proto
shemnon: There is no meaning for a missing field or a default field here like there is with the tracing valueWritten field, so a wrapper type would just add more bytes to the record.

services/contract_call_local.proto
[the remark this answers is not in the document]
shemnon: Wrapper type adds no bytes.  The context of the record provides the meaning: is it a child transaction?  Wrapper types are usefull for when null is different from a zero length array.  Nulls won't matter in the code.

shemnon: No protocol imposed limit.  The functional limit is how much gas you want to pay.

services/contract_call_local.proto
shemnon: `gas` is the name of the field used in the three paired transaciton types, and I want to try and keep it the same here.  I updated the docs.

steven-sheehy: Right, was just arguing for a doc fix here. Looks good

steven-sheehy (approved):

services/contract_call_local.proto
Nana-EC: Sounds good

Nana-EC (approved):
LGTM

PR #168, opened 2022-03-24
Remove serials in GetAccountInfo for NFT allowances

Neeharika-Sompalli:
Signed-off-by: Neeharika-Sompalli <neeharika.sompalli@hedera.com>

- Since there is no more a map in account state with list of serials for NFTs and is moved to `MerkleUniqueToken` , removing it from the getAccountInfo `GrantedNftAllowance`

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1571,6 +1571,12 @@ message NftAllowance {
      * should be empty.
      */
     google.protobuf.BoolValue approvedForAll = 5;
+
+    /**
+     * The account ID of the spender who is granted approvedForAll allowance and granting
+     * approval on an NFT serial to another spender.
+     */
+    AccountID delegatingSpender = 6;
 }
 
 /**
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1614,7 +1614,7 @@ message GrantedCryptoAllowance {
 }
 
 /**
- * A granted allowance of non-fungible token transfers for a spender relative to the owner account.
+ * A granted allowance for all the NFTs of a token for a spender relative to the owner account.
  */
 message GrantedNftAllowance {
     /**
@@ -1623,21 +1623,9 @@ message GrantedNftAllowance {
     TokenID token_id = 1;
 
     /**
-     * The account ID of the token allowance spender.
+     * The account ID of the spender that has been granted access to all NFTs of the owner
      */
     AccountID spender = 2;
-
-    /**
-     * The list of serial numbers that the spender is permitted to transfer.
-     */
-    repeated int64 serial_numbers = 3;
-
-    /**
-     * If true, the spender has access to all of the account owner's NFT instances (currently
-     * owned and any in the future). If this field is set to true the serialNumbers field
-     * should be empty.
-     */
-    bool approved_for_all = 4;
 }
 
 /**
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1563,20 +1563,20 @@ message NftAllowance {
     /**
      * The list of serial numbers that the spender is permitted to transfer.
      */
-    repeated int64 serialNumbers = 4;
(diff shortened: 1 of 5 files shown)


anighanta (dismissed):

services/basic_types.proto
```diff
@@ -1571,6 +1571,12 @@ message NftAllowance {
      * should be empty.
      */
     google.protobuf.BoolValue approvedForAll = 5;
+
+    /**
+     * The account ID of the spender who is granted approvedForAll allowance and granting
+     * approval on an NFT serial to another spender.
+     */
+    AccountID delegatingSpender = 6;
```
tinker-michaelj, line 1579: I think we're trying to standardize on snake-case (`delegate_spender`, `approved_for_all`).

tinker-michaelj (dismissed):
LGTM!

services/basic_types.proto
Neeharika-Sompalli: Fixed @tinker-michaelj  ! Thanks !

services/basic_types.proto
```diff
@@ -1623,21 +1628,9 @@ message GrantedNftAllowance {
...
-     * The list of serial numbers that the spender is permitted to transfer.
-     */
-    repeated int64 serial_numbers = 3;
-
-    /**
-     * If true, the spender has access to all of the account owner's NFT instances (currently
-     * owned and any in the future). If this field is set to true the serialNumbers field
-     * should be empty.
-     */
-    bool approved_for_all = 4;
```
steven-sheehy, line 1640: Why can't we keep approvedForAll in the query? It should be associated with the spender account and performant to retrieve.

services/basic_types.proto
```diff
@@ -1623,21 +1628,9 @@ message GrantedNftAllowance {
...
-     * The account ID of the token allowance spender.
+     * The account ID of the spender that has been granted access to all NFTs of the owner
      */
     AccountID spender = 2;
-
-    /**
-     * The list of serial numbers that the spender is permitted to transfer.
-     */
-    repeated int64 serial_numbers = 3;
-
```
steven-sheehy, line 1634: Shouldn't we also return the new delegating_spender field?

steven-sheehy (changes requested):

services/basic_types.proto
anighanta: No point in having that @steven-sheehy, it will always be true.

services/basic_types.proto
anighanta: All spenders that have approve_for_all allowance grant can be delegating_spenders. It will be redundant to have to this field here.

services/basic_types.proto
```diff
@@ -1563,14 +1563,19 @@ message NftAllowance {
...
 
     /**
-     * If true, the spender has access to all of the account owner's NFT instances (currently
-     * owned and any in the future). If this field is set to true the serialNumbers field
-     * should be empty.
+     * The account ID of the spender who is granted approvedForAll allowance and granting
+     * approval on an NFT serial to another spender.
      */
-    google.protobuf.BoolValue approvedForAll = 5;
+    AccountID delegating_spender = 6;
```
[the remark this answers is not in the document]
anighanta: `NftApproveAllowance{owner=Alice, spender=Carol, tokenId=FOOBAR, approvedForAll=false, serialNumbers=[1,2,3], delegatingSpender=Bob}`

This would be correct syntax to will let `Bob` give allowance to `Carol`. Mind that this txn has to be signed by `Bob` instead of owner `Alice` .

Bob can still spend the serials 1, 2 ,3.

Only approveForAll spenders can delegate. So there is only 1 level of such delegation.

steven-sheehy (approved):
LGTM

SimiHunjan (approved):

PR #182, opened 2022-04-07
Query for getting complete account details

Neeharika-Sompalli:
Added `GetAccountDetails` query that returns allowances for **internal use only**. 
This returns all info of an account including allowance details. This query **WILL NOT** be documented in HAPI


superboo (approved):
LGTM

services/get_account_details.proto
```diff
@@ -0,0 +1,242 @@
...
+        /**
+         * The current balance of account in tinybars
+         */
+        uint64 balance = 8;
+
+        /**
+         * [Deprecated]. The threshold amount, in tinybars, at which a record is created of any
+         * transaction that decreases the balance of this account by more than the threshold
+         */
+        uint64 generate_send_record_threshold = 9 [deprecated=true];
```
tinker-michaelj, line 110: We don't even have these thresholds in state any more; so probably no reason include them in this message. 😁

services/get_account_details.proto
```diff
@@ -0,0 +1,242 @@
...
+         * insufficient funds, then it extends as long as possible. If it is empty when it expires,
+         * then it is deleted.
+         */
+        Duration auto_renew_period = 13;
+
+        /**
+         * All of the livehashes attached to the account (each of which is a hash along with the
+         * keys that authorized it and can delete it)
+         */
+        repeated LiveHash live_hashes = 14;
```
tinker-michaelj, line 139: Also wouldn't include this.

tinker-michaelj (commented):
Noticed a couple unsupported fields that we probably gain nothing from including in the response.

Also, we need a new `NetworkService` `rpc` [here](https://github.com/hashgraph/hedera-protobufs/blob/main/services/network_service.proto#L41); and a new `HederaFunctionality` constant [here](https://github.com/hashgraph/hedera-protobufs/blob/main/services/basic_types.proto#L1099).

services/get_account_details.proto
Neeharika-Sompalli: Removed them . Thanks Michael !

services/get_account_details.proto
Neeharika-Sompalli: Removed . Thanks Michael !

tinker-michaelj (approved):
LGTM!

PR #184, opened 2022-04-11
feat(#183): add changes for long term scheduled transactions

### LINKED ISSUE #183: HIP-423: Long Term Scheduled Transactions - protobuf changes
### Problem

Implement protobuf changes for HIP-423: Long Term Scheduled Transactions

### Solution

Implement protobuf changes for HIP-423: Long Term Scheduled Transactions

### Alternatives

_No response_

statop:
**Description**:
Implement protobuf changes for HIP-423: Long Term Scheduled Transactions

**Related issue(s)**:

Fixes #183

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/services/schedule_create.proto
+++ b/services/schedule_create.proto
@@ -26,26 +26,36 @@ option java_package = "com.hederahashgraph.api.proto.java";
 option java_multiple_files = true;
 
 import "basic_types.proto";
+import "timestamp.proto";
 import "schedulable_transaction_body.proto";
 
 /**
  * Create a new <i>schedule entity</i> (or simply, <i>schedule</i>) in the network's action queue.
  * Upon <tt>SUCCESS</tt>, the receipt contains the `ScheduleID` of the created schedule. A schedule
- * entity includes a <tt>scheduledTransactionBody</tt> to be executed when the schedule has
- * collected enough signing Ed25519 keys to satisfy the scheduled transaction's signing
- * requirements. Upon `SUCCESS`, the receipt also includes the <tt>scheduledTransactionID</tt> to
+ * entity includes a <tt>scheduledTransactionBody</tt> to be executed.
+ * When the schedule has collected enough signing Ed25519 keys to satisfy the schedule's signing
+ * requirements, the schedule can be executed.
+ *
+ * If Long Term Scheduled Transactions are enabled and <tt>waitForExpiry</tt> is set to <tt>true</tt>, then the schedule
+ * will execute at it's <tt>expirationTime</tt>.
+ *
+ * Otherwise it will execute immediately after the transaction that provided enough Ed25519 keys, a <tt>ScheduleCreate</tt>
+ * or <tt>ScheduleSign</tt>.
+ *
+ * Upon `SUCCESS`, the receipt also includes the <tt>scheduledTransactionID</tt> to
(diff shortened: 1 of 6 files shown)


services/schedule_create.proto
```diff
@@ -90,4 +100,33 @@ message ScheduleCreateTransactionBody {
...
   AccountID payerAccountID = 4;
+
+  /**
+   * An optional timestamp for specifying when the transaction should be evaluated for execution and then expire.
+   * Defaults to 30 minutes after the transaction's consensus timestamp.
+   *
+   * Note: This field is unused and forced to be unset until Long Term Scheduled Transactions are enabled - Transactions will always
+   *       expire in 30 minutes if Long Term Scheduled Transactions are not enabled.
+   */
+  Timestamp expirationTime = 5;
```
steven-sheehy, line 111: We should use snake_case for new fields to provide better compatibility across languages.

services/schedule_create.proto
```diff
@@ -90,4 +100,33 @@ message ScheduleCreateTransactionBody {
...
+   * to sign the contained transaction. During the initial ScheduleCreate transaction or via ScheduleSign transactions.
+   *
+   * Defaults to false.
+   *
+   * Setting this to false does not necessarily mean that the transaction will never execute at expirationTime.
+   *  For Example - If the signature requirements for a Scheduled Transaction change via external means (e.g. CryptoUpdate)
+   *  such that the Scheduled Transaction would be allowed to execute, it will do so autonomously at expirationTime, unless a
+   *  ScheduleSign comes in to “poke” it and force it to go through immediately.
+   *
+   * Note: This field is unused and forced to be unset until Long Term Scheduled Transactions are enabled. Before Long Term
```
steven-sheehy, line 126: I have a concern about adding fields to the protobuf if we don't intend to use them soon. We've done this a lot in the past and the API is littered with unused fields. Even worse is if we add them then later find out we need something different like `proxyAccountId` for staking. If this isn't going in 0.26 then we should consider not putting these in the protobuf until later.

statop: I'm very sure this is what we need. If we want to wait for 0.27 I guess we can. @SimiHunjan ?

services/schedule_create.proto
statop: the hip uses camel case, since all the existing fields use it what benefit would be to changing now? I can change it, just curious.

services/schedule_create.proto
SimiHunjan: Can we make the changes in the feature branch and merge to develop/main when ready for the 0.27 release cycle?

services/schedule_create.proto
steven-sheehy: It's the standard we [agreed](https://github.com/hashgraph/hedera-protobufs/pull/13) upon for the protobuf repo. We just haven't had time to convert all the existing fields. This is also Google's [recommendation](https://developers.google.com/protocol-buffers/docs/style).

statop: done

services/schedule_create.proto
statop: Had call with @steven-sheehy and others, we decided that since this is going in 0.26 it can be merged now.

steven-sheehy (approved):
LGTM

SimiHunjan (approved):

PR #193, opened 2022-04-25
Add deprecated tag to deprecated response codes

Neeharika-Sompalli:
Signed-off-by: Neeharika-Sompalli <neeharika.sompalli@hedera.com>

Since , we accepted duplicate entries in allowance transaction bodies , adding `deprecated =true` for all old response codes. These are not being deleted because some old version of allowance changes used them

### DIFF
--- a/services/response_code.proto
+++ b/services/response_code.proto
@@ -1155,9 +1155,9 @@ enum ResponseCodeEnum {
   NEGATIVE_ALLOWANCE_AMOUNT = 290;
 
   /**
-   * The approveForAll flag cannot be set for a fungible token.
+   * [Deprecated] The approveForAll flag cannot be set for a fungible token.
    */
-  CANNOT_APPROVE_FOR_ALL_FUNGIBLE_COMMON = 291;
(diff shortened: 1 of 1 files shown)


superboo (approved):
LGTM

anighanta (approved):

services/response_code.proto
```diff
@@ -1155,9 +1155,9 @@ enum ResponseCodeEnum {
   NEGATIVE_ALLOWANCE_AMOUNT = 290;
 
   /**
-   * The approveForAll flag cannot be set for a fungible token.
+   * [Deprecated] The approveForAll flag cannot be set for a fungible token.
    */
-  CANNOT_APPROVE_FOR_ALL_FUNGIBLE_COMMON = 291;
+  CANNOT_APPROVE_FOR_ALL_FUNGIBLE_COMMON = 291 [deprecated = true];
```
xin-hedera, line 1160: what will we return if the token in NftAllowance with `approved_for_all = true`  is a fungible token? `FUNGIBLE_TOKEN_IN_NFT_ALLOWANCES`?

Neeharika-Sompalli: That's correct @xin-hedera

steven-sheehy (approved):

Issue #175, opened 2022-03-29
Deprecate staking-related protobuf fields not applicable to HIP-406 spec


tinker-michaelj:
### Problem

We need to announce deprecation of several protobuf fields that will not be used in the actual implementation of HIP-406.

### Solution

- [x] Deprecate `NodeAddress#stake`
- [x] Deprecate `ContractCreateTransactionBody#proxyAccountID`
- [x] Deprecate `ContractUpdateTransactionBody#proxyAccountID`
- [x] Deprecate `CryptoCreateTransactionBody#proxyAccountID`
- [x] Deprecate `CryptoUpdateTransactionBody#proxyAccountID`
- [x] Deprecate `CryptoGetInfoResponse.AccountInfo#proxyAccountID`
- [x] Deprecate `GetAccountDetailsResponse.AccountDetails#proxyAccountID`

Issue #174, opened 2022-03-29
Externalize staking info in `*GetInfo` queries


tinker-michaelj:
### Problem

A user should be able to review the staking metadata of an account or contract via `getAccountInfo` and `getContractInfo` queries:

- [x] `bool decline_reward`
- [x] `int64 stake_period_start`
- [x] `int64 pending_reward` (that is, the tinybar that would be received now in a reward situation)
- [x] `int64 staked_to_me`
- [x] `int64 staked_node_id`
- [x] `AccountID staked_account_id`

### Solution

Update the obvious protobuf messages to support the above fields.

Issue #173, opened 2022-03-29
Update `ContractUpdateTransactionBody` for staking 


tinker-michaelj:
### Problem

A user should be able to update a contract with a cryptographic `adminKey` with new staking configuration that includes:
- [ ] `google.protobuf.Int64Value staked_node_id` **--or--** `AccountID staked_account_id`
- [ ] `google.protobuf.BoolValue decline_reward`

A user should **NOT** be allowed to update the deprecated `AccountID proxyAccountID` field. (Such a `ContractUpdate` should fail with `PROXY_ACCOUNT_ID_FIELD_IS_DEPRECATED`.)

### Solution

Update the obvious protobuf messages to support the above fields.

Issue #170, opened 2022-03-29
Update `CryptoCreateTransactionBody` for staking 


tinker-michaelj:
### Problem

A user should be able to create an account with staking configuration that includes:
- [x] `google.protobuf.Int64Value staked_node_id` **--or--** `AccountID staked_account_id`
- [x] `google.protobuf.BoolValue decline_reward`

A user should **NOT** be allowed to use the deprecated `AccountID proxyAccountID` field. (Such a `CryptoCreate` should fail with `PROXY_ACCOUNT_ID_FIELD_IS_DEPRECATED`.)

### Solution

Update the obvious protobuf messages to support the above fields.

PR #195, opened 2022-05-04
Staking changes

### LINKED ISSUE #169: Itemize paid staking rewards in `TransactionRecord`
### Problem

Since [any transaction](https://github.com/hashgraph/hedera-improvement-proposal/blob/87ace24459a44f6930f63098d07b57404cdfea03/HIP/hip-406.md) can trigger reward payments from `0.0.800`, and the network will always redirect part of the service fee into `0.0.800`, for multi-party transactions it may be hard to see how staking credits and debits affected the aggregated `TransactionRecord#transferList`.

### Solution

Add a `TransactionRecord` field,
```
repeated AccountAmount paid_staking_rewards = 20;
```

Just as with the `assessed_custom_fees` field, this will **not** change the semantics of the `transferList`. (The `transferList` still shows the net change in ℏ balance for each account touched in the transaction.)

The purpose of this field is just to explain why a given `TransactionBody` resulted in its observed record---since depending on the status of pending rewards, the exact same `TransactionBody` could lead to very records.

### Alternatives

_No response_

Neeharika-Sompalli:
Fixes #169  #170  #171 #172 #173 #174 #175

- Deprecated `proxyAccountID` in `(Crypto/Contract)CreateTransactionBody` , `(Crypto/Contract)UpdateTransactionBody`, (Crypto/Contract)GetInfo
- Deprecated `NodeAddress#stake`
- Added new staking fields to `(Crypto/Contract)CreateTransactionBody` , `(Crypto/Contract)UpdateTransactionBody`
- Added `StakingInfo` and returned in `(Crypto/Contract)GetInfo` query
- Added `paid_staking_rewards` to `TransactionRecord`
- Added ResponseCodes needed

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1546,7 +1546,7 @@ message StakingInfo {
     /**
      * The staking period during which either the staking settings for this account or contract changed (such as starting
      * staking or changing staked_node_id) or the most recent reward was earned, whichever is later. If this account or contract
-     * is not currently staked to a node, then the value is -1.
+     * is not currently staked to a node, then this field is not set.
      */
     Timestamp stake_period_start = 2;
 
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1362,9 +1362,9 @@ message NodeAddress {
     string description = 9;
 
     /**
-     * The amount of tinybars staked to the node
+     * [Deprecated] The amount of tinybars staked to the node
      */
-    int64 stake = 10;
+    int64 stake = 10 [deprecated = true];
 }
 
 /**
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1548,7 +1548,7 @@ message StakingInfo {
      * staking or changing staked_node_id) or the most recent reward was earned, whichever is later. If this account or contract
      * is not currently staked to a node, then the value is -1.
      */
-    int64 stake_period_start = 2;
+    Timestamp stake_period_start = 2;
 
     /**
      * The amount in tinybars that will be received in the next reward situation.
@@ -1561,15 +1561,19 @@ message StakingInfo {
     int64 staked_to_me = 4;
 
     /**
-     * The ID of the node this account or contract is staked to. If this account or contract is not staked to a node,
-     * the value is -1. The default value is -1.
+     * ID of the account or node to which this account or contract is staking. It can never happen that staked_account_id is
+     * non-null and staked_node_id is non-negative at the same time.
      */
-    int64 staked_node_id = 5;
+    oneof staked_id {
 
-    /**
-     * The account to which this account or contract is staking. If not staked to an account, the value is null.
-     * The default value is null. It can never happen that staked_account_id is non-null and staked_node_id is
-     * non-negative at the same time.
-     */
-    AccountID staked_account_id = 6;
+        /**
+         * The account to which this account or contract is staking.The default value is null.
+         */
+        AccountID staked_account_id = 5;
+
+        /**
+         * The ID of the node this account or contract is staked to. The default value is -1.
+         */
+        int64 staked_node_id = 6;
+    }
 }
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1561,18 +1561,17 @@ message StakingInfo {
     int64 staked_to_me = 4;
 
     /**
-     * ID of the account or node to which this account or contract is staking. It can never happen that staked_account_id is
-     * non-null and staked_node_id is non-negative at the same time.
+     * ID of the account or node to which this account or contract is staking.
      */
     oneof staked_id {
 
         /**
-         * The account to which this account or contract is staking.The default value is null.
+         * The account to which this account or contract is staking.
          */
         AccountID staked_account_id = 5;
 
         /**
-         * The ID of the node this account or contract is staked to. The default value is -1.
+         * The ID of the node this account or contract is staked to.
          */
         int64 staked_node_id = 6;
     }
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1532,3 +1532,40 @@ message TokenAssociation {
     TokenID token_id = 1;  // The token involved in the association
     AccountID account_id = 2; // The account involved in the association
 }
+
+/* Staking metadata for an account or a contract */
+message StakingInfo {
+    /**
+     * If true, this account or contract declined to receive a staking reward.
+     */
+    bool decline_reward = 1;
+
+    /**
+     * The staking period during which either the staking settings for this account or contract changed (such as starting
+     * staking or changing stakedNode) or the most recent reward was earned, whichever is later. If this account or contract
+     * is not currently staked to a node, then the value is -1.
+     */
+    int64 stake_period_start = 2;
+
+    /**
+     * The amount in tinybars that will be received in the next reward situation.
+     */
+    int64 pending_reward = 3;
+
+    /**
+     * The sum of balances of all accounts staked to this account or contract.
+     */
+    int64 staked_to_me = 4;
+
+    /**
+     * The ID of the node this account or contract is staked to. If this contract is not staked to a node, the value is -1. The default value is -1.
+     */
+    int64 staked_node_id = 5;
+
+    /**
+     * The account to which the account or contract is staking. If not staked to an account, the value is null.
+     * The default value is null. It can never happen that staked_account_id is non-null and staked_node_id is
+     * non-negative at the same time.
+     */
+    AccountID staked_account_id = 6;
+}
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1533,8 +1533,11 @@ message TokenAssociation {
     AccountID account_id = 2; // The account involved in the association
 }
 
-/* Staking metadata for an account or a contract */
+/**
+ * Staking metadata for an account or a contract returned in CryptoGetInfo or ContractGetInfo queries
+ */
 message StakingInfo {
+
     /**
      * If true, this account or contract declined to receive a staking reward.
      */
@@ -1542,7 +1545,7 @@ message StakingInfo {
 
     /**
      * The staking period during which either the staking settings for this account or contract changed (such as starting
-     * staking or changing stakedNode) or the most recent reward was earned, whichever is later. If this account or contract
+     * staking or changing staked_node_id) or the most recent reward was earned, whichever is later. If this account or contract
      * is not currently staked to a node, then the value is -1.
      */
     int64 stake_period_start = 2;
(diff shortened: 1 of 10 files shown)


services/basic_types.proto
```diff
@@ -1532,3 +1532,44 @@ message TokenAssociation {
...
+     * The staking period during which either the staking settings for this account or contract changed (such as starting
+     * staking or changing staked_node_id) or the most recent reward was earned, whichever is later. If this account or contract
+     * is not currently staked to a node, then the value is -1.
+     */
+    int64 stake_period_start = 2;
+
+    /**
+     * The amount in tinybars that will be received in the next reward situation.
+     */
+    int64 pending_reward = 3;
```
anighanta, line 1556: @tinker-michaelj what are uses of this field ?

anighanta (approved):

services/contract_create.proto
```diff
@@ -177,4 +178,14 @@ message ContractCreateTransactionBody {
...
+
+    /**
+     * ID of the new account to which this contract is staking. If not staked to an account, the value is null. The default value is null.
+     */
+    AccountID staked_account_id = 17;
+
+    /**
+     * If true, the contract declines receiving a staking reward. The default value is false.
+     */
+    google.protobuf.BoolValue decline_reward = 18;
```
steven-sheehy, line 190: This should just be a primitive boolean on creates. Only updates need a wrapper. Same comment for other creates.

services/contract_create.proto
```diff
@@ -177,4 +178,14 @@ message ContractCreateTransactionBody {
      * cover auto-renewal fees.
      */
     AccountID auto_renew_account_id = 15;
+
+    /**
+     * ID of the new account to which this contract is staking. If not staked to an account, the value is null. The default value is null.
+     */
+    AccountID staked_account_id = 17;
```
steven-sheehy, line 185: I believe this should be a oneof with `int64 staked_node_id`. For updates, they can use the Int64Value wrapper. Same comment for other creates.

services/contract_update.proto
```diff
@@ -109,4 +109,14 @@ message ContractUpdateTransactionBody {
      * account. Otherwise it updates the contract's auto-renew account to the referenced account.
      */
     AccountID auto_renew_account_id = 12;
+
+    /**
+     * ID of the new account to which this contract is staking. If not staked to an account, the value is null. The default value is null.
+     */
+    AccountID staked_account_id = 13;
```
steven-sheehy, line 116: Is it possible to clear an existing `staked_account_id` or `staked_node_id`? How would that look like with a `oneof`?

steven-sheehy (changes requested):

services/contract_create.proto
Neeharika-Sompalli: Fixed. Thanks @steven-sheehy

services/contract_create.proto
Neeharika-Sompalli: Fixed. Thanks @steven-sheehy

services/contract_update.proto
Neeharika-Sompalli: I have updated to use `oneof` with the `Int64Value` Wrapper. And in the implementation we will make sure only one filed is set.

services/contract_create.proto
```diff
@@ -166,15 +166,38 @@ message ContractCreateTransactionBody {
...
+    oneof staked_id {
+
+        /**
+         * ID of the account to which this contract is staking. If not staked to an account, the value is null.
+         * The default value is null.
+         */
+        AccountID staked_account_id = 17;
+
+        /**
+         * ID of the node this contract is staked to. If not staked to a node, the value is -1. The default value is -1.
```
xin-hedera, line 194: Since it's in wrapped in a `oneof`, no need to have a sentinel value to indicate it's not staked to a node.

```suggestion
         * ID of the node this contract is staked to.
```

services/contract_update.proto
```diff
@@ -109,4 +109,27 @@ message ContractUpdateTransactionBody {
...
+        AccountID staked_account_id = 13;
+
+        /**
+         * ID of the new node this contract is staked to. If not staked to a node, the value is -1. The default value is -1.
+         */
+        google.protobuf.Int64Value staked_node_id = 14;
+    }
+
+    /**
+     * If true, the contract declines receiving a staking reward. The default value is false.
```
xin-hedera, line 132: Since it's BoolValue, the default is the field is not set instead of `false`.

```suggestion
     * If true, the contract declines receiving a staking reward.
```

services/basic_types.proto
```diff
@@ -1532,3 +1532,44 @@ message TokenAssociation {
...
+     * If true, this account or contract declined to receive a staking reward.
+     */
+    bool decline_reward = 1;
+
+    /**
+     * The staking period during which either the staking settings for this account or contract changed (such as starting
+     * staking or changing staked_node_id) or the most recent reward was earned, whichever is later. If this account or contract
+     * is not currently staked to a node, then the value is -1.
+     */
+    int64 stake_period_start = 2;
```
xin-hedera, line 1551: should this be a `Timestamp` instead?

services/basic_types.proto
```diff
@@ -1532,3 +1532,44 @@ message TokenAssociation {
...
+     * the value is -1. The default value is -1.
+     */
+    int64 staked_node_id = 5;
+
+    /**
+     * The account to which this account or contract is staking. If not staked to an account, the value is null.
+     * The default value is null. It can never happen that staked_account_id is non-null and staked_node_id is
+     * non-negative at the same time.
+     */
+    AccountID staked_account_id = 6;
```
xin-hedera, line 1574: make these into the same oneof staked_id as in Contract / Crypto create transaction body.

Neeharika-Sompalli: Fixed. Thanks @xin-hedera

services/basic_types.proto
Neeharika-Sompalli: Fixed. Thanks @xin-hedera

services/contract_update.proto
Neeharika-Sompalli: Left the default value in place and fixed

services/contract_create.proto
Neeharika-Sompalli: Left the default value in place and fixed

services/contract_update.proto
steven-sheehy: But the default value is unset not false. If you don't supply an explicit value the field will not be updated.

services/contract_create.proto
steven-sheehy: Again, the default is not -1 it's unset. The comment should remove the default. Same thing for staked_account_id.

services/contract_update.proto
```diff
@@ -109,4 +109,26 @@ message ContractUpdateTransactionBody {
...
     AccountID auto_renew_account_id = 12;
+
+    /**
+     * ID of the new account or node to which this contract is staking. It can never happen that staked_account_id
+     * is non-null and staked_node_id is non-negative at the same time.
+     */
+    oneof staked_id {
+
+        /**
+         * ID of the new account to which this contract is staking. The default value is null.
```
steven-sheehy, line 120: I don't think the oneof by itself provides a way to clear a value. So we still need to call `contractUpdateBuilder.setStakedAccountId(AccountID.getDefaultInstance())` to indicate both that you want to changed the staked ID and that it should be cleared.

```suggestion
         * ID of the new account to which this contract is staking. If set to the sentinel <tt>0.0.0</tt> AccountID, this field removes the contract's staked account.
```

steven-sheehy (changes requested):

services/crypto_update.proto
```diff
@@ -129,4 +129,27 @@ message CryptoUpdateTransactionBody {
...
+         * ID of the new account to which this account is staking. If set to the sentinel <tt>0.0.0</tt> AccountID,
+         * this field removes this account's staked account ID.
+         */
+        AccountID staked_account_id = 16;
+
+        /**
+         * ID of the new node this account is staked to. If set to the sentinel <tt>-1</tt>, this field
+         * removes this account's staked node ID.
+         */
+        google.protobuf.Int64Value staked_node_id = 17;
```
steven-sheehy, line 148: I did some local testing. Having an `Int64Value` does not provide any value here. The wrapper type's main purpose is to allow set versus unset semantics for primitive types. But here, the `oneof` already provides that capability since by default the `oneof` has a state of unset. The wrapper does not help with being able to clear a value since it will throw a NPE if you try to set staked_account_id or stake_node_id to null in the builder. Recommend we change back to primitive for updates:

```suggestion
        int64 staked_node_id = 17;
```

services/contract_create.proto
```diff
@@ -166,15 +166,37 @@ message ContractCreateTransactionBody {
...
-     * account with zero hbar balance, the contract's own hbar balance will be used to 
+     * An account to charge for auto-renewal of this contract. If not set, or set to an
+     * account with zero hbar balance, the contract's own hbar balance will be used to
      * cover auto-renewal fees.
      */
     AccountID auto_renew_account_id = 15;
+
+    /**
+     * ID of the new account or node to which this contract is staking. It can never happen that staked_account_id is
+     * non-null and staked_node_id is non-negative at the same time.
```
steven-sheehy, line 183:
```suggestion
     * ID of the new account or node to which this contract is staking.
```

steven-sheehy (changes requested):

services/contract_update.proto
```diff
@@ -109,4 +109,27 @@ message ContractUpdateTransactionBody {
...
+         * ID of the new account to which this contract is staking. If set to the sentinel <tt>0.0.0</tt> AccountID,
+         * this field removes the contract's staked account ID.
+         */
+        AccountID staked_account_id = 13;
+
+        /**
+         * ID of the new node this contract is staked to. If set to the sentinel <tt>-1</tt>, this field
+         * removes the contract's staked node ID.
+         */
+        google.protobuf.Int64Value staked_node_id = 14;
```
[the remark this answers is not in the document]
Neeharika-Sompalli: @steven-sheehy  Updated

xin-hedera (dismissed):
LGTM

steven-sheehy (dismissed):
LGTM

SimiHunjan (dismissed):

services/basic_types.proto
tinker-michaelj: It lets a user see how much ℏ their account has as a "pending" reward that will be paid in the next reward situation.

services/basic_types.proto
```diff
@@ -1532,3 +1532,47 @@ message TokenAssociation {
...
+
+    /**
+     * If true, this account or contract declined to receive a staking reward.
+     */
+    bool decline_reward = 1;
+
+    /**
+     * The staking period during which either the staking settings for this account or contract changed (such as starting
+     * staking or changing staked_node_id) or the most recent reward was earned, whichever is later. If this account or contract
+     * is not currently staked to a node, then the value is -1.
```
tinker-michaelj, line 1549: Maybe, `...then this field is not set`.

Neeharika-Sompalli: Fixed @tinker-michaelj  ! Thanks

tinker-michaelj (approved):
LGTM, very precise 😎

xin-hedera (approved):

PR #202, opened 2022-05-19
Record Stream V6 Protobufs

ivanparnarev:
Protobuf class implementation for Record Stream V6.

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -6,7 +6,7 @@ package proto;
  * ‌
  * Hedera Network Services Protobuf
  * ​
- * Copyright (C) 2018 - 2021 Hedera Hashgraph, LLC
+ * Copyright (C) 2018 - 2022 Hedera Hashgraph, LLC
  * ​
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/streams/hash_object.proto
+++ b/streams/hash_object.proto
@@ -1,0 +1,56 @@
+syntax = "proto3";
+
+package proto;
+
+/*-
+ * ‌
+ * Hedera Network Services Protobuf
+ * ​
+ * Copyright (C) 2018 - 2022 Hedera Hashgraph, LLC
+ * ​
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ‍
+ */
+
+option java_package = "com.hedera.services.stream.proto";
+option java_multiple_files = true;
+
+/**
+ * List of hash algorithms
+ */
+enum HashAlgorithm {
+  HASH_ALGORITHM_UNKNOWN = 0;
+  SHA_384 = 1;
+}
+
+/**
+ * Encapsulates an object hash so that additional hash algorithms
+ * can be added in the future without requiring a breaking change.
+ */
+message HashObject {
+
+  /**
+   * Specifies the hashing algorithm
+   */
+  HashAlgorithm algorithm = 1;
+
+  /**
+   * Hash length
+   */
+  int32 length = 2;
+
+  /**
+   * Specifies the result of the hashing operation in bytes
+   */
+  bytes hash = 3;
+}
--- a/streams/record_stream_file.proto
+++ b/streams/record_stream_file.proto
@@ -1,0 +1,106 @@
+syntax = "proto3";
+
+package proto;
+
+/*-
+ * ‌
+ * Hedera Network Services Protobuf
+ * ​
+ * Copyright (C) 2018 - 2022 Hedera Hashgraph, LLC
+ * ​
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ‍
+ */
+
+option java_package = "com.hedera.services.stream.proto";
+option java_multiple_files = true;
+
+import "basic_types.proto";
+import "transaction.proto";
+import "transaction_record.proto";
+import "hash_object.proto";
+
+/**
+ * RecordStreamFile is used to serialize all RecordStreamItems that are part of the
+ * same period into record stream files.
+ * This structure represents a block in Hedera (HIP-415).
+ */
+message RecordStreamFile {
+  /**
+   * Version of HAPI that was used to serialize the file.
+   */
+  SemanticVersion hapi_proto_version = 1;
+
+  /**
+   * Running Hash of all RecordStreamItems before writing this file.
+   */
+  HashObject start_object_running_hash = 2;
+
+  /**
+   * List of all the record stream items from that period.
+   */
+  repeated RecordStreamItem record_stream_items = 3;
+
+  /**
+   * Running Hash of all RecordStreamItems before closing this file.
+   */
+  HashObject end_object_running_hash = 4;
+
+  /**
+   * The block number associated with this period.
+   */
+  int64 block_number = 5;
+
+  /**
+   * List of the hashes of all the sidecar record files created for the same period.
+   * Allows multiple sidecar files to be linked to this RecordStreamFile.
+   */
+  repeated SidecarMetadata sidecars = 6;
+}
+
+/**
+ * A RecordStreamItem consists of a Transaction and a TransactionRecord,
+ * which are already defined protobuf messages.
+ */
+message RecordStreamItem {
+  Transaction transaction = 1;
+  TransactionRecord record = 2;
+}
+
+/**
+ * Information about a single sidecar file.
+ */
+message SidecarMetadata {
+  /**
+   * The hash of the entire file.
+   */
+  HashObject hash = 1;
+
+  /**
+   * The id of the sidecar record file
+   */
+  int32 id = 2;
+
+  /**
+   * The types of sidecar records that will be included in the file.
+   */
+  repeated SidecarType types = 3;
+}
+
+/**
+ * The type of sidecar records contained in the sidecar record file
+ */
+enum SidecarType {
+  SIDECAR_TYPE_UNKNOWN = 0;
+  CONTRACT_STATE_CHANGE = 1;
+}
--- a/streams/sidecar_file.proto
+++ b/streams/sidecar_file.proto
@@ -1,0 +1,67 @@
+syntax = "proto3";
+
+package proto;
+
+/*-
+ * ‌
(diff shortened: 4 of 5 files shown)


services/sidecar_record.proto
```diff
@@ -0,0 +1,101 @@
...
+option java_multiple_files = true;
+
+import "basic_types.proto";
+import "timestamp.proto";
+
+/**
+ * A SidecarFile contains a list of TransactionSidecarRecords that are all created
+ * in the same period and related to the same RecordStreamFile.
+ */
+message SidecarFile {
```
steven-sheehy, line 35: Should name the file to match the message inside.

`sidecar_record.proto` -> `sidecar_file.proto`

services/signature.proto
```diff
@@ -0,0 +1,83 @@
...
+option java_package = "com.hederahashgraph.api.proto.java";
+option java_multiple_files = true;
+
+import "basic_types.proto";
+
+/**
+ * The record signature file which is created for each record stream file
+ * that signs the hash of the entire corresponding stream file.
+ */
+message SignatureFile {
```
steven-sheehy, line 34: Name file to match. `signature_file.proto`

services/sidecar_record.proto
```diff
@@ -0,0 +1,101 @@
...
+ * Contains a list of ContractStateChanges
+ */
+message ContractStateChanges {
+  repeated ContractStateChange stateChanges = 1;
+}
+
+/**
+ * Defines the structure of a sidecar record for Contract State Changes
+ */
+message ContractStateChange{}
```
steven-sheehy, line 53: There already exists a `ContractStateChange` defined in `contract_call_local.proto` that we should reuse here.

services/sidecar_record.proto
```diff
@@ -0,0 +1,101 @@
...
+ * Contains a list of ContractActions
+ */
+message ContractActions {
+  repeated ContractAction actions = 1;
+}
+
+/**
+ * Defines the structure of a sidecar record for Contract Actions
+ */
+message ContractAction {}
```
steven-sheehy, line 65: We should probably omit `ContractActions` and `ContractBytecodes` until we have a concrete implementation for them. When we go to actually implement them we may find a different name more suitable.

services/sidecar_record.proto
```diff
@@ -0,0 +1,101 @@
...
+  Timestamp consensusTimestamp = 1;
+
+  /*
+  * List of sidecar types.
+  * In future there will be other categories.
+  */
+  oneof sidecar_records {
+    ContractStateChanges stateChanges = 2;
+    ContractActions actions = 3;
+    ContractBytecodes bytecodes = 4;
```
steven-sheehy, line 98: Use snake_case here and elsewhere.

services/signature.proto
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * A Signature defined by its type, length, checksum and signature bytes and the hash that is signed
+ */
+message SignatureObject {
+
+  /**
+   * The signature type
+   */
+  enum SignatureType {
+    SHA_384_WITH_RSA = 0;
```
steven-sheehy, line 56: Is this really considered field=0 within `SignatureObject`? Maybe we should move its definition outside of `SignatureObject` so `type = 0`?

services/basic_types.proto
```diff
@@ -1576,3 +1576,31 @@ message StakingInfo {
...
+
+/**
+ * A Hash Object consisting of algorithm used, hash length and the hash bytes
+ */
+message HashObject {
+
+    /**
+     * List of hash algorithms
+     */
+    enum HashAlgorithm {
```
steven-sheehy, line 1588: Perhaps need to move out of `HashObject`.

services/sidecar_record.proto
shemnon: Here's what I need for `ContractAction` and supporting types:

https://github.com/hashgraph/hedera-protobufs-java/blob/smartcontract-mirror/src/main/proto/contract_call_local.proto#L168 - `ContractActionType`
https://github.com/hashgraph/hedera-protobufs-java/blob/smartcontract-mirror/src/main/proto/contract_call_local.proto#L201 - `ContractAction`

I have old POC code that populates these.

shemnon: This tracability information was the reason for sidcars.  I don' see it as expanding the scope of the hip.  It can be moved to a new HIP as it is orthagonal to generating sidecars but this should ship with v6.

steven-sheehy: I disagree. Mirror node doesn't have bandwidth to take on sidecar work as I explained in a meeting. We agreed it would be deferred to a future release.

services/signature.proto
ivanparnarev: I moved the definitions of the enum outside the messages, but still the type cannot be 0.

https://developers.google.com/protocol-buffers/docs/proto3#assigning_field_numbers

Moreover, there is an error "test.proto:2:28: Field numbers must be positive integers." when trying to assign the field to 0.

services/sidecar_record.proto
ivanparnarev: ContractActions and ContractBytecodes are omitted. The main goal of the PR is to implement the core protobufs. The other types will be added in future HIP.

streams/record_stream_file.proto
```diff
@@ -0,0 +1,103 @@
...
+
+  /**
+   * Running Hash of all RecordStreamItems before writing this file.
+   */
+  HashObject start_object_running_hash = 2;
+
+  /**
+   * List of all the record file objects from that period.
+   */
+  repeated RecordStreamItem record_file_objects = 3;
```
stoqnkpL, line 50: nit: rename the field
```suggestion
  repeated RecordStreamItem record_stream_items = 3;
```

streams/sidecar_file.proto
```diff
@@ -0,0 +1,64 @@
...
+   * way to match record to sidecar.
+   */
+  Timestamp consensus_timestamp = 1;
+
+  /*
+  * List of sidecar types.
+  * In future there will be other categories.
+  */
+  oneof sidecar_records {
+    ContractStateChange state_changes = 2;
```
stoqnkpL, line 61: This should be a list of contract state changes. We may need to define it in a separate message like:
message ContractStateChanges {
  repeated ContractStateChange contract_state_changes = 1;
}

stoqnkpL (changes requested):

services/basic_types.proto
```diff
@@ -1576,3 +1576,31 @@ message StakingInfo {
...
+ * List of hash algorithms
+ */
+enum HashAlgorithm {
+    SHA_384 = 0;
+}
+
+/**
+ * A Hash Object consisting of algorithm used, hash length and the hash bytes
+ */
+message HashObject {
```
steven-sheehy, line 1590: This message and enum should move to a `streams/hash_object.proto`

streams/hash_object.proto
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ‍
+ */
+
+option java_package = "com.hederahashgraph.api.proto.java";
```
steven-sheehy, line 25: Should match the balance file package. Same comment for other files.
```suggestion
option java_package = "com.hedera.services.stream.proto";
```

streams/hash_object.proto
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * List of hash algorithms
+ */
+enum HashAlgorithm {
+  HASH_UNKNOWN = 0;
+  SHA_384 = 1;
+}
+
+/**
+ * A Hash Object consisting of algorithm used, hash length and the hash bytes
```
steven-sheehy, line 37: Listing the fields is not very useful and would require changes later to keep it in sync. Instead we should explain why a wrapper object is useful.

```suggestion
 * Encapsulates an object hash so that additional hash algorithms can be added in the future without requiring a breaking change.
```

streams/record_stream_file.proto
```diff
@@ -0,0 +1,105 @@
...
+   */
+  SemanticVersion hapi_proto_version = 1;
+
+  /**
+   * Running Hash of all RecordStreamItems before writing this file.
+   */
+  HashObject start_object_running_hash = 2;
+
+  /**
+   * List of all the record file objects from that period.
```
steven-sheehy, line 49:
```suggestion
   * List of all the record stream items from that period.
```

streams/record_stream_file.proto
```diff
@@ -0,0 +1,105 @@
...
+option java_multiple_files = true;
+
+import "basic_types.proto";
+import "transaction.proto";
+import "transaction_record.proto";
+import "hash_object.proto";
+
+/**
+ * RecordStreamFile is used to serialize all RecordStreamItems that are part of the
+ * same period into record stream files.
```
steven-sheehy, line 35: Should we mention that this structure represents a block in Hedera as defined in HIP-415?

streams/record_stream_file.proto
```diff
@@ -0,0 +1,105 @@
...
+message SidecarMetadata {
+  /**
+   * The hash of the entire file.
+   */
+  HashObject hash = 1;
+
+  /**
+   * The id of the sidecar record file
+   */
+  int64 id = 2;
```
steven-sheehy, line 91: Perhaps int32 is sufficient. There's no way we could support uploading more than a 100 sidecar files.

streams/record_stream_file.proto
```diff
@@ -0,0 +1,105 @@
...
+   */
+  repeated SidecarType types = 3;
+}
+
+/**
+ * The type of sidecar records contained in the sidecar record file
+ */
+enum SidecarType {
+  SIDECAR_TYPE_UNKNOWN = 0;
+  ContractStateChangeType = 1;
```
steven-sheehy, line 104: The mixture of cases is weird. Suggest we use upper snake case for enum values. Also we don't need `SidecarType` or `Type` in the value since that's implied from the enum context.

```suggestion
  UNKNOWN = 0;
  CONTRACT_STATE_CHANGE = 1;
```

steven-sheehy (changes requested):

steven-sheehy (dismissed):
Going to clear my request for changes since I will be out of office. But please fix the java namespace for all files before merging.

streams/hash_object.proto
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+option java_package = "com.hedera.services.stream.proto";
+option java_multiple_files = true;
+
+/**
+ * List of hash algorithms
+ */
+enum HashAlgorithm {
+  HASH_UNKNOWN = 0;
```
xin-hedera, line 32: nit:
```suggestion
  UNKNOWN = 0;
```

streams/signature_file.proto
```diff
@@ -0,0 +1,84 @@
...
+   * The hash that is signed by this signature
+   */
+  HashObject hash_object = 5;
+}
+
+/**
+ * The signature type
+ */
+enum SignatureType {
+  SIGNATURE_TYPE_UNKNOWN = 0;
```
xin-hedera, line 82: nit:

```suggestion
  UNKNOWN = 0;
```

ivanparnarev: I tried making each first value to be "UNKNOWN", but the IDE is giving error "'proto.UNKNOWN' is already defined in file 'hash_object.proto'". That is the reason, why the values are different.

ivanparnarev: I consider this as a proper way of dealing with the problem:

https://github.com/protobuf-net/protobuf-net/issues/60#issuecomment-461509160

We will preserve the name, but the generated class will point to explicit value.

I will commit the change once I receive your feedback.

steven-sheehy (approved):

stoqnkpL (approved):
LGTM!

steven-sheehy (approved):

tinker-michaelj (approved):
Makes sense to me, nice work!

SimiHunjan (approved):

steven-sheehy (approved):

PR #204, opened 2022-06-06
Publish reward rates per node

tinker-michaelj:
**Description**:
 - Moves `reward_rate` into the `NodeStake` message and adds more detail in comments.
 - Fixes descriptions of `stake_rewarded` and `stake` fields per @xin-hedera.

### DIFF
--- a/services/node_stake_update.proto
+++ b/services/node_stake_update.proto
@@ -53,12 +53,12 @@ message NodeStake {
   int64 node_id = 1;
 
   /**
-   * Consensus weight for this node, at the end of current staking period.
+   * Consensus weight of this node for the new staking period.
    */
   int64 stake = 2;
 
   /**
-   * Total of balance of all accounts staked to this node that have declineReward set to false, at the end of staking period.
+   * Total of (balance + stakedToMe) for all accounts staked to this node with declineReward=false, at the beginning of the new staking period.
    */
   int64 stake_rewarded = 3;
 
--- a/services/node_stake_update.proto
+++ b/services/node_stake_update.proto
@@ -38,21 +38,15 @@ message NodeStakeUpdateTransactionBody {
   Timestamp end_of_staking_period = 1;
 
   /**
-   * The daily reward rate for this staking period in tinybars
-   */
-  int64 reward_rate = 2;
-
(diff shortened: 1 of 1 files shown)


services/node_stake_update.proto
```diff
@@ -37,34 +37,43 @@ message NodeStakeUpdateTransactionBody {
...
+  /**
+   * The reward rate _per whole hbar_ that was staked to this node with declineReward=false from the start of 
+   * the staking period that is ending. 
+   * 
+   * There are three main cases to consider:
+   *   1. The reward rate is zero because the node's starting stake was below its minStake.
+   *   2. The node's stake was in the range [minStake, maxStake], and this is the "base" reward rate; namely,
+   *           baseRewardRate = totalRewardRate / (totalStakeRewardStart / 100_000_000)
+   *   3. The node's stake exceeded its maxStake, so this is a down-scaling of the base rate:
+   *           scaledRewardRate = (baseRewardRate * stakeToReward) / maxStake
```
xin-hedera, line 74: I think the condition for 3 should be: the node's stakeToReward at the beginning of the ending staking period (let's name it `stakeToRewardStartNotCapped`) exceeds its maxStake. And to downscale it, we should do

```
scaledRewardRate = baseRewardRate * maxStake / stakeToRewardStartNotCapped 
```

steven-sheehy: But perhaps we don't need to list the 3 conditions as these are internal implementation details? We would only need to list details that would cause consumers to have to react differently, but here I think for all 3 the mirror node would still simply store the value and aggregate for use in a pending reward calculation.

tinker-michaelj: Haha @xin-hedera I was trying to just mindlessly copy from [the source](https://github.com/hashgraph/hedera-services/blob/redirect-deleted-account-rewards/hedera-node/src/main/java/com/hedera/services/state/merkle/MerkleStakingInfo.java#L212) and somehow inverted the fraction in the process. 😝 

Deleted these implementations details and rearranged fields alphabetically.

steven-sheehy (approved):

xin-hedera (approved):

PR #209, opened 2022-06-21
HIP-351 : RandomGenerate transaction

Neeharika-Sompalli:
Addresses [HIP-351](https://hips.hedera.com/hip/hip-351) changes

1. Added a top level `RandomGenerate` grpc end point and `HederaFunctionality`
2. Add `random_generated_number` to transaction record

### DIFF
--- a/services/crypto_service.proto
+++ b/services/crypto_service.proto
@@ -63,7 +63,7 @@ service CryptoService {
     rpc deleteAllowances (Transaction) returns (TransactionResponse);
 
     /**
-     * Generates a pseudorandom number for use by both smart contracts and users.
+     * Generates a pseudorandom number.
      */
     rpc randomGenerate (Transaction) returns (TransactionResponse);
 
--- a/services/crypto_service.proto
+++ b/services/crypto_service.proto
@@ -63,11 +63,6 @@ service CryptoService {
     rpc deleteAllowances (Transaction) returns (TransactionResponse);
 
     /**
-     * Generates a pseudorandom number.
-     */
-    rpc randomGenerate (Transaction) returns (TransactionResponse);
-
-    /**
      * (NOT CURRENTLY SUPPORTED) Adds a livehash
      */
     rpc addLiveHash (Transaction) returns (TransactionResponse);
--- a/services/crypto_service.proto
+++ b/services/crypto_service.proto
@@ -63,6 +63,11 @@ service CryptoService {
     rpc deleteAllowances (Transaction) returns (TransactionResponse);
 
     /**
+     * Generates a pseudorandom number for use by both smart contracts and users.
+     */
+    rpc randomGenerate (Transaction) returns (TransactionResponse);
+
+    /**
      * (NOT CURRENTLY SUPPORTED) Adds a livehash
      */
     rpc addLiveHash (Transaction) returns (TransactionResponse);
--- a/services/random_generate.proto
+++ b/services/random_generate.proto
@@ -31,7 +31,7 @@ option java_multiple_files = true;
 message RandomGenerateTransactionBody {
   /**
    * If provided and is positive, returns a 32-bit pseudorandom number from the given range in the transaction record.
-   * If not set, will return a 384-bit pseudorandom number in the record.
+   * If not set or set to zero, will return a 384-bit pseudorandom number in the record.
    */
   int32 range = 1;
 }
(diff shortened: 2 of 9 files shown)


services/schedulable_transaction_body.proto
```diff
@@ -276,5 +277,10 @@ message SchedulableTransactionBody {
      * Marks a schedule in the network's action queue as deleted, preventing it from executing
      */
     ScheduleDeleteTransactionBody scheduleDelete = 34;
+
+    /**
+     * Generates a pseudorandom number.
+     */
+    RandomGenerateTransactionBody randomGenerate = 40;
```
anighanta, line 284: snake case here

anighanta (dismissed):
LGTM other than minor comments

services/schedulable_transaction_body.proto
Neeharika-Sompalli: Fixed ! Thanks

services/transaction_body.proto
```diff
@@ -343,5 +344,10 @@ message TransactionBody {
      * Updates the staking info at the end of staking period to indicate new staking period has started.
      */
     NodeStakeUpdateTransactionBody node_stake_update = 51;
+
+    /**
+     * Generates a pseudorandom number.
+     */
+    RandomGenerateTransactionBody randomGenerate = 52;
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Fixed ! Thanks

services/transaction_record.proto
```diff
@@ -133,4 +133,10 @@ message TransactionRecord {
      * List of accounts with the corresponding staking rewards paid as a result of a transaction.
      */
     repeated AccountAmount paid_staking_rewards = 18;
+
+    /**
+     * An array of 384-bit pseudorandom bytes in the record of RandomGenerate transaction.
+     * It will be a 32-bit pseudorandom integer if a positive range is provided in the transaction.
+     */
+    bytes random_generated_number = 19;
```
steven-sheehy, line 141: Since it's not necessarily a number.

```suggestion
    bytes random_bytes = 19;
```

services/transaction_record.proto
```diff
@@ -133,4 +133,10 @@ message TransactionRecord {
      * List of accounts with the corresponding staking rewards paid as a result of a transaction.
      */
     repeated AccountAmount paid_staking_rewards = 18;
+
+    /**
+     * An array of 384-bit pseudorandom bytes in the record of RandomGenerate transaction.
```
steven-sheehy, line 138:
```suggestion
     * An array of 384-bit pseudorandom data in the record of RandomGenerate transaction.
```

services/crypto_service.proto
```diff
@@ -62,6 +62,11 @@ service CryptoService {
      */
     rpc deleteAllowances (Transaction) returns (TransactionResponse);
 
+    /**
+     * Generates a pseudorandom number.
+     */
+    rpc randomGenerate (Transaction) returns (TransactionResponse);
```
steven-sheehy, line 68: Why is this being associated with the crypto service? That decision seems a bit random (sorry for the pun).

Perhaps this should be a network_service or possible a new type of util service?

services/random_generate.proto
```diff
@@ -0,0 +1,37 @@
...
+option java_package = "com.hederahashgraph.api.proto.java";
+option java_multiple_files = true;
+
+/**
+ * Generates a pseudorandom number
+ */
+message RandomGenerateTransactionBody {
+  /**
+   * If provided and is positive, returns a 32-bit pseudorandom number from the given range in the transaction record.
+   * If not set, will return a 256-bit pseudorandom number in the record.
```
steven-sheehy, line 34: What happens if it's negative? Do we need an error code for that or any other error scenarios?

steven-sheehy (changes requested):

services/random_generate.proto
Neeharika-Sompalli: I thought, if we can return the regular 384-bit array if the value is not positive. But it seems better to return a response code instead. Added `INVALID_RANDOM_GENERATE_RANGE`

services/crypto_service.proto
Neeharika-Sompalli: Added `UtilService`

tinker-michaelj (dismissed):
LGTM

steven-sheehy (approved):
LGTM

tinker-michaelj (approved):
LGTM!

PR #211, opened 2022-06-22
Traceability extension protobufs

mustafauzunn:
**Description**:

Traceability extension protobufs

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/streams/contract_action.proto
+++ b/streams/contract_action.proto
@@ -1,0 +1,160 @@
+syntax = "proto3";
+
+package proto;
+
+/*-
+ * ‌
+ * Hedera Network Services Protobuf
+ * ​
+ * Copyright (C) 2018 - 2022 Hedera Hashgraph, LLC
+ * ​
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ‍
+ */
+
+option java_package = "com.hedera.services.stream.proto";
+option java_multiple_files = true;
+
+import "basic_types.proto";
+
+
+message ContractActions {
+  repeated ContractAction contract_actions = 1;
+}
+
+/**
+ * The type of action described by the action proto.
+ */
+enum ContractActionType {
+  /**
+   * default non-value.
+   */
+  NO_ACTION = 0;
+
+  /**
+   * Most CALL, CALLCODE, DELEGATECALL, and STATICCALL, and first action of ContractCall/ContractCallLocal to deployed
+   * contracts. This does not include calls to system or precompiled contracts.
+   */
+  CALL = 1;
+
+  /**
+   * CREATE, CREATE2, and first action of ContractCreate.
+   */
+
+  CREATE = 2;
+
+  /**
+   * like Call, but to precompiled contracts (0x1 to 0x9 as of Berlin)
+   */
+  PRECOMPILE = 3;
+
+  /**
+   * Call, but to system contract like HTS or ERC20 facades over Token accounts
+   */
+  SYSTEM = 4;
+}
+
+/**
+ * A finer grained action with a function result. Sometimes called "internal transactions." The function call itself
+ * will be the first action in a list, followed by sub-action in the order they were executed.
+ */
+message ContractAction {
+
+  /**
+   * The type of this action.
+   */
+  ContractActionType call_type = 1;
+
+  /**
+   * Only the first action can come from an account, the rest will come from contracts.  Because of DELEGATECALL
+   * and CALLCODE the caller of actions whose parent is an account may also be an account.
+   */
+  oneof caller {
+    /**
+     * If the caller was a regular account, the AccountID.
+     */
+    AccountID calling_account = 2;
+
+    /**
+     * If the caller was a smart contract account, the ContractID.
+     */
+    ContractID calling_contract = 3;
+  }
+
+  /**
+   * The upper limit of gas this action can spend.
+   */
+  int64 gas = 4;
+
+  /**
+   * Bytes passed in as input data to this action.
+   */
+  bytes input = 5;
+
+  /**
+   * Who this action is directed to.
+   */
+  oneof recipient {
+    /**
+     * The AccountID of the recipient if the recipient is an account. Only HBars will be transferred, no other side
+     * effects should be expected.
+     */
+    AccountID recipient_account = 6;
+
+    /**
+     * The ContractID of the recipient if the recipient is a smart contract.
+     */
+    ContractID recipient_contract = 7;
+
+    /**
+     * If the action was directed at an invalid solidity address, what that address was.
+     */
+    bytes invalid_solidity_address = 8;
+  }
+
+  /**
+   * The value (in tinybars) that is associated with this action.
+   */
+  int64 value = 9;
+
+  /**
+   * The actual gas spent by this action.
+   */
+  int64 gas_used = 10;
+
+  /**
+   * The result data of the action.
+   */
+  oneof result_data {
+
+    /**
+     * If successful, the output bytes of the action.
+     */
+    bytes output = 11;
+
+    /**
+     * The contract itself caused the transaction to fail via the `REVERT` operation
+     */
+    bytes revert_reason = 12;
+
+    /**
+     * The transaction itself failed without an explicit `REVERT`
+     */
+    bytes error = 13;
+  }
+
+  /**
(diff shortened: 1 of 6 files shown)


streams/sidecar_file.proto
```diff
@@ -53,15 +55,43 @@ message TransactionSidecarRecord {
...
+
   /*
   * List of sidecar types.
   * In future there will be other categories.
   */
   oneof sidecar_records {
-    ContractStateChanges state_changes = 2;
+    ContractStateChanges state_changes = 3;
+    ContractActions actions = 4;
+    ContractBytecode bytecode = 5;
```
dimitar-dinev, line 67: Q: Why do we not have a `ContractBytecodes` wrapper of `repeated ContractBytecode` as we have for `ContractStateChange` and `ContractAction`? How will we externalize all contract bytecodes when a `ContractCall/Create` creates > 1 contracts?

streams/sidecar_file.proto
```diff
@@ -53,15 +55,43 @@ message TransactionSidecarRecord {
...
+
+  /**
+  * Sidecar generated from migration.
+  */
+  MIGRATION = 1;
+
+  /**
+  * Sidecar generated from trace.
+  */
+  TRACE = 2;
```
steven-sheehy, line 88: I imagine in the future we might put other, non-traceability information in the sidecar like errata records so we should name this something more general. Or we can just change it to a `bool migration` which might be clearer.

Whichever data type, we should choose the default value to be the non-migration value so it doesn't incur the cost of writing extra bytes indefinitely to the file for the one-time migration. If we use `bool migration` then the default value is false and would not serialize the field for the majority of record files.

steven-sheehy (changes requested):

streams/contract_action.proto
```diff
@@ -0,0 +1,155 @@
...
+    AccountID calling_account = 2;
+
+    /**
+     * If the caller was a smart contract account, the ContractID.
+     */
+    ContractID calling_contract = 3;
+  }
+
+  /**
+   * Gas given to this action, the upper limit of gas this action can spend.
```
Nana-EC, line 90: nit: current wording makes it unclear which gas it refers to
```suggestion
   * The upper limit of gas this action can spend.
```

streams/contract_state_change.proto
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * The storage changes to a smart contract's storage as a side effect of the function call.
+ */
+message ContractStateChange {
+
+  /**
+   * The contract to which the storage changes apply to
+   */
+  ContractID contractID = 1;
```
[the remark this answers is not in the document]
mustafauzunn: DONE

streams/record_stream_file.proto
```diff
@@ -103,4 +103,6 @@ message SidecarMetadata {
 enum SidecarType {
   SIDECAR_TYPE_UNKNOWN = 0;
```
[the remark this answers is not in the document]
mustafauzunn: DONE

streams/contract_bytecode.proto
```diff
@@ -0,0 +1,39 @@
+syntax = "proto3";
+
+package proto;
+
+/*-
+ * ‌
+ * Hedera Network Services Protobuf
+ * ​
+ * Copyright (C) 2018 - 2021 Hedera Hashgraph, LLC
```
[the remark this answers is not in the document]
mustafauzunn: DONE

streams/contract_action.proto
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+   * The actual gas spent by this action.
+   */
+  int64 gas_used = 10;
+
+  /**
+   * The result data of the action.
+   */
+  oneof resultData {
```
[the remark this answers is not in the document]
mustafauzunn: DONE

streams/sidecar_file.proto
```diff
@@ -53,15 +55,43 @@ message TransactionSidecarRecord {
...
+}
+
 message ContractStateChanges {
   repeated ContractStateChange contract_state_changes = 1;
 }
+
+message ContractActions {
+  repeated ContractAction contract_actions = 1;
+}
```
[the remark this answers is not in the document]
mustafauzunn: DONE

streams/sidecar_file.proto
mustafauzunn: DONE

streams/contract_action.proto
mustafauzunn: DONE

steven-sheehy (dismissed):
LGTM

streams/contract_state_change.proto
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ‍
+ */
+
+option java_package = "com.hederahashgraph.api.proto.java";
```
xin-hedera, line 25: same here

xin-hedera (changes requested):

streams/contract_action.proto
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * The type of action described by the action proto.
+ */
+enum ContractActionType {
+  /**
+   * Most CALL, CALLCODE, DELEGATECALL, and STATICCALL, and first action of ContractCall/ContractCallLocal to deployed
+   * contracts. This does not include calls to system or precompiled contracts.
+   */
+  CALL = 0;
```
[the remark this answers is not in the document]
steven-sheehy: What old protobuf? This is a new message for sidecar and it's a required field set by consensus nodes so there shouldn't be a need for an unknown value. I had him just remove it in the last commit.

streams/contract_bytecode.proto
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ‍
+ */
+
+option java_package = "com.hederahashgraph.api.proto.java";
```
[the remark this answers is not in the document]
steven-sheehy: Good catch. All files in `streams` should have a separate java package from services folder.

streams/contract_action.proto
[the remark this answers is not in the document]
xin-hedera: never mind. I thought if we add new fields to the enum in the future, the parser generated from the current version would parse new enum fields to `CALL`, it'll actually parse it to `ContractActionType.UNRECOGNIZED(-1)`.

streams/contract_bytecode.proto
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+  * Contract bytecode during deployment
+  */
+  bytes initcode = 1;
+
+  /**
+  * Contract bytecode after deployment
+  */
+  bytes runtime_bytecode = 2;
```
steven-sheehy, line 38: I think we would also need the `ContractID contract_id` here as well for migrations. For non-migration we can use the associated consensus timestamp to link it to the contract create. 

For migrations, the consensus timestamp is for a random transaction after startup and not associated with the contract create. So we would need the contract ID here or we would need the consensus timestamp to be linked to some contract create synthetic transaction.

streams/contract_bytecode.proto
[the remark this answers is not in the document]
mustafauzunn: DONE

streams/contract_state_change.proto
mustafauzunn: DONE

xin-hedera (approved):
LGTM

steven-sheehy (dismissed):
LGTM

streams/sidecar_file.proto
stoqnkpL: Yes, each contract deployment creates a separate record, e.g.
- deploying 5 contracts with a single `ContractCreate` grpc transaction would create the top-level creation record and 4 child creation records; each one will have a corresponding `ContractBytecode` sidecar record.
- deploying 5 contracts via a `ContractCall` grpc transaction would create the top-level call record and 5 creation records; each child will have a corresponding `ContractBytecode` sidecar record.

stoqnkpL (approved):
LGTM

stoqnkpL (changes requested):
I think we should resolve this [comment](https://github.com/hashgraph/hedera-improvement-proposal/pull/513#discussion_r905277554) before merging.

streams/sidecar_file.proto
dimitar-dinev: Got it, thanks for the clarifications!

steven-sheehy (dismissed):

xin-hedera (approved):

tinker-michaelj (dismissed):
LGTM

steven-sheehy:
Need to gpg sign all your commits

xin-hedera (approved):

PR #214, opened 2022-06-27
Add more staking metadata to `NodeStakeUpdate` message

tinker-michaelj:
**Description**:
Add fields to export the following staking Services properties:
```
### Dynamic properties, can be changed via FileUpdate to 0.0.121
staking.fees.nodeRewardPercentage=0                                              
staking.fees.stakingRewardPercentage=100                                         
staking.maxDailyStakeRewardThPerH=17_808                                                                                           
staking.rewardRate=0                                                         
staking.startThreshold=250_000_000_00_000_000
### Static properties, can only change on network restart
staking.periodMins=1440
staking.rewardHistory.numStoredPeriods=365
```

### DIFF
--- a/services/node_stake_update.proto
+++ b/services/node_stake_update.proto
@@ -40,47 +40,47 @@ message NodeStakeUpdateTransactionBody {
   /**
    * The maximum reward rate, in tinybars per whole hbar, that any account can receive in a day.
    */
-  int64 max_daily_reward_rate = 2;
+  int64 max_daily_reward_rate = 3;
 
   /**
    * The fraction of the network and service fees (as a whole number in the range [0, 100]) paid to 
    * the node reward account 0.0.801.
    */
-  int32 node_reward_fees_percent = 3;
+  int32 node_reward_fees_percent = 4;
 
   /**
    * Staking info of each node at the beginning of the new staking period
    */
-  repeated NodeStake node_stake = 4;
+  repeated NodeStake node_stake = 2;
 
   /**
    * The maximum number of trailing periods for which a user can collect rewards. For example, if this
    * is 365 with a UTC calendar day period, then users must collect rewards at least once per calendar
    * year to avoid missing any value.
    */
-  int64 num_rewardable_trailing_periods = 5;
+  int64 num_rewardable_trailing_periods = 6;
 
   /**
    * The number of minutes in a staking period. Note for the special case of 1440 minutes, periods are 
    * treated as UTC calendar days, rather than repeating 1440 minute periods left-aligned at the epoch.
    */
-  int64 period_mins = 6;
+  int64 period_mins = 7;
 
   /**
    * The fraction of the network and service fees (as a whole number in the range [0, 100]) paid to 
    * the staking reward account 0.0.800.
    */
-  int32 staking_reward_fees_percent = 7;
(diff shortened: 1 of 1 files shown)


services/node_stake_update.proto
```diff
@@ -37,10 +37,50 @@ message NodeStakeUpdateTransactionBody {
...
+   * The fraction of the network and service fees (as a whole number in the range [0, 100]) paid to 
+   * the node reward account 0.0.801.
+   */
+  int32 node_reward_fees_percent = 3;
+
   /**
    * Staking info of each node at the beginning of the new staking period
    */
-  repeated NodeStake node_stake = 2;
+  repeated NodeStake node_stake = 4;
```
xin-hedera, line 54: believe we can't have breaking changes this time, so existing fields number should stay the same:

```suggestion
  repeated NodeStake node_stake = 2;
```

services/node_stake_update.proto
```diff
@@ -37,10 +37,50 @@ message NodeStakeUpdateTransactionBody {
...
+
+  /**
+   * The minimum balance of staking reward account 0.0.800 required to active rewards.
+   */
+  int64 start_threshold = 8;
+
+  /**
+   * The total number of tinybars to be distributed as staking rewards each period.
+   */
+  int64 total_rewards_per_period = 9;
```
xin-hedera, line 83: Q: is this just the sum of `node_stake.reward_rate`? Or `rewardRate`, a.k.a, 

`StakingRewardRate - the total tinybars earned as staking reward during each staking period (example: 100_000_000L * 1_000_000_000 / 365 to emit rewards at a total rate of one billion hbars per year)
` in the HIP?

tinker-michaelj: It's the `StakingRewardRate` from the HIP. 👍

services/node_stake_update.proto
```diff
@@ -37,10 +37,50 @@ message NodeStakeUpdateTransactionBody {
    */
   Timestamp end_of_staking_period = 1;
 
+  /**
+   * The maximum reward rate, in tinybars per whole hbar, that any account can receive in a day.
+   */
+  int64 max_daily_reward_rate = 3;
```
steven-sheehy, line 43: Will there be a HIP update to add this property? I don't think it's there currently.

steven-sheehy (changes requested):
Requests to align naming with HIP defined names for consistency. I also think the services properties should be named to be consistent with the HIP to avoid confusion.

tinker-michaelj:
> Requests to align naming with HIP defined names for consistency. I also think the services properties should be named to be consistent with the HIP to avoid confusion.

Aligned names; will change Services property names in 0.28. 👍

tinker-michaelj:
> missing field number 5 in `NodeStakeUpdateTransactionBody`. should order fields by its number

Fixed and re-ordered!

xin-hedera (approved):

Neeharika-Sompalli (approved):
LGTM ! But, will need to discuss better approach for notifying mirror node on property changes.

services/node_stake_update.proto
tinker-michaelj: That could be done. 👀

steven-sheehy (approved):

PR #216, opened 2022-07-01
Update transaction name for HIP-351

Neeharika-Sompalli:
Update name of `RandomGenerate` transaction to `PrngGenerate` transaction as per latest discussion.

### DIFF
--- a/services/prng_generate.proto
+++ b/services/prng_generate.proto
@@ -28,7 +28,7 @@ option java_multiple_files = true;
 /**
  * Generates a pseudorandom number
  */
-message RandomGenerateTransactionBody {
+message PrngGenerateTransactionBody {
   /**
    * If provided and is positive, returns a 32-bit pseudorandom number from the given range in the transaction record.
    * If not set or set to zero, will return a 384-bit pseudorandom number in the record.
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1143,7 +1143,7 @@ enum HederaFunctionality {
     /**
      * Generates a pseudorandom number.
      */
-    PrngGenerate = 86;
+    PRNG = 86;
 }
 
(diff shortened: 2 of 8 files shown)


tinker-michaelj (dismissed):
LGTM!

services/prng_generate.proto
```diff
@@ -28,7 +28,7 @@ option java_multiple_files = true;
 /**
  * Generates a pseudorandom number
  */
-message RandomGenerateTransactionBody {
+message PrngGenerateTransactionBody {
```
steven-sheehy, line 31: Isn't it redundant to say "Pseudorandom Number Generator Generate Transaction"?

Neeharika-Sompalli: True, in a recent call there was a consensus on this name since SDK wanted to use this name.
So it could be `PRNGenerate` ?

Neeharika-Sompalli: True, in a recent call there was a consensus on this name since SDK wanted to use this name.
So it could be `PRNGenerate` ?

steven-sheehy: Would've liked to have been on that call. I would think it would be `PseudorandomNumberGeneratorTransaction` or `PrngTransaction`.

rbair23: I understand it to be `PrngTransaction`

Neeharika-Sompalli: Yes, updating it to `PrngTransaction` and HederaFunctionality to `PRNG`

steven-sheehy (approved):
LGTM

xin-hedera (approved):
LGTM

keirabee (approved):
These changes look alright to me as long as everyone wants these names for these:)

PR #218, opened 2022-07-21
Hip-351 : Change name of `prng` txn to `utilPrng`

Neeharika-Sompalli:
Related to HIP-351 .
Change name of `prng` transaction to `utilPrng` as per latest discussion, to maintain naming standards for the transaction and transaction body.

### DIFF
--- a/services/util_service.proto
+++ b/services/util_service.proto
@@ -34,5 +34,5 @@ service UtilService {
   /**
    * Generates a pseudorandom number.
    */
-  rpc utilPrng (Transaction) returns (TransactionResponse);
+  rpc prng (Transaction) returns (TransactionResponse);
 }
--- a/services/util_service.proto
+++ b/services/util_service.proto
@@ -34,5 +34,5 @@ service UtilService {
   /**
    * Generates a pseudorandom number.
    */
-  rpc prng (Transaction) returns (TransactionResponse);
+  rpc utilPrng (Transaction) returns (TransactionResponse);
 }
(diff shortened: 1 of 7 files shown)


tinker-michaelj (dismissed):
LGTM!

services/util_service.proto
```diff
@@ -34,5 +34,5 @@ service UtilService {
   /**
    * Generates a pseudorandom number.
    */
-  rpc prng (Transaction) returns (TransactionResponse);
+  rpc utilPrng (Transaction) returns (TransactionResponse);
```
steven-sheehy, line 37: This is actually not consistent with other services. Since we're already in the `UtilService` we don't need to specify `util` again. Instead it should just be the `(action)` in `(service name)(action)(TransactionBody)` like we do in crypto_service, etc. So should revert this change

steven-sheehy (changes requested):

services/util_service.proto
Neeharika-Sompalli: @steven-sheehy  `TokenService` has all transaction names ending with `Token` like `mintToken`, instead of `mint`. Similarly in other services as well.

steven-sheehy: No service method has the service name first. Some of them have the Hedera entity it's acting upon after the action. This makes sense since the `TokenService.mintToken()` is minting a token, the `CryptoService.createAccount()` is creating an account, etc. The `UtilService.utilPrng()` is not prng'ing a util or util'ing a prng.

Neeharika-Sompalli: 😄 Fixed. Thanks !

steven-sheehy (approved):
LGTM

lbaird (approved):
LGTM!

PR #219, opened 2022-08-15
Deprecate token relationships fields

tinker-michaelj:
**Description**:
 - Mark several fields `deprecated` that will be removed in Nov release (c.f. [here](https://www.notion.so/swirldslabs/HIP-367-Token-Information-Deprecation-15aff829f32f488b90cbc3732b17d305)).

### DIFF
--- a/services/contract_get_info.proto
+++ b/services/contract_get_info.proto
@@ -122,7 +122,7 @@ message ContractGetInfoResponse {
         /**
          * The tokens associated to the contract
          */
-        repeated TokenRelationship tokenRelationships = 11;
+        repeated TokenRelationship tokenRelationships = 11 [deprecated = true];
 
         /**
          * The ledger ID the response was returned from; please see <a href="https://github.com/hashgraph/hedera-improvement-proposal/blob/master/HIP/hip-198.md">HIP-198</a> for the network-specific IDs. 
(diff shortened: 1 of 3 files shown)


services/crypto_get_info.proto
```diff
@@ -141,7 +141,7 @@ message CryptoGetInfoResponse {
         /**
          * All tokens related to this account
          */
-        repeated TokenRelationship tokenRelationships = 15;
+        repeated TokenRelationship tokenRelationships = 15 [deprecated = true];
```
Neeharika-Sompalli, line 144: This field is also used in `get_account_details.proto`. Can be deprecated there as well

Neeharika-Sompalli (dismissed):
LGTM !

services/contract_get_info.proto
```diff
@@ -122,7 +122,7 @@ message ContractGetInfoResponse {
         /**
          * The tokens associated to the contract
          */
-        repeated TokenRelationship tokenRelationships = 11;
+        repeated TokenRelationship tokenRelationships = 11 [deprecated = true];
```
steven-sheehy, line 125: For each deprecation, should we add to the field comment about why it's deprecated and what is the alternate source of information? Or at least link to the HIP?

tinker-michaelj: Added some justification + link to HIP.

Neeharika-Sompalli (approved):

steven-sheehy (approved):

PR #223, opened 2022-08-23
Add alias bytes to crypto create

### LINKED ISSUE #222: Expand alias set support for CryptoCreate
### Problem

Currently the alias can only be set via the CryptoTransfer transaction.

### Solution

Expand alias set support for CryptoCreate protobuf to allow account creates with alias

### Alternatives

_No response_

Nana-EC:
Signed-off-by: Nana Essilfie-Conduah <nana@swirldslabs.com>

**Description**:
To support the increased support of the alias in the CryptoCreate Transaction

- Add alias bytes to crypto create
- Update `AccountID.alias` description
- Update `TransactionRecord.alias` description

**Related issue(s)**:

Fixes #222 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -95,7 +95,8 @@ message AccountID {
          * a protobuf Key message for any primitive key type. Currently only primitive key bytes are supported as an alias
          * (ThresholdKey, KeyList, ContractID, and delegatable_contract_id are not supported)
          *
-         * May also be the ethereum account 20-byte EVM address to be used initially in place of the public key bytes.
+         * May also be the ethereum account 20-byte EVM address to be used initially in place of the public key bytes. This EVM
+         * address may be either the encoded form of the shard.realm.num or the keccak-256 hash of a ECDSA_SECP256K1 primitive key.
          *
          * At most one account can ever have a given alias and it is used for account creation if it
          * was automatically created using a crypto transfer. It will be null if an account is created normally.
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -104,8 +104,8 @@ message AccountID {
          * If a transaction auto-creates the account, any further transfers to that alias will simply be deposited
          * in that account, without creating anything, and with no creation fee being charged.
          *
-         * If a transaction lazy-create this account, a susequent transaction will be required containing the public key bytes 
-         * that map to the ethereum account address bytes. The provided public key bytes will then serve as the final alias bytes.
+         * If a transaction lazily-creates this account, a subsequent transaction will be required containing the public key bytes 
+         * that map to the EVM address bytes. The provided public key bytes will then serve as the final alias bytes.
          */
         bytes alias = 4;
 
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -95,12 +95,17 @@ message AccountID {
          * a protobuf Key message for any primitive key type. Currently only primitive key bytes are supported as an alias
          * (ThresholdKey, KeyList, ContractID, and delegatable_contract_id are not supported)
          *
+         * May also be the ethereum account 20-byte EVM address to be used initially in place of the public key bytes.
+         *
          * At most one account can ever have a given alias and it is used for account creation if it
          * was automatically created using a crypto transfer. It will be null if an account is created normally.
          * It is immutable once it is set for an account.
          *
          * If a transaction auto-creates the account, any further transfers to that alias will simply be deposited
          * in that account, without creating anything, and with no creation fee being charged.
+         *
+         * If a transaction lazy-create this account, a susequent transaction will be required containing the public key bytes 
+         * that map to the ethereum account address bytes. The provided public key bytes will then serve as the final alias bytes.
          */
         bytes alias = 4;
 
--- a/services/crypto_create.proto
+++ b/services/crypto_create.proto
@@ -160,7 +160,8 @@ message CryptoCreateTransactionBody {
      * supported as the key for an account with an alias. ThresholdKey, KeyList, ContractID, and
      * delegatable_contract_id are not supported.
      *
-     * At most only one account can ever have a given alias on the network.
+     * A given alias can map to at most one account on the network at a time. This uniqueness will be enforced
+     * relative to aliases currently on the network at alias assignment.
      *
      * If a transaction creates an account using an alias, any further crypto transfers to that alias will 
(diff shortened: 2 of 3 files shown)


shemnon (approved):

services/crypto_create.proto
```diff
@@ -151,4 +151,18 @@ message CryptoCreateTransactionBody {
      * If true, the account declines receiving a staking reward. The default value is false.
      */
     bool decline_reward = 17;
+
+    /**
+     * The bytes to be used as the account's alias. The bytes will be 1 of 2 options. 
+     * It will be the serialization of a protobuf Key message for an ED25519/ECDSA primitive key type.
```
steven-sheehy, line 157: There are two ECDSA key types in the `Key` proto. Should adjust the other references too.
```suggestion
     * It will be the serialization of a protobuf Key message for an ED25519/ECDSA_SECP256K1 primitive key type. 
```

services/basic_types.proto
```diff
@@ -95,12 +95,17 @@ message AccountID {
...
+         * May also be the ethereum account 20-byte EVM address to be used initially in place of the public key bytes.
+         *
          * At most one account can ever have a given alias and it is used for account creation if it
          * was automatically created using a crypto transfer. It will be null if an account is created normally.
          * It is immutable once it is set for an account.
          *
          * If a transaction auto-creates the account, any further transfers to that alias will simply be deposited
          * in that account, without creating anything, and with no creation fee being charged.
+         *
+         * If a transaction lazy-create this account, a susequent transaction will be required containing the public key bytes
```
steven-sheehy, line 107:
```suggestion
         * If a transaction lazily creates this account, a subsequent transaction will be required containing the public key bytes 
```

services/basic_types.proto
```diff
@@ -95,12 +95,17 @@ message AccountID {
...
+         *
          * At most one account can ever have a given alias and it is used for account creation if it
          * was automatically created using a crypto transfer. It will be null if an account is created normally.
          * It is immutable once it is set for an account.
          *
          * If a transaction auto-creates the account, any further transfers to that alias will simply be deposited
          * in that account, without creating anything, and with no creation fee being charged.
+         *
+         * If a transaction lazy-create this account, a susequent transaction will be required containing the public key bytes 
+         * that map to the ethereum account address bytes. The provided public key bytes will then serve as the final alias bytes.
```
steven-sheehy, line 108:
```suggestion
         * that map to the EVM address bytes. The provided public key bytes will then serve as the final alias bytes.
```

services/crypto_create.proto
```diff
@@ -151,4 +151,18 @@ message CryptoCreateTransactionBody {
      * If true, the account declines receiving a staking reward. The default value is false.
      */
     bool decline_reward = 17;
+
+    /**
+     * The bytes to be used as the account's alias. The bytes will be 1 of 2 options. 
+     * It will be the serialization of a protobuf Key message for an ED25519/ECDSA primitive key type. 
+     * Or the if the account is ECDSA based it may also be the public address, calcluated as the last 20 bytes of the keccak-256 of the ECDSA primitive key.
```
steven-sheehy, line 158:
```suggestion
     * Or the if the account is ECDSA based it may also be the public address, calcluated as the last 20 bytes of the keccak-256 hash of the ECDSA primitive key.
```

steven-sheehy, line 158: This line looks a bit longer than the margin should allow. Also, it's hard to read this paragraph since every line starts on a new line instead of flowing together as a paragraph. Makes it hard to tell if they were supposed to be separate paragraphs or not.

services/basic_types.proto
```diff
@@ -95,12 +95,17 @@ message AccountID {
          * a protobuf Key message for any primitive key type. Currently only primitive key bytes are supported as an alias
          * (ThresholdKey, KeyList, ContractID, and delegatable_contract_id are not supported)
          *
+         * May also be the ethereum account 20-byte EVM address to be used initially in place of the public key bytes.
```
steven-sheehy, line 98: Currently, this EVM address can only be the `shard.realm.num` format encoded as 20 bytes and not the hash form. If that is remaining true, then we should explicitly state that it only supports the zero-prefixed `shard-realm.num` format. If that is changing, then we should explicitly state both formats are supported now.

steven-sheehy (changes requested):

services/crypto_create.proto
Nana-EC: Fixed

services/basic_types.proto
Nana-EC: Fixed

services/basic_types.proto
Nana-EC: Fixed

services/crypto_create.proto
Nana-EC: Fixed

Nana-EC: Fixed

services/basic_types.proto
Nana-EC: Updated

steven-sheehy (approved):
LGTM

services/crypto_create.proto
```diff
@@ -151,4 +151,19 @@ message CryptoCreateTransactionBody {
...
+
+    /**
+     * The bytes to be used as the account's alias. The bytes will be 1 of 2 options. It will be the 
+     * serialization of a protobuf Key message for an ED25519/ECDSA_SECP256K1 primitive key type. If the 
+     * account is ECDSA_SECP256K1 based it may also be the public address, calcluated as the last 20 bytes of
+     * the keccak-256 hash of the ECDSA_SECP256K1 primitive key. Currently only primitive key bytes are 
+     * supported as the key for an account with an alias. ThresholdKey, KeyList, ContractID, and
+     * delegatable_contract_id are not supported.
+     *
+     * At most only one account can ever have a given alias on the network.
```
tinker-michaelj, line 163: The "can _ever_ have" may be confusing; we will only enforce uniqueness relative to aliases _currently_ on the network.

tinker-michaelj (dismissed):
LGTM!

tinker-michaelj (approved):
LGTM

Neeharika-Sompalli (approved):
LGT<M!

services/crypto_create.proto
Nana-EC: Fixed

steven-sheehy (approved):

PR #232, opened 2022-09-27
03916 remove uints

georg-getz:
Signed-off-by: Georgi Georgiev <georgi.getz@outlook.com>

**Description**:
This PR is the first of several to be parallelly merged fixing the misalignment of types between the protobuf files and the contracts and it also includes replacement of all uints in the protobuf files. The reason the uints are a problem is because they are ultimately being converted to a signed type regardless and the upper half of their values is unusable. 
image
Part of the work in the other repos depends on feedback given to this PR.

**Related issue(s)**:

Fixes hashgraph/hedera-services#3921
hashgraph/hedera-services#3916

**Notes for reviewer**:
This PR won't be merged until the necessary follow-up PRs in the contracts/docs/services/sdk repos have all been opened and approved as well.

**Checklist**

### DIFF
--- a/mirror/consensus_service.proto
+++ b/mirror/consensus_service.proto
@@ -55,7 +55,7 @@ message ConsensusTopicQuery {
      * The maximum number of messages to receive before stopping. If not set or set to zero it will
      * return messages indefinitely.
      */
-    uint64 limit = 4;
+    int64 limit = 4;
 }
 
 message ConsensusTopicResponse {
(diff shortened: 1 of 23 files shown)


rbair23:
I have reservations. By specifying uint, we are telling the caller (through this schema which is our API definition) that we do not allow negative numbers. Of course we also say we allow all uint64 bits, but actually we only use uint63 bits. So which is worse, using uint64 and documenting we only use 2^63 as valid input, or int64 and documenting that we don't support negative numbers?

Issue #241, opened 2022-11-10
Add new fields for externalizing account alias


stoqnkpL:
Creating accounts with EC keys can result in an EVM public address alias being tracked by the consensus nodes. This alias should be externalized so that mirror nodes have that information.

PR #242, opened 2022-11-10
Add new fields for externalizing account alias

MiroslavGatsanoga:
Signed-off-by: Miroslav Gatsanoga <miroslav.gatsanoga@limechain.tech>

**Description**:

Creating accounts with EC keys can result in an EVM public address alias being tracked by the consensus nodes. This alias should be externalized so that mirror nodes have that information.

**Related issue(s)**:

- Fixes: https://github.com/hashgraph/hedera-protobufs/issues/241
- https://github.com/hashgraph/hedera-services/issues/4146

### DIFF
--- a/services/crypto_update.proto
+++ b/services/crypto_update.proto
@@ -29,6 +29,7 @@ import "basic_types.proto";
 import "duration.proto";
 import "timestamp.proto";
 import "google/protobuf/wrappers.proto";
+import "transaction_record.proto";
 
 /**
  * Change properties for the given account. Any null field is ignored (left unchanged). This
@@ -160,4 +161,6 @@ message CryptoUpdateTransactionBody {
      * account. Otherwise it updates the account's auto-renew account to the referenced account.
      */
     AccountID auto_renew_account = 19;
+
+    VirtualAddressUpdate virtualAddressUpdate = 20;
 }
--- a/services/transaction_record.proto
+++ b/services/transaction_record.proto
@@ -145,4 +145,28 @@ message TransactionRecord {
          */
         int32 prng_number = 20;
     }
+
+   /**
+    * Virtual address updated info. At most one update can be performed at a time.
+    */
+    VirtualAddressUpdate virtualAddressUpdate = 21;
+}
+
+message VirtualAddressUpdate {
+    /**
+     * A 20-byte EVM address account identifier. This is the keccak-256 hash of a ECDSA_SECP256K1 primitive key.
+     *
+     * At most one account can ever have a given evm address
+     */
+    bytes evmAddress = 1;
+
+    /**
+     * Flag if this address is being added
+     */
+    bool addition = 2;
+
+    /**
+     * Flag if this address should now be set or is the default address on the account
+     */
+    bool isDefaultAddress = 3;
 }
--- a/services/crypto_get_info.proto
+++ b/services/crypto_get_info.proto
@@ -188,10 +188,26 @@ message CryptoGetInfoResponse {
          * cover auto-renewal fees.
          */
         AccountID auto_renew_account = 23;
+
+        repeated VirtualAddress virtualAddresses = 24;
+
+        bytes defaultVirtualAddress = 25;
     }
 
     /**
      * Info about the account (a state proof can be generated for this)
      */
     AccountInfo accountInfo = 2;
+}
+
+message VirtualAddress {
+    /**
+     * The 20-byte EVM address
+     */
+    bytes virtualAddress = 1;
+
+    /**
+     * The ethereum transaction nonce associated with this address.
+     */
+    int64 nonce = 2;
 }


services/transaction_record.proto
```diff
@@ -145,4 +146,21 @@ message TransactionRecord {
...
+message VirtualAddressUpdate {
+    /**
+     * An accounts 20-byte EVM address account identifier and nonce.
+     */
+    VirtualAddress virtualAddress = 1;
+
+    /**
+     * Flag if this address is being removed. If false virtual address is being added
+     */
+    bool removed = 2;
```
Nana-EC, line 165: Also we should make this nullable  since on virtualAddressUpdate.virtualAddress.isDefaultAddress updates this will be null.
```suggestion
    BoolValue remove = 2;
```

Nana-EC (changes requested):
Initial thoughts reflecting so recent updates

services/basic_types.proto
```diff
@@ -110,6 +110,7 @@ message AccountID {
          */
         bytes alias = 4;
 
+        bytes evmAddress = 5;
```
tinker-michaelj, line 113: We've been using "snake-case" now---so `evm_address`, `is_default_address`, etc.

services/transaction_record.proto
```diff
@@ -31,6 +31,7 @@ import "custom_fees.proto";
 import "transaction_receipt.proto";
 import "crypto_transfer.proto";
 import "contract_call_local.proto";
+import "crypto_get_info.proto";
```
tinker-michaelj, line 34: Unneeded import.

services/crypto_update.proto
```diff
@@ -29,6 +29,7 @@ import "basic_types.proto";
 import "duration.proto";
 import "timestamp.proto";
 import "google/protobuf/wrappers.proto";
+import "transaction_record.proto";
```
tinker-michaelj, line 32: ⚠️ Unneeded import, and causes circular import:
```
[ERROR] /Users/michaeltinker/Dev/hedera-protobufs-java/src/main/proto/freeze_service.proto [0:0]: query_header.proto:28:1: File recursively imports itself: query_header.proto -> transaction.proto -> transaction_body.proto -> crypto_update.proto -> transaction_record.proto -> contract_call_local.proto -> query_header.proto
```

tinker-michaelj (changes requested):
Just realized there are some unneeded imports causing circular imports.

services/transaction_record.proto
```diff
@@ -145,4 +145,9 @@ message TransactionRecord {
          */
         int32 prng_number = 20;
     }
+
+    /**
+     * Virtual address update info, if virtual address is specified in the CryptoUpdate transaction. At most one update can be performed at a time.
+     */
+    VirtualAddressUpdate virtual_address_update = 21;
```
dimitar-dinev, line 152: Won't `VirtualAddress` be sufficient here?'

Оnly `CryptoUpdate` can remove virtual addresses, and it has the intended `VirtualAddressUpdate` in the transaction body. Therefore, if the `CryptoUpdate` is successful, mirror node can get the needed virtual address update info from the transaction body, and we won't have to duplicate it in the record, correct?

And in all other transactions with virtual address info in the record, we won't be removing the address, but adding it/updating nonce, so the `remove` field of  `VirtualAddressUpdate` won't be needed -> we can directly use `VirtualAddress` here.

services/crypto_update.proto
dimitar-dinev: Removed

services/basic_types.proto
dimitar-dinev: Renamed

services/transaction_record.proto
Nana-EC: Good point.
I like the simplification

services/basic_types.proto
```diff
@@ -1592,3 +1593,32 @@ message StakingInfo {
...
+message VirtualAddressUpdate {
+    /**
+     * An accounts 20-byte EVM address account identifier and nonce.
+     */
+    VirtualAddress virtual_address = 1;
+
+    /**
+     * Flag if this address is being removed. If false, the virtual address is being added.
+     */
+    google.protobuf.BoolValue removed = 2;
```
xin-hedera, line 1623: what if `removed` is not set?

services/basic_types.proto
```diff
@@ -110,6 +110,7 @@ message AccountID {
          */
         bytes alias = 4;
```
steven-sheehy, line 112: There should be some docs around this including what it is, size, how it's generated, whether it supports shard.realm.num format or only keccak hash, etc. Also, will the `alias` field still allow 20 byte EVM addresses or reject?

services/basic_types.proto
```diff
@@ -1592,3 +1593,32 @@ message StakingInfo {
...
+
+    /**
+     * The ethereum transaction nonce associated with this address.
+     */
+    int64 nonce = 2;
+
+    /**
+     * Flag if this address should now be set or is the default address on the account.
+     */
+    bool is_default_address = 3;
```
steven-sheehy, line 1611: When used together with its container as in `VirtualAddress.isDefaultAddress` it seems redundant to include address twice.

```suggestion
    bool is_default = 3;
```

services/basic_types.proto
```diff
@@ -1592,3 +1593,32 @@ message StakingInfo {
...
+     */
+    bytes address = 1;
+
+    /**
+     * The ethereum transaction nonce associated with this address.
+     */
+    int64 nonce = 2;
+
+    /**
+     * Flag if this address should now be set or is the default address on the account.
```
steven-sheehy, line 1609: `VirtualAddress` shouldn't be for updates and only returned on queries. So set terminology doesn't make sense.

```suggestion
     * Flag if this address is the default address on the account.
```

services/basic_types.proto
```diff
@@ -1592,3 +1593,32 @@ message StakingInfo {
...
+     */
+    bool is_default_address = 3;
+}
+
+message VirtualAddressUpdate {
```
steven-sheehy, line 1614: What if we combined `VirtualAddress` and `VirtualAddressUpdate` like so?

```proto
enum VirtualAddressState {
    PRIMARY,
    SECONDARY,
    DELETED
}

message VirtualAddress {
    /**
     * The 20-byte EVM address.
     */
    bytes address = 1;

    /**
     * The ethereum transaction nonce associated with this address.
     */
    int64 nonce = 2;

    VirtualAddressState state = 3;
}
```

services/basic_types.proto
```diff
@@ -1592,3 +1593,32 @@ message StakingInfo {
...
+    /**
+     * An accounts 20-byte EVM address account identifier and nonce.
+     */
+    VirtualAddress virtual_address = 1;
```
steven-sheehy, line 1618: So to remove a VirtualAddress you need to provide a nonce and possibly a default flag? Bit weird even if those values might be ignored for removals. Another possible simplification is to have the field names make the action explicit:

```proto
VirtualAddress virtual_address_add = 1;
bytes virtual_address_remove = 2;
```

or

```proto
oneof virtual_address {
    VirtualAddress add = 1;
    bytes remove = 2;
}

**Review by steven-sheehy — CHANGES_REQUESTED** [review]

File: services/transaction_record.proto (hunk 5 above)
Comment (stoqnkpL) [reply services/transaction_record.proto:152]: The `VirtualAddressUpdate` field can be used in a couple of situations:
1. When we're initially creating an account. In a couple of use cases, we need to externalize the `evmAddress` that is calculated on the consensus node side, but a simple `bytes calculated_evm_address` field will be sufficient. The nonce value is always 0.
2. When we're updating an account and adding/removing virtual addresses from this account. In this case, the mirror nodes can extract this information from the transaction body and we don't need to write it in the records.

Conclusion: it looks like we can remove the `VirtualAddressUpdate` field and add `bytes calculated_evm_address` instead. @steven-sheehy @Nana-EC wdyt?

File: services/basic_types.proto (hunk 7 above)
Comment (stoqnkpL) [reply services/basic_types.proto:112]: So currently if an account has an EC key, we calculate an `EVM address` alias, for example:
Alice's account:
Hedera account id = 0.0.1932
Admin key = EC key
EVM address = `0x71C7656EC7ab88b098defB751B7401B5f6d8976F` -> currently only stored in the in-memory map on the consensus node

The EVM address can be used in a couple of cases, for example in `AccountAmounts` constructed for a crypto transfer.

So we would probably want to continue accepting EVM address values for the alias at least in these cases where it's currently allowed in order to be backward compatible?

Apart from that, I would suggest defining the fields like this:
- `alias` can only be either ECDSA or ED key alias
- `evm_address` can only be the calculated EVM address or Virtual address added to an account

File: services/basic_types.proto [hunk 12]
```diff
@@ -1592,3 +1593,32 @@ message StakingInfo {
...
+message VirtualAddress {
+    /**
+     * The 20-byte EVM address.
+     */
+    bytes address = 1;
+
+    /**
+     * The ethereum transaction nonce associated with this address.
+     */
+    int64 nonce = 2;
```
stoqnkpL, line 1606: I'm not convinced we need a `nonce` field at all. 
For a non-contract account's EVM address, the nonce can only be updated by one for each `EthereumTransaction` initiated by this EVM address. The mirror node already has logic to increment the nonce by one in this case. It only needs to be updated and increment the nonce field of the appropriate EVM address belonging to this account.
We currently don't update the nonces of contract accounts and there is a different approach discussed for contract account nonces anyway (reference: https://github.com/hashgraph/hedera-protobufs/issues/245)

services/basic_types.proto
Nana-EC: We'd basically pull the 20 byte section description out of the alias above and use it here

services/basic_types.proto
Nana-EC: No on deletion the nonce would be null.
The goal was to share the message object and populate the appropriate items on input or output

services/transaction_record.proto
Nana-EC: If we maintain the nonce exposure for the account then simplifying to `VirtualAddress` would be right.
If we say we don't care and aren't concerned with capturing nonce for ContractCall/ContractCreate cases then sure just capturing the evmAddress is fine as `calculated_evm_address`.

Btw, why `calculated_evm_address` vs just `evm_address` or `account_evm_address`

services/basic_types.proto
Nana-EC: Also the VirtualAddress was also intended to expose the newest nonce value to the mirror node on transactions. In which case it's neither an addition nor removal of a virtual address. Just an update to the evmAddresses nonce.

services/basic_types.proto
stoqnkpL: I agree that incrementing an account's nonce only when executing `EthereumTransactions` may be a bit confusing. But it would require very careful consideration if we were to start updating the nonce with each HAPI ContractCall and ContractCreate transaction. Some things we would need to consider:
- If we do start updating the nonce field, then we should verify it always has a correct value in the signed transaction body. But we don't need to do this, Hedera has a different mechanism for protection against replay attacks.
- In Ethereum any transaction would increase the nonce, even a simple transfer, so why not include `CryptoTransfer` transactions as well, and then, why not all other types of transactions as well, like TokenMint? 
- Some external tools that work with the Relay have their own logic for tracking account nonces, and having Hedera increase the nonce value for this account address for all HAPI transactions submitted by this account may just confuse these tools and make them harder to integrate with Hedera
- What about synthetic child transactions, should they increase the nonce of the accounts? Would that mean that one HAPI transaction can increase an account nonce several times (if for example multiple calls to the precompiled contracts are executed)

shemnon: >If we do start updating the nonce field, then we should verify it always has a correct value in the signed transaction body. But we don't need to do this, Hedera has a different mechanism for protection against replay attacks.

`EthereumTransaction` does this.  ContratCallTransaction and all other transactions don't specify a nonce.

> In Ethereum any transaction would increase the nonce, even a simple transfer, so why not include CryptoTransfer transactions as well, and then, why not all other types of transactions as well, like TokenMint?

Ethereum nonce management is a nightmare, we don't want to put that upon ourselves.  We only have nonces in `EthereumTransaction` for semantic compatibility, otherwise we would not track it.

> Some external tools that work with the Relay have their own logic for tracking account nonces, and having Hedera increase the nonce value for this account address for all HAPI transactions submitted by this account may just confuse these tools and make them harder to integrate with Hedera

Which is why it was limited to `EthereumTransaction` - the act of signing a tx with that nonce is where the tooling should track and manage nonces.

> What about synthetic child transactions, should they increase the nonce of the accounts? Would that mean that one HAPI transaction can increase an account nonce several times (if for example multiple calls to the precompiled contracts are executed)

Another reason to only track for top level `EthereumTransaction` objects.

services/basic_types.proto
shemnon: because nonce is an `int64` there is no null, an absent field becomes `0`, which may be sufficient for our needs.  If we need to distinguish from null and 0 we need `Int64Value`

shemnon (commented):
Outside of the comments other reviewers have requested LGTM, so I'll defer my approvals to them.

services/basic_types.proto
Nana-EC: I'm persuaded by the above collective sentiments.
The hope was we'd cover the nonce for EVM interacting transactions only.
However, seems the cons outweigh the pros at this juncture.

services/basic_types.proto
Nana-EC: Yeah `Int64Value` would have been better.
However, seems like no nonce is the way and the `oneof virtual_address_update` suffices.

services/transaction_record.proto
Nana-EC: Just `evm_address` is fine

MiroslavGatsanoga: Updated.

services/basic_types.proto
MiroslavGatsanoga: Updated.

services/basic_types.proto
MiroslavGatsanoga: Updated.

services/basic_types.proto
MiroslavGatsanoga: Updated.

services/basic_types.proto
```diff
@@ -1592,3 +1594,15 @@ message StakingInfo {
         int64 staked_node_id = 6;
     }
 }
+
+message VirtualAddress {
+    /**
+     * The 20-byte EVM address.
```
Nana-EC, line 1600: nit. Add explicit note to says this is the keccak-256 hash of a ECDSA_SECP256K1 primitive key

services/crypto_create.proto
```diff
@@ -175,4 +173,9 @@ message CryptoCreateTransactionBody {
      * cover auto-renewal fees.
      */
     AccountID auto_renew_account = 19;
+
+    /**
+     * EOA 20-byte address to create.
```
Nana-EC, line 178: note keccak-256 hash of a ECDSA_SECP256K1 primitive key

Nana-EC (approved):
Looks good. Just had some comment suggestions to make it explicit where the evm address had to be the keccak-256 hash of a ECDSA_SECP256K1 primitive key

steven-sheehy (approved):

tinker-michaelj (commented):
Makes sense to me! 👍

services/basic_types.proto
```diff
@@ -95,21 +95,23 @@ message AccountID {
...
-         * that map to the EVM address bytes. The provided public key bytes will then serve as the final alias bytes.
          */
         bytes alias = 4;
 
+        /**
+         * The ethereum account 20-byte EVM address to be used initially in place of the public key bytes. This EVM
+         * address may be either the encoded form of the shard.realm.num or the keccak-256 hash of a ECDSA_SECP256K1 primitive key.
+         *
+         * If a transaction lazily-creates this account, a subsequent transaction will be required containing the public key bytes
+         * that map to the EVM address bytes.
```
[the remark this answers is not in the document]
MiroslavGatsanoga: Added.

services/basic_types.proto
MiroslavGatsanoga: Added.

services/crypto_create.proto
MiroslavGatsanoga: Added.

Nana-EC (approved):
LGTM

tinker-michaelj (approved):
LGTM

PR #249, opened 2022-12-06
Add virtual_address_update disable

### LINKED ISSUE #248: Update virtual_address_update with diable logic
### Problem

The initial version of HIP 631 did not yet contain the ability to disable virtual addresses prior to removal.

### Solution

Add `disable` 

### Alternatives

_No response_

Nana-EC:
Signed-off-by: Nana Essilfie-Conduah <nana@swirldslabs.com>

**Description**:
To support the ability to disable a virtual address but keep it associated with an account
- add `disable` to `CryptoUpdateTransactionBody.virtual_address_update`

**Related issue(s)**:

Fixes #248 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/services/crypto_update.proto
+++ b/services/crypto_update.proto
@@ -167,17 +167,8 @@ message CryptoUpdateTransactionBody {
     oneof virtual_address_update {
         /**
          * The virtual address to be added.
+         * The address one successfully addition becomes par tof the accounts ACTIVE virtual addresses
          */
         VirtualAddress add = 20;
-
-        /**
-        * The 20-byte EVM address of the virtual address that is being disabled and therefore removed from active use on the account.
-        */
-        bytes disable = 21;
-        
-        /**
-        * The 20-byte EVM address of the virtual address that is being removed from the account.
-        */
-        bytes remove = 22;
     }
 }
--- a/services/crypto_update.proto
+++ b/services/crypto_update.proto
@@ -171,8 +171,13 @@ message CryptoUpdateTransactionBody {
         VirtualAddress add = 20;
 
         /**
-         * The 20-byte EVM address of the virtual address that is being removed.
-         */
-        bytes remove = 21;
+        * The 20-byte EVM address of the virtual address that is being disabled and therefore removed from active use on the account.
+        */
+        bytes disable = 21;
+        
+        /**
+        * The 20-byte EVM address of the virtual address that is being removed from the account.
+        */
+        bytes remove = 22;
     }
 }
--- a/services/crypto_update.proto
+++ b/services/crypto_update.proto
@@ -167,7 +167,7 @@ message CryptoUpdateTransactionBody {
     oneof virtual_address_update {
         /**
          * The virtual address to be added.
-         * The address one successfully addition becomes par tof the accounts ACTIVE virtual addresses
+         * The address one successfully addition becomes part of the accounts ACTIVE virtual addresses
          */
         VirtualAddress add = 20;
     }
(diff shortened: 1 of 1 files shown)


services/crypto_update.proto
```diff
@@ -171,8 +171,13 @@ message CryptoUpdateTransactionBody {
         VirtualAddress add = 20;
 
         /**
-         * The 20-byte EVM address of the virtual address that is being removed.
-         */
-        bytes remove = 21;
+        * The 20-byte EVM address of the virtual address that is being disabled and therefore removed from active use on the account.
+        */
+        bytes disable = 21;
```
steven-sheehy, line 176: Since a disabled address stays in state, shouldn't this flag be moved to `VirtualAddress`? If queries only return active addresses how would users know which old, inactive addresses they need to delete?

services/crypto_update.proto
```diff
@@ -171,8 +171,13 @@ message CryptoUpdateTransactionBody {
         VirtualAddress add = 20;
 
         /**
-         * The 20-byte EVM address of the virtual address that is being removed.
-         */
-        bytes remove = 21;
+        * The 20-byte EVM address of the virtual address that is being disabled and therefore removed from active use on the account.
```
steven-sheehy, line 174: The only difference between disable and remove is that the former retains ownership and the latter allows reuse. Add a comment to each field about the impacts of address reuse.

steven-sheehy (changes requested):

services/crypto_update.proto
Nana-EC: Yeah good suggestion. Will leave till we're ready to turn on disable

services/crypto_update.proto
Nana-EC: Will do when we circle back

services/crypto_update.proto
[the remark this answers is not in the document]
Nana-EC: That's fair.
We'll coordinate the disable addition.
Will remove as suggested since adding to a oneof later on (even if the ordinal value is greater than existing outer fields) won't break the API

services/crypto_update.proto
```diff
@@ -167,17 +167,8 @@ message CryptoUpdateTransactionBody {
     oneof virtual_address_update {
         /**
          * The virtual address to be added.
+         * The address one successfully addition becomes par tof the accounts ACTIVE virtual addresses
```
steven-sheehy, line 170: nit: part of

Nana-EC: Fixed, thanks

steven-sheehy (approved):

services/crypto_update.proto
```diff
@@ -167,7 +167,7 @@ message CryptoUpdateTransactionBody {
     oneof virtual_address_update {
         /**
          * The virtual address to be added.
-         * The address one successfully addition becomes par tof the accounts ACTIVE virtual addresses
+         * The address one successfully addition becomes part of the accounts ACTIVE virtual addresses
```
kimbor, line 170: Once successfully added, the address becomes part of the account's ACTIVE virtual addresses.

Nana-EC: @kimbor fixed

MarcKriguerAtHedera (approved):
Despite the PR being called "Add virtual_address_update disable", it's really removing the remove action.  If that's what is intended, this little change looks good to me.

Nana-EC:
> Despite the PR being called "Add virtual_address_update disable", it's really removing the remove action. If that's what is intended, this little change looks good to me.

Yes, will move this into draft.
The final form will see `remove` replaced with `disable`

david-bakin-sl:
So what's here now is removing the field `remove` because it is not yet implemented?  If so the title and main comment of this PR should be changed to match?

Nana-EC:
Closing until HIP 631 is revisited with HIP 583 induced changes

PR #250, opened 2022-12-22
Change PBJ package definition from custom option to specially comment

jasperpotts:
PBJ wants different package names for generated classes from the SDKs so we can not change the java_package option to a new package as that will break Java SDK. So we added a custom option for PBJ package, unfortunately custom options cause issues with Go SDK and LITE Java Compiler in Java SDK. So this PR removes Them and replaces them with a specially formatted comment that will be ignored but Goggle ProtoC compiler.


rbair23 (approved):

tinker-michaelj (approved):
LGTM

PR #252, opened 2023-01-11
Temporarily remove auto-renew fields for files, accounts for `v0.34.0` tag

tinker-michaelj:
**Description**:
- Temporarily removes auto-renew metadata for files and accounts.

### DIFF
--- a/services/crypto_create.proto
+++ b/services/crypto_create.proto
@@ -169,13 +169,6 @@ message CryptoCreateTransactionBody {
      bytes alias = 18;
 
     /**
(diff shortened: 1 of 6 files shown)


jasperpotts (approved):

kimbor (approved):

xin-hedera (approved):

jsync-swirlds:
For future reference, as claimed by `steven-sheehy` in another PR, 2024-05-16, the versions of these messages with the now-removed fields were never deployed to any version of any software, so there is no reason to reserve the removed field numbers.

PR #253, opened 2023-01-13
Add sub-interface metadata to the TransactionRecord for PBJ

jasperpotts:
We want TransactionRecord to be built up from a set of super interfaces. This will allow it and its builder to be passed around by those sub interfaces, only exposing a subset of TransactionRecord.


nickpoorman:
@jasperpotts is this stale or do you still need it?

jsync-swirlds:
Stale.  Please resubmit if needed.

PR #257, opened 2023-02-01
Add contract nonces related fields

MiroslavGatsanoga:
Signed-off-by: Miroslav Gatsanoga <miroslav.gatsanoga@limechain.tech>

**Description**:

**Related issue(s)**:

- https://github.com/hashgraph/hedera-services/issues/6512

**Notes for reviewer**:

Adding protobuf changes for HIP-729.

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1608,7 +1608,19 @@ message VirtualAddress {
     bool is_default = 2;
 }
 
+/**
+ * Info about a contract account's nonce value.
+ * A nonce of a contract is only incremented when that contract creates another contract.
+ */
 message ContractNonceInfo {
+
+    /**
+    * Id of the contract
+    */
     ContractID contract_id = 1;
+
+    /**
+    * The current value of the contract account's nonce property
+    */
     int64 nonce = 2;
 }
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -96,23 +96,20 @@ message AccountID {
          * a protobuf Key message for any primitive key type. Currently only primitive key bytes are supported as an alias
          * (ThresholdKey, KeyList, ContractID, and delegatable_contract_id are not supported)
          *
+         * May also be the ethereum account 20-byte EVM address to be used initially in place of the public key bytes. This EVM
+         * address may be either the encoded form of the shard.realm.num or the keccak-256 hash of a ECDSA_SECP256K1 primitive key.
+         *
          * At most one account can ever have a given alias and it is used for account creation if it
          * was automatically created using a crypto transfer. It will be null if an account is created normally.
          * It is immutable once it is set for an account.
          *
          * If a transaction auto-creates the account, any further transfers to that alias will simply be deposited
          * in that account, without creating anything, and with no creation fee being charged.
-         */
-        bytes alias = 4;
-
-        /**
-         * The ethereum account 20-byte EVM address to be used initially in place of the public key bytes. This EVM
-         * address may be either the encoded form of the shard.realm.num or the keccak-256 hash of a ECDSA_SECP256K1 primitive key.
          *
          * If a transaction lazily-creates this account, a subsequent transaction will be required containing the public key bytes
-         * that map to the EVM address bytes. Lazy account creates will only support the keccak-256 hash of a ECDSA_SECP256K1 primitive key form.
+         * that map to the EVM address bytes. The provided public key bytes will then serve as the final alias bytes.
          */
-        bytes evm_address = 5;
+        bytes alias = 4;
     }
 }
 
@@ -1594,18 +1591,6 @@ message StakingInfo {
          */
         int64 staked_node_id = 6;
     }
-}
-
-message VirtualAddress {
-    /**
-     * The 20-byte EVM address that is derived from the keccak-256 hash of a ECDSA_SECP256K1 primitive key.
-     */
-    bytes address = 1;
-
-    /**
-     * Flag if this address should now be set or is the default address on the account.
-     */
-    bool is_default = 2;
 }
 
 /**
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1592,20 +1592,3 @@ message StakingInfo {
         int64 staked_node_id = 6;
     }
 }
-
-/**
- * Info about a contract account's nonce value.
- * A nonce of a contract is only incremented when that contract creates another contract.
- */
-message ContractNonceInfo {
-
-    /**
-    * Id of the contract
-    */
-    ContractID contract_id = 1;
-
-    /**
-    * The current value of the contract account's nonce property
-    */
-    int64 nonce = 2;
-}
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1607,3 +1607,8 @@ message VirtualAddress {
      */
     bool is_default = 2;
 }
+
+message ContractNonceInfo {
+    ContractID contract_id = 1;
+    int64 nonce = 2;
+}
--- a/services/contract_call_local.proto
+++ b/services/contract_call_local.proto
@@ -162,6 +162,12 @@ message ContractFunctionResult {
      * ContractCreateTransactionBody or a ContractCallTransactionBody.     
      */
     AccountID sender_id = 13;
+
+    /**
+    * Map used to track contract nonces for all contracts that have participated in the transaction.
+    * The keys would have the different contract IDs as strings in "shard.realm.num" format and the values are the new nonce values.
+    */
+    map<string, int64> contract_nonces = 14;
 }
 
 /**
--- a/services/contract_call_local.proto
+++ b/services/contract_call_local.proto
@@ -65,13 +65,13 @@ message ContractLoginfo {
 message ContractNonceInfo {
 
     /**
-    * Id of the contract
-    */
+     * Id of the contract
+     */
     ContractID contract_id = 1;
 
     /**
-    * The current value of the contract account's nonce property
-    */
+     * The current value of the contract account's nonce property
+     */
     int64 nonce = 2;
 }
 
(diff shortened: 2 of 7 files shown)


services/contract_call_local.proto
```diff
@@ -162,6 +162,12 @@ message ContractFunctionResult {
      * ContractCreateTransactionBody or a ContractCallTransactionBody.     
      */
     AccountID sender_id = 13;
+
+    /**
+    * Map used to track contract nonces for all contracts that have participated in the transaction.
+    * The keys would have the different contract IDs as strings in "shard.realm.num" format and the values are the new nonce values.
+    */
+    map<string, int64> contract_nonces = 14;
```
stoqnkpL, line 170: Instead of providing the contract ID as a string and having to parse it by each client, an alternative solution will be to create a new type:
```
message ContractNonceInfo {
  ContractID contract_id = 1;
  int64 nonce = 2;
}
```

and then add `repeated ContractNonceInfo = 14` to the ContractFunctionResult type.

MarcKriguerAtHedera: I think Stoyan's idea is the better approach here.

MarcKriguerAtHedera (approved):
Looks good to me.

services/contract_call_local.proto
stoqnkpL: Updated.

services/basic_types.proto
```diff
@@ -1607,3 +1607,8 @@ message VirtualAddress {
      */
     bool is_default = 2;
 }
+
```
Neeharika-Sompalli, line 1610: Please add description for this

stoqnkpL: Added

Neeharika-Sompalli (dismissed):
LGTM !

Nana-EC (dismissed):

services/crypto_get_info.proto
```diff
@@ -182,10 +182,22 @@ message CryptoGetInfoResponse {
...
+        message VirtualAddressInfo {
+            /**
+             * A virtual address entry.
+             */
+            VirtualAddress virtual_address = 1;
+
+            /**
+             * The ethereum transaction nonce associated with this virtual address.
+             */
+            int64 nonce = 2;
```
steven-sheehy, line 194: From an end user perspective, this is bit unwieldy. I think we should have a flattened structure returned by queries that has all the necessary info in one `message`. I understand for code reuse it makes sense to have the above but we have to design APIs for end users.

services/contract_call_local.proto
```diff
@@ -162,6 +162,11 @@ message ContractFunctionResult {
      * ContractCreateTransactionBody or a ContractCallTransactionBody.     
      */
     AccountID sender_id = 13;
+
+    /**
+    * A list of updated contract account nonces containing the new nonce value for each contract account
+    */
+    repeated ContractNonceInfo contract_nonce_updates = 14;
```
steven-sheehy, line 169: Perhaps the simpler
```suggestion
    repeated ContractNonceInfo contract_nonces = 14;
```

steven-sheehy (changes requested):

services/crypto_get_info.proto
xin-hedera, line 194: for consistency, this field should be `ethereum_nonce`

Nana-EC: @steven-sheehy what's the suggestion here to flatten it?

Nana-EC: Agree with consistency but i think we should make an exception and leave it as `nonce`

services/crypto_get_info.proto
```diff
@@ -182,10 +182,22 @@ message CryptoGetInfoResponse {
...
         StakingInfo staking_info = 22;
 
+        message VirtualAddressInfo {
+            /**
+             * A virtual address entry.
+             */
+            VirtualAddress virtual_address = 1;
+
+            /**
+             * The ethereum transaction nonce associated with this virtual address.
```
Nana-EC, line 192:
```suggestion
             * The account transaction nonce associated with this virtual address.
```

services/crypto_get_info.proto
steven-sheehy:
```proto
message VirtualAddressInfo {
    /**
     * The 20-byte EVM address that is derived from the keccak-256 hash of a ECDSA_SECP256K1 primitive key.
     */
    bytes address = 1;

    /**
     * Flag if this address should now be set or is the default address on the account.
     */
    bool is_default = 2;

    /**
     * The ethereum transaction nonce associated with this virtual address.
     */
    int64 nonce = 3;
}

Comment (MiroslavGatsanoga) [reply services/crypto_get_info.proto]: Thanks @steven-sheehy, updated as suggested.

File: services/contract_call_local.proto (hunk 4 above)
Comment (MiroslavGatsanoga) [reply services/contract_call_local.proto]: Updated.

File: services/crypto_get_info.proto (hunk 5 above)
Comment (MiroslavGatsanoga) [reply services/crypto_get_info.proto]: Updated.

**Review by Nana-EC — APPROVED** [review]

**Review by Nana-EC — DISMISSED** [review]

**Review by steven-sheehy — APPROVED** [review]

**Review by stoqnkpL — DISMISSED** [review]

**Review by stoqnkpL — DISMISSED** [review]

**Review by Nana-EC — DISMISSED** [review]

**Review by steven-sheehy — DISMISSED** [review]

**Review by Neeharika-Sompalli — APPROVED** [review]

**Review by xin-hedera — APPROVED** [review]

**Review by MarcKriguerAtHedera — DISMISSED** [review]
LGTM as well -- but note that as HIP 631 was deferred (if not cancelled), there are no short-term plans for the mirror node to make any chances as a result of this PR going in.

**Comment by Nana-EC** [comment]
> LGTM as well -- but note that as HIP 631 was deferred (if not cancelled), there are no short-term plans for the mirror node to make any chances as a result of this PR going in.

@MarcKriguerAtHedera though mentioned in HIP 631 this is more of an equivalence bug than HIP 631's main focus of multiple addresses.
We'll bring more clarity once HIP 583 update is merged but Mirror Node should plan to support this fix regardless.
Will reach out

File: services/basic_types.proto [hunk 6]
```diff
@@ -1592,3 +1592,20 @@ message StakingInfo {
...
+    /**
+    * Id of the contract
+    */
+    ContractID contract_id = 1;
+
+    /**
+    * The current value of the contract account's nonce property
+    */
+    int64 nonce = 2;
+}
```
rbair23, line 1611: Is there a reason this is in basic_types.proto instead of contract_call_local.proto?

MiroslavGatsanoga: Checked with @stoqnkpL and we don't think so, will move it to `contract_call_local.proto`.

steven-sheehy (dismissed):
LGTM

xin-hedera (approved):
LGTM

Nana-EC (approved):

services/contract_call_local.proto
```diff
@@ -58,6 +58,23 @@ message ContractLoginfo {
...
 
+/**
+ * Info about a contract account's nonce value.
+ * A nonce of a contract is only incremented when that contract creates another contract.
+ */
+message ContractNonceInfo {
+
+    /**
+    * Id of the contract
+    */
```
rbair23, line 69: Indentation on these comments (and other new comments) are off.

services/contract_call_local.proto
```diff
@@ -58,6 +58,23 @@ message ContractLoginfo {
     bytes data = 4;
 }
 
+/**
+ * Info about a contract account's nonce value.
+ * A nonce of a contract is only incremented when that contract creates another contract.
+ */
+message ContractNonceInfo {
```
rbair23, line 65: I'm a little confused about this type appearing in the contract_call_local.proto. Does it also apply to normal contract calls?

MiroslavGatsanoga: It is in this file because of `ContractFunctionResult` (it is used only there) and the idea was to keep related types close together.
The comment on `ContractFunctionResult` states that:
```
This is part of the response to a ContractCallLocal query, and is in the record for a ContractCall or ContractCreateInstance transaction.
```
so yes, as `ContractNonceInfo` is part of `ContractFunctionResult` it will apply to normal contract calls as well.

services/contract_call_local.proto
```diff
@@ -162,6 +179,11 @@ message ContractFunctionResult {
      * ContractCreateTransactionBody or a ContractCallTransactionBody.     
      */
     AccountID sender_id = 13;
+
+    /**
+    * A list of updated contract account nonces containing the new nonce value for each contract account
```
tinker-michaelj, line 184: We could note, 
```
This is always empty in the `ContractFunctionResult` field of a `ContractCallLocalResponse`, since no internal creations can happen in a static EVM call.
```

services/contract_call_local.proto
Nana-EC: I think the issue is that `ContractFunctionResult` should likely not have been in `contract_call_local.proto`.
Placing `ContractNonceInfo` here seems to be a colocation attempt for readability since `ContractFunctionResult` is the intended consumer of it.

Another option would be `basic_types.proto` or a new custom file for Contract related messages.

@rbair23 does this current location work or would you prefer it in another location as suggested?

services/contract_call_local.proto
MiroslavGatsanoga: Fixed.

services/contract_call_local.proto
MiroslavGatsanoga: Added.

services/contract_call_local.proto
MiroslavGatsanoga: Btw now I noticed [this comment](https://github.com/hashgraph/hedera-protobufs/pull/257#discussion_r1117668864) and it seems that initially we had it in `basic_types.proto` but we moved it here.

Nana-EC: @MiroslavGatsanoga can you move this to a new `contract_types.proto` file.
I think that should address concerns and we can get this PR in to unblock HIP 729 development.
In a future protobuf version we can explore moving more contract types into this file

steven-sheehy (approved):

nickpoorman (approved):

Neeharika-Sompalli (approved):

PR #259, opened 2023-02-07
Remove the notion of virtual addresses and replace with alias(es)

lukelee-sl:
Signed-off-by: lukelee-sl <luke.lee@swirldslabs.com>

**Description**:
HIP 583 and HIP 631 introduce the idea of virtual addresses.  Upon discussion, it was decided 
to try to replace virtual addresses by extending the existing alias(es).

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -104,15 +104,6 @@ message AccountID {
          * in that account, without creating anything, and with no creation fee being charged.
          */
         bytes alias = 4;
-
-        /**
-         * The ethereum account 20-byte EVM address to be used initially in place of the public key bytes. This EVM
-         * address may be either the encoded form of the shard.realm.num or the keccak-256 hash of a ECDSA_SECP256K1 primitive key.
-         *
-         * If a transaction lazily-creates this account, a subsequent transaction will be required containing the public key bytes
-         * that map to the EVM address bytes. Lazy account creates will only support the keccak-256 hash of a ECDSA_SECP256K1 primitive key form.
-         */
-        bytes evm_address = 5;
     }
 }
 
@@ -1594,16 +1585,4 @@ message StakingInfo {
          */
         int64 staked_node_id = 6;
     }
-}
-
-message VirtualAddress {
-    /**
-     * The 20-byte EVM address that is derived from the keccak-256 hash of a ECDSA_SECP256K1 primitive key.
-     */
-    bytes address = 1;
-
-    /**
-     * Flag if this address should now be set or is the default address on the account.
-     */
-    bool is_default = 2;
 }
--- a/services/crypto_get_info.proto
+++ b/services/crypto_get_info.proto
@@ -182,10 +182,27 @@ message CryptoGetInfoResponse {
          */
         StakingInfo staking_info = 22;
 
+        message AliasInfo {
+            /**
+             * The alias
+             */
+            bytes alias = 1;
+
+            /**
+             * Flag if this alias should now be set or is the default alias on the account.
+             */
+            bool is_default = 2;
+
+            /**
+             * The EthereumTransaction type nonce associated with this alias.
+             */
+            int64 nonce = 3;
+        }
+
         /**
-         * List of virtual addresses each of which is an EVM address that maps to an ECDSA key pair a user must prove ownership of.
+         * List of alias information each of which is an EVM address that maps to an ECDSA key pair a user must prove ownership of.
          */
-        repeated VirtualAddress virtual_addresses = 24;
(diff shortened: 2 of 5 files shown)


services/basic_types.proto
```diff
@@ -104,15 +104,6 @@ message AccountID {
          * in that account, without creating anything, and with no creation fee being charged.
          */
         bytes alias = 4;
-
-        /**
-         * The ethereum account 20-byte EVM address to be used initially in place of the public key bytes. This EVM
```
MiroslavGatsanoga, line 109: This description was taken from the description of the `alias` field above, we should move it back there since we will be expecting EVM address values in the `alias` field.

MiroslavGatsanoga (commented):
One additional thing to consider: since this PR is dealing with the changes introduced in https://github.com/hashgraph/hedera-protobufs/pull/242, one of the changes previously introduced was to add `evm_address` field to [`transaction_record.proto`](https://github.com/hashgraph/hedera-protobufs/pull/242/files#diff-282b092daf92d72c0573a1e75fb80df9c68ce358c2fab955956a5f744878fd5aR148-R153). 
I am not sure if we want to remove that as well, just mentioning it so that it is not accidentally left out.

lukelee-sl:
> One additional thing to consider: since this PR is dealing with the changes introduced in #242, one of the changes previously introduced was to add `evm_address` field to [`transaction_record.proto`](https://github.com/hashgraph/hedera-protobufs/pull/242/files#diff-282b092daf92d72c0573a1e75fb80df9c68ce358c2fab955956a5f744878fd5aR148-R153). I am not sure if we want to remove that as well, just mentioning it so that it is not accidentally left out.

@Nana-EC This is a good question.  Should we update the name of the field to `alias`?  `evm_address` I think seems more accurate in this context but what do you think?

stoqnkpL (changes requested):

services/crypto_get_info.proto
```diff
@@ -182,10 +182,27 @@ message CryptoGetInfoResponse {
          */
         StakingInfo staking_info = 22;
 
+        message AliasInfo {
```
Nana-EC, line 185: We should remove this whiles we decide when 631

Nana-EC (changes requested):
583 focused suggestion

services/crypto_update.proto
```diff
@@ -155,16 +155,4 @@ message CryptoUpdateTransactionBody {
      * If true, the account declines receiving a staking reward. The default value is false.
      */
     google.protobuf.BoolValue decline_reward = 18;
-
-    oneof virtual_address_update {
-        /**
-         * The virtual address to be added.
-         */
-        VirtualAddress add = 20;
```
Nana-EC, line 163: Can you also remove `VirtualAddress` from `basic_types.proto`

Nana-EC (commented):
LG, also remove VirtualAddress from basic_types.proto

services/crypto_update.proto
lukelee-sl: `VirtualAddress` was removed `basic_types.proto` from in a previous commit

Nana-EC: I missed that, thanks

stoqnkpL:
> Do we still need `TransactionRecord.evm_address` from #242?

We populate `evm_address` when we execute a `CryptoCreate` that has as input args an `ECDSA` key and we want to externalize the calculated EVM address assigned to the account too. I think the Mirror Node requested this, so maybe we should discuss this before deleting it.

Nana-EC:
> > Do we still need `TransactionRecord.evm_address` from #242?
> 
> We populate `evm_address` when we execute a `CryptoCreate` that has as input args an `ECDSA` key and we want to externalize the calculated EVM address assigned to the account too. I think the Mirror Node requested this, so maybe we should discuss this before deleting it.

Yeah Mirror Node requested this so it explicitly had the value. This will now only be applicable when an ECDS key alias is set.

Nana-EC (approved):

steven-sheehy (approved):

stoqnkpL (approved):

Neeharika-Sompalli (approved):

tinker-michaelj (approved):
LGTM

nickpoorman (approved):

jsync-swirlds:
For future reference, as claimed by `steven-sheehy` in another PR, 2024-05-16, the versions of these messages with the now-removed fields were never deployed to any version of any software, so there is no reason to reserve the removed field numbers.

PR #263, opened 2023-03-30
Initial draft of PBJ types for token state

Neeharika-Sompalli:
Fixes https://github.com/hashgraph/hedera-services/issues/5804
Fixes https://github.com/hashgraph/hedera-protobufs/issues/264

Adds **initial draft** of PBJ types Account, Token, Token-relation for token-service implementation.

Currently same fields in `MerkleAccount`, `MerkleToken` and `MerkleTokenRelStatus` are used in the protobuf messages. They are subject to change in the future.


services/state/token/account.proto
```diff
@@ -0,0 +1,175 @@
...
+ *
+ * As with all network entities, an account has a unique entity number, which is usually given along
+ * with the network's shard and realm in the form of a shard.realm.number id.
+ */
+
+message Account {
+    /**
+     * The unique entity number of the account.
+     */
+    int64 accountNumber = 1;
```
tinker-michaelj, line 42: `account_number`

services/state/token/account.proto
```diff
@@ -0,0 +1,175 @@
...
+     */
+    int64 auto_renew_secs = 22;
+    /**
+     * If this account is a smart-contract, number of key-value pairs stored on the contract.
+     */
+    int32 num_contract_kv_pairs = 23;
+    /**
+     * (Optional) List of crypto allowances approved by the account.
+     */
+    repeated CryptoAllowance cryptoAllowances = 24;
```
tinker-michaelj, line 140: `crypto_allowances`

services/state/token/account.proto
```diff
@@ -0,0 +1,175 @@
...
+    /**
+     * (Optional) List of non-fungible token allowances approved by the account.
+     */
+    repeated NftAllowance nftAllowances = 25;
```
tinker-michaelj, line 144: `nft_allowances`

services/state/token/token.proto
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * As with all network entities, a token has a unique entity number, which is usually given along
+ * with the network's shard and realm in the form of a shard.realm.number id.
+ */
+
+message Token {
+    /**
+     * The unique entity number of this token.
+     */
+    int64 tokenNum = 1;
```
tinker-michaelj, line 43: `token_num`

services/state/token/token.proto
```diff
@@ -0,0 +1,144 @@
...
+     */
+    bool accounts_kyc_granted_by_default = 23;
+    /**
+     * The type of this token.
+     */
+    TokenType tokenType = 24;
+    /**
+     * The supply type of this token.
+     */
+    TokenSupplyType supplyType = 25;
```
tinker-michaelj, line 139: `supply_type`

services/state/token/token_relation.proto
```diff
@@ -0,0 +1,73 @@
...
+     */
+    bool kycGranted = 5;
+    /**
+     * The flag indicating if the token relationship was deleted.
+     */
+    bool deleted = 6;
+    /**
+     * The flag indicating if the token relationship was created using automatic association.
+     */
+    bool automaticAssociation = 8;
```
tinker-michaelj, line 64: `automatic_association`

services/state/token/token.proto
Neeharika-Sompalli: Fixed. Thanks

services/state/token/token.proto
Neeharika-Sompalli: Fixed.

services/state/token/account.proto
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * (Optional) List of fungible token allowances approved by the account.
+     */
+    repeated TokenAllowance tokenAllowances = 26;
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Fixed. Thanks

services/state/token/account.proto
Neeharika-Sompalli: Fixed. Thanks

services/state/token/account.proto
Neeharika-Sompalli: Fixed. Thanks

mhess-swl (approved):

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+     */
+    uint64 head_token_number = 27;
+    /**
+     * The NftId of the head of the linked list from unique tokens map for the account.
+     */
+    uint64 head_nft_id = 28;
+    /**
+      * The serial number of the head NftId of the linked list from unique tokens map for the account.
+      */
+    uint64 head_nft_serial_number = 29;
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Yes @xin-hedera . Thanks. Fixed

services/state/token/token_relation.proto
```diff
@@ -0,0 +1,73 @@
...
+ */
+message TokenRelation {
+    /**
+     * The token involved in this relation.
+     */
+    uint64 token_number = 1;
+    /**
+     * The account involved in this association.
+     */
+    uint64 account_number = 2;
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Fixed. Thanks

services/state/token/token_relation.proto
```diff
@@ -0,0 +1,73 @@
...
+option java_package = "com.hederahashgraph.api.proto.java";
+// <<<pbj.java_package = "com.hedera.hapi.node.state.token">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+/**
+ * First-draft representation of a Hedera Token Service token relationship entity in the network Merkle tree.
+ *
+ * As with all network entities, a token relationship has a unique entity number pair, which is represented
+ * with the account and the token involved in the relationship.
+ */
+message TokenRelation {
```
jsync-swirlds, line 36: Is this related to the existing TokenRelationship entity?

Neeharika-Sompalli: yes, `MerkleTokenRelStatus`

services/state/token/token_relation.proto
```diff
@@ -0,0 +1,73 @@
...
+     */
+    int64 token_number = 1;
+    /**
+     * The account involved in this association.
+     */
+    int64 account_number = 2;
+    /**
+     * The balance of the token relationship.
+     */
+    int64 balance = 3;
```
jsync-swirlds, line 48: Do we want this to be a flat number, or would it make sense to use a TokenBalance entity?

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+     */
+    Key key = 3;
+    /**
+     * The expiration time of the account, in seconds since the epoch.
+     */
+    int64 expiry = 4;
+    /**
+     * The balance of the account, in tiny-bars.
+     */
+    int64 hbar_balance = 5;
```
tinker-michaelj, line 59: There's a few spots we can be more expressive about the units of a quantity...maybe here `tinybar_balance`?

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+     */
+    int32 used_auto_associations = 12;
+    /**
+     * The number of tokens associated with the account.
+     */
+    int32 number_associations = 13;
+    /**
+     * The number of tokens with a positive balance associated with the account.
+     */
+    int32 number_positive_balances = 14;
```
tinker-michaelj, line 95: `number_non_zero_token_balances` maybe?

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * The nonce of the account, used for Ethereum interoperability.
+     */
+    int64 ethereum_nonce = 15;
+    /**
+     * The amount of hbars staked to the account.
+     */
+    int64 staked_to_me = 16;
```
tinker-michaelj, line 103: `whole_hbars_staked_to_me`?

tinker-michaelj (dismissed):
All seems quite reasonable!

services/state/token/token.proto
```diff
@@ -0,0 +1,144 @@
...
+option java_multiple_files = true;
+
+/**
+ * First-draft representation of a Hedera Token Service token entity in the network Merkle tree.
+ *
+ * As with all network entities, a token has a unique entity number, which is usually given along
+ * with the network's shard and realm in the form of a shard.realm.number id.
+ */
+
+message Token {
```
jsync-swirlds, line 39: Is this related to the existing TokenInfo entity?

services/state/token/token_relation.proto
jsync-swirlds: I guess what I'm asking, is if we're duplicating structure here, changing existing entities, or creating a new and unrelated entity for internal use?
If we're duplicating the entity, we need to be careful about unsynchronized changes to the two and should probably document in both that they're related and need to be kept in sync.
If we're changing the existing entity, we probably need to document that carefully and deprecate the other definition.
If we're creating a new (more-or-less) unrelated entity, then it should probably be documented here that this isn't a replacement or compliment to that entity.

services/state/token/token.proto
Neeharika-Sompalli: It is `MerkleToken` entity

services/state/token/token_relation.proto
Neeharika-Sompalli: I think the idea is we will not have `Merkle*` entities after refactor. We will be using the PBJ types in all Maps in state.

services/state/token/token_relation.proto
Neeharika-Sompalli: It is a long value [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/merkle/MerkleTokenRelStatus.java#L47).

services/state/token/account.proto
Neeharika-Sompalli: Yes, done. Thanks !

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * The expiration time of the account, in seconds since the epoch.
+     */
+    int64 expiry = 4;
+    /**
+     * The balance of the account, in tiny-bars.
+     */
+    int64 hbar_balance = 5;
+    /**
+     * An optional description of the topic with UTF-8 encoding up to 100 bytes.
```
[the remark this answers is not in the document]
Neeharika-Sompalli: It is enforced on the Transaction level. We throw `MEMO_TOO_LONG` if its larger than 100 bytes

xin-hedera (approved):

services/state/token/token_relation.proto
jsync-swirlds: Thanks.  So this is replacing internal data structures, not replicating or replacing external entities.  That makes sense, and helps to understand where these fit.

jsync-swirlds (dismissed):

tinker-michaelj (approved):
LGTM!

services/state/token/token_relation.proto
```diff
@@ -0,0 +1,73 @@
...
+     */
+    bool kyc_granted = 5;
+    /**
+     * The flag indicating if the token relationship was deleted.
+     */
+    bool deleted = 6;
+    /**
+     * The flag indicating if the token relationship was created using automatic association.
+     */
+    bool automatic_association = 8;
```
nickpoorman, line 64: How come 7 is skipped and this protobuf field went directly to 8?

nickpoorman (changes requested):

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+ * with the network's shard and realm in the form of a shard.realm.number id.
+ */
+
+message Account {
+    /**
+     * The unique entity number of the account.
+     */
+    int64 account_number = 1;
+    /**
+     * The public key bytes to be used as the account's alias, if any.
```
rbair23, line 45:
```suggestion
     * The alias to use for this account, if any.
```

Alias is no longer just key bytes, so either we put in a long description of the possible encodings, or we just keep it short. Either way works.

services/state/token/account.proto
rbair23, line 1: We should move from `services/state/token` to `state/token`. Ideally we would have had `services/state` and `services/grpc` with sub folders for all the different kinds of schemas, but unfortunately we didn't and moving things around now breaks clients. And putting a mixture of state protobuf schemas and grpc protobuf schemas seems like it will become quite confusing. So I think moving `state` up to be a top level directory would be better.

rbair23: And we should have a README that says that the `state` directory and sub directories are in preview mode and not final.

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+
+option java_package = "com.hederahashgraph.api.proto.java";
+// <<<pbj.java_package = "com.hedera.hapi.node.state.token">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+/**
+ * First-draft representation of a Hedera Token Service account entity in the network Merkle tree.
+ *
+ * As with all network entities, an account has a unique entity number, which is usually given along
+ * with the network's shard and realm in the form of a shard.realm.number id.
```
rbair23, line 36: Not a blocker for this PR, but we really should have better docs here. These schemas will generate Java objects that we will use in our code. So if I'm working on some code and I want to understand the Account class, I'm going to be looking at and reading that javadoc. So It would be nice to describe more what an account is, how it fits into our system, etc. At least at a high level.

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+ */
+
+message Account {
+    /**
+     * The unique entity number of the account.
```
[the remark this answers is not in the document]
rbair23: It might also be good to document that although we have an int64, only non-negative values are permitted.

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+     */
+    int64 account_number = 1;
+    /**
+     * The public key bytes to be used as the account's alias, if any.
+     */
+    bytes alias = 2;
+    /**
+     * (Optional) The key to be used to sign transactions from the account, if any.
```
rbair23, line 49: We should describe a little better as to when having a key is optional and when it is not. Most accounts require it. Only special accounts like 0.0.800 permit an empty key. Maybe we should document that.

services/state/token/account.proto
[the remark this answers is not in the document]
rbair23: Later we should debate whether the shard and realm should be included here or not. Because I'm not sure if we need them to be part of the object for state proofs or not.

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+     */
+    Key key = 3;
+    /**
+     * The expiration time of the account, in seconds since the epoch.
+     */
+    int64 expiry = 4;
+    /**
+     * The balance of the account, in tiny-bars.
+     */
+    int64 tinybar_balance = 5;
```
rbair23, line 59: Minor nit: should it be `tinybar_balance` or `tiny_bar_balance`? The difference will be `TinybarBalance` or `TinyBarBalance`. I think what you have is right, I was just wondering.

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+     */
+    bool deleted = 7;
+    /**
+     * A boolean marking if the account is a smart contract.
+     */
+    bool smart_contract = 8;
+    /**
+     * A boolean marking if the account requires a receiver signature.
+     */
+    bool receiver_sig_required = 9;
```
rbair23, line 75: These all look good. However there is one point worth making, which is that in protobuf serialization, false is represented by omitting the data from the data, whereas "true" is represented by a single byte. So wherever we can, we should make sure the "false" setting is more frequent than the "true". In these three cases, that is what we see. most accounts have `deleted` as false, most accounts are not smart contracts, and most accounts don't have `receiver_sig_required` set to true, so for most accounts these three fields will encode as 0 bytes. Which is great.

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * The number of NFTs owned by the account.
+     */
+    int64 number_owned_nfts = 10;
```
rbair23, line 79: Is there any reason we use `int64` here, but `int32` in associations? Should we use `int32` in all these places? I'm not sure if anybody is going to have more than 2 billion NFTs, but I guess it could happen... but maybe they should use multiple accounts at that point?

At the protobuf level int64 and int32 are just as efficient up until the limit of int32. But at the Java level it is a long vs. an int. Otherwise it wouldn't matter that much.

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+     */
+    int32 max_auto_associations = 11;
+    /**
+     * The number of used auto-association slots.
+     */
+    int32 used_auto_associations = 12;
+    /**
+     * The number of tokens associated with the account.
+     */
+    int32 number_associations = 13;
```
rbair23, line 91: What happens if number_associations is maxed out at the int32 max value, but used_auto_associations are not maxed out, and somebody tries to then create a new auto_association? Do we error out, or do we end up overflowing on number_associations? Just queries.

rbair23: Also, what do we use number_associations for? Is it just for a response to a query?

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * The number of tokens with a positive balance associated with the account.
+     */
+    int32 number_positive_balances = 14;
```
rbair23, line 95: Is this only needed to answer some queries? Or do we use it during checking for solvency? Or something else?

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * The amount of hbars staked to the account at the start of the last rewarded period.
+     */
+    int64 stake_at_start_of_last_rewarded_period = 20;
+    /**
+     * (Optional) The number of an auto-renew account, in the same shard and realm as the account, that
+     * has signed a transaction allowing the network to use its balance to automatically extend the account's
+     * expiration time when it passes.
+     */
+    int64 auto_renew_account_number = 21;
```
rbair23, line 128: If we end up with shard and realm encoded in these files, we'll need it here as well. Otherwise we can leave it off.

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * The number of seconds the network should automatically extend the account's expiration by, if the
+     * account has a valid auto-renew account, and is not deleted upon expiration.
+     */
+    int64 auto_renew_secs = 22;
```
rbair23, line 133: What is the behavior if this is ever 0? Do we just make sure it never is?

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * If this account is a smart-contract, number of key-value pairs stored on the contract.
+     */
+    int32 contract_kv_pairs_number = 23;
```
rbair23, line 137: This appears to be the only field that is smart contract specific (other than the smart_contract boolean). I'm wondering why we need it? Is it again for the purposes of a query?

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * (Optional) List of fungible token allowances approved by the account.
+     */
+    repeated TokenAllowance token_allowances = 26;
+    /**
+     * The token ID of the head of the linked list from token relations map for the account.
+     */
+    int64 head_token_number = 27;
+    /**
+     * The NftId of the head of the linkeint64d list from unique tokens map for the account.
```
rbair23, line 156: `linkeint64d` looks like a typo

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+    repeated CryptoAllowance crypto_allowances = 24;
+    /**
+     * (Optional) List of non-fungible token allowances approved by the account.
+     */
+    repeated NftAllowance nft_allowances = 25;
+
+    /**
+     * (Optional) List of fungible token allowances approved by the account.
+     */
+    repeated TokenAllowance token_allowances = 26;
```
rbair23, line 150: I'm not sure these are right. It seems like we should be using our "map of linked list nodes" concept here instead. It isn't a blocker for this PR, but we need to discuss this.

services/state/token/token.proto
```diff
@@ -0,0 +1,144 @@
...
+     */
+    int64 token_number = 1;
+    /**
+     * The name of this token.
+     */
+    string name = 2;
+    /**
+     * The symbol of this token.
+     */
+    string symbol = 3;
```
rbair23, line 51: We should document max lengths

services/state/token/token.proto
```diff
@@ -0,0 +1,144 @@
...
+     */
+    Key supply_key = 11;
+    /**
+     * (Optional) The fee schedule key of this token.
+     */
+    Key fee_schedule_key = 12;
+    /**
+     * (Optional) The pause key of this token.
+     */
+    Key pause_key = 13;
```
rbair23, line 91: We should document the expectation for these keys -- like if a key is not set when the object is created, it cannot be set later, or only if there is an admin key, or whatever. Just so whoever is working in the code can see the docs right there, and not have to go back and look at the grpc objects to understand the invariants.

services/state/token/token.proto
```diff
@@ -0,0 +1,144 @@
...
+     */
+    bool accounts_frozen_by_default = 22;
+    /**
+      * The flag indicating if this token has accounts associated to it that are KYC granted by default.
+     */
+    bool accounts_kyc_granted_by_default = 23;
+    /**
+     * The type of this token.
+     */
+    TokenType token_type = 24;
```
rbair23, line 135: Ooh, I just realized, we're reusing the HAPI objects here, which is interesting. It isn't a bad thing, but I wonder if it will give the PBJ compiler fits when it is generating code, if `state` is at the top level and not under `services`...

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ‍
+ */
+
+import "basic_types.proto";
+import "crypto_approve_allowance.proto";
+
+option java_package = "com.hederahashgraph.api.proto.java";
```
rbair23, line 28: We should use the same package name here as we do for PBJ, just for clarity (no reason for it to be different).

services/state/token/account.proto
```diff
@@ -0,0 +1,176 @@
...
+// <<<pbj.java_package = "com.hedera.hapi.node.state.token">>> This comment is special code for setting PBJ Compiler java package
```
rbair23, line 29: This doesn't seem like the right package. Should it be in `com.hedera.node.app.service.token.model` or something like that.

services/state/token/account.proto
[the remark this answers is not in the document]
nickpoorman: If only non-negative numbers are permitted, should we make it a `uint32` or `uint64` and deal with it on the backend?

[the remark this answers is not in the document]
Neeharika-Sompalli: Agreed @rbair23 . I think shard and realm can be common for a node in state or somewhere in one place.

services/state/token/account.proto
Neeharika-Sompalli: Makes sense. Thanks

services/state/token/account.proto
Neeharika-Sompalli: Yeah, I wondered about the same and confirmed `tinybar` is being used in sdks and other places in code🙂

Neeharika-Sompalli:
> I'm not sure about where we should put things. Ideally I like moving it out of `services` just so we don't have confusion about states vs. HAPI/gRPC, but maybe right now we have no choice but to put it where you have it in this PR. I had a lot of comments and suggestions, but I don't want to hold up development for them right now, but they are questions we will need to have answered before we're ready for production.
> 
> 1. Where to put things?
> 2. Do we need to encode shard and realm as well?
> 3. Should we reuse HAPI objects? (If we do reuse them, it makes it almost impossible to fork these objects if we need more state or something else later)
> 4. What should the java package be for generated code? And what module runs the compiler to build it?

As per the discussion
1. We will retain them under `services/state`
2. We will need to add `shard` and `realm` to the objects. Need to check with platform team before adding them. So will leave for now on this PR.
3. We will reuse HAPI objects. But the PBJ compiler will keep generated objects under different directories. 
4. Decided on `pbj.java_package =  "com.hedera.hapi.node.state.token"` and `java_package = "com.hederahashgraph.api.proto.java.state"`

services/state/token/account.proto
Neeharika-Sompalli: Fixed as per discussion.

services/state/token/account.proto
Neeharika-Sompalli: Fixed as per discussion.

services/state/token/token.proto
Neeharika-Sompalli: As discussed , PBJ compiler will put under different directories i the future.

services/state/token/account.proto
Neeharika-Sompalli: Great insight. Thanks !

services/state/token/account.proto
Neeharika-Sompalli: Just used int64 where we were using long and `int32` for `int`. Will revisit before these going to production.

services/state/token/token.proto
Neeharika-Sompalli: Done.

services/state/token/account.proto
Neeharika-Sompalli: I think this is needed for Storage rent.

services/state/token/account.proto
Neeharika-Sompalli: Checked with @tinker-michaelj . It is not only needed for query, but also for calculating fee for renewal.

services/state/token/account.proto
Neeharika-Sompalli: It is needed because account can't be deleted if there are any positive balances in it.

services/state/token/account.proto
Neeharika-Sompalli: Yeah we are checking if auto renew seconds is in the range in precheck based on these properties
```
ledger.autoRenewPeriod.maxDuration=8000001
ledger.autoRenewPeriod.minDuration=2592000
```

services/state/token/account.proto
Neeharika-Sompalli: And hollow accounts ?

services/state/token/token.proto
Neeharika-Sompalli: Done

services/state/token/account.proto
```diff
@@ -0,0 +1,185 @@
...
+     */
+    string memo = 6;
+    /**
+     * A boolean marking if the account has been deleted.
+     */
+    bool deleted = 7;
+    /**
+     * A boolean marking if the account is a smart contract.
+     */
+    bool smart_contract = 8;
```
jsync-swirlds, line 75: reordering fields one of the few ways to break protobuf compatabilty.  Do we want to move this to an ordinal higher than 16 now?

services/state/token/token.proto
```diff
@@ -0,0 +1,164 @@
...
+     */
+    int64 expiry = 18;
+    /**
+     * An optional description of the token with UTF-8 encoding up to 100 bytes.
+     */
+    string memo = 19;
+    /**
+     * The maximum supply of this token.
+     */
+    int64 max_supply = 20;
```
jsync-swirlds, line 137: Should this be in the first 16 fields?

Neeharika-Sompalli: I think all keys should also be in first 16 fields , so the fields are already filled in. 🙂

services/state/token/account.proto
Neeharika-Sompalli: yes fixed. Thanks !

jsync-swirlds (approved):

rbair23 (approved):

nickpoorman (approved):

xin-hedera (approved):

PR #269, opened 2023-04-17
Change allowance maps on `Account` PBJ Object

Neeharika-Sompalli:
Since `Owner` field is not required for the allowances map on a specific account, added new messages without `Owner` field in `Account` PBJ Object for state.

### DIFF
--- a/services/state/token/account.proto
+++ b/services/state/token/account.proto
@@ -159,16 +159,16 @@ message Account {
     /**
      * (Optional) List of crypto allowances approved by the account.
      */
-    repeated CryptoAllowance crypto_allowances = 27;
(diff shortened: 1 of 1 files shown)


tinker-michaelj (approved):
LGTM! I noticed old types were awkward when writing [this](https://github.com/hashgraph/hedera-services/blob/replay-develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/utils/replay/PbjLeafConverters.java#L92) `Merkle`-to-PBJ leaf converter. 😝

jsync-swirlds (approved):

povolev15 (approved):

PR #270, opened 2023-04-18
Add missing fields in Account

Neeharika-Sompalli:
- Add missing fields in `account.proto`
- Change allowance field names

### DIFF
--- a/services/state/token/account.proto
+++ b/services/state/token/account.proto
@@ -23,7 +23,6 @@ package proto;
  */
 
 import "basic_types.proto";
-import "crypto_approve_allowance.proto";
 
 option java_package = "com.hederahashgraph.api.proto.java";
 // <<<pbj.java_package = "com.hedera.hapi.node.state.token">>> This comment is special code for setting PBJ Compiler java package
@@ -193,7 +192,7 @@ message Account {
      * The first key in the doubly-linked list of this contract's storage mappings;
      * It will be null if if the account is not a contract or the contract has no storage mappings.
      */
-    Uint256Value first_contract_storage_key = 32;
+    Int256Value first_contract_storage_key = 32;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


services/state/token/account.proto
```diff
@@ -189,29 +189,78 @@ message Account {
      * and the time the system task actually auto-renews it.
      */
     bool expired_and_pending_removal = 31;
+    /**
+     * The first 32 bytes of uint256 key in the doubly-linked list of this contract's storage mappings;
```
jsync-swirlds, line 193: Do we mean 32 bits of the uint256?

mhess-swl (approved):
LGTM 😄

services/state/token/account.proto
```diff
@@ -189,29 +189,78 @@ message Account {
...
+    int32 first_storage_key_seventh_int = 38;
+    /**
+     * The eighth 32 bytes of uint256 key in the doubly-linked list of this contract's storage mappings;
+     * It is null if this account is not a contract, or a contract with no storage.
+     */
+    int32 first_storage_key_eighth_int = 39;
+    /*
+     * Number of the low-order bytes in first Uint256 Key that contain ones
+     */
+    bytes first_key_non_zero_bytes = 40;
```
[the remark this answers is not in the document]
Neeharika-Sompalli: yeah I like the idea. I was trying to add these fields [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/merkle/MerkleAccountState.java#L94)

jsync-swirlds (approved):

services/state/token/account.proto
```diff
@@ -189,29 +189,64 @@ message Account {
...
     int64 amount = 2;
 }
+/**
+ * Message representing a uint256 value.
+ */
+message Uint256Value {
+    /**
+     * The number low-order bytes in the 256-bits that contain ones
+     */
+    uint32 non_zero_bits = 1;
```
tinker-michaelj, line 235: Have we adopted a policy on use of the `uint` protobuf variants? (Since in Java we always end up with signed integers, I think at some point Leemon suggested we use strictly `int32` and `int64`.)

Neeharika-Sompalli: Since this message says `Uint256Value` I have used `uint` protobuf types. Will change to int32. Thanks @tinker-michaelj

jsync-swirlds: The non_zero_bits value could be signed, as it's not part of the 256-bit value.
If we used int64 for the segments, however, the bits get shifted around for 2s compliment when the high bit is set, so we can't easily concatenate the bits into a single 256 bit value without some tricky manipulation.
We could use a single bytes field and try to limit the length, but using the uint64 segments lets us interpret the java `long` as unsigned for bit manipulations and keep varint packing efficiency; though we do have to be somewhat careful not to perform regular arithmetic operations on the `long` primitives.

tinker-michaelj (approved):
LGTM!

PR #272, opened 2023-04-21
Update node weight

### LINKED ISSUE #271: Add consensus `weight` to `NodeStake`
Since we will be updating consensus weight of all nodes midnight UTC everyday, we need to expose the weight fields to mirror node using record stream

Neeharika-Sompalli:
Fixes #271 

Related to https://github.com/hashgraph/hedera-services/issues/6204

Adds Consensus Weight to NodeStake to update mirror nodes with the staking info updated midnight UTC

### DIFF
--- a/services/node_stake_update.proto
+++ b/services/node_stake_update.proto
@@ -128,11 +128,13 @@ message NodeStake {
    */
   int64 stake_rewarded = 7;
   /**
-   * The consensus weight of this node in the network. This is computed based on the stake of this node
-   * at midnight UTC of the current day. If the stake of this node is less than minStake, then the
-   * weight is 0. Sum of all weights of nodes in the network should be less than 500.
-   * If the stake of this node A is greater than minStake,
-   * then A's weight is computed as Math.max(1, Math.ceil(A's stake * 500 / total stake of all nodes)).
+   * The consensus weight of this node in the network.
+   * The network normalizes the weights of nodes above minStake so that the total sum of
+   * weight is approximately 500. Specifically, the weight of a node A with stake > minStake
+   * is computed using integer arithmetic as,
+   * A_weight = Math.max((# of whole hbars staked by A) * 500 / (total # of whole hbars staked), 1)
+   * so that every node above minStake has weight at least 1; but any node that has staked
+   * at least 1 out of every 250 whole hbars staked will have weight >= 2.
    */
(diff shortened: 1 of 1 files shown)


services/node_stake_update.proto
```diff
@@ -127,4 +127,12 @@ message NodeStake {
    * beginning of the new staking period.
    */
   int64 stake_rewarded = 7;
+  /**
+   * The consensus weight of this node in the network. This is computed based on the stake of this node
+   * at midnight UTC of the current day. If the stake of this node is less than minStake, then the
+   * weight is 0. Sum of all weights of nodes in the network should be less than 500.
```
tinker-michaelj, line 133: I might tweak this to give a little more context,
```
The network normalizes the weights of nodes above minStake so that the total sum of
weight is approximately 500. Specifically, the weight of a node A with stake > minStake
is computed using integer arithmetic as,

A_weight = Math.max((# of whole hbars staked by A) * 500 / (total # of whole hbars staked), 1)

so that every node above minStake has weight at least 1; but any node that has staked
more than 1 out of every 250 whole hbars staked will have weight >= 2. 
```

tinker-michaelj (dismissed):
One suggestion on comment, otherwise LGTM!

services/node_stake_update.proto
Neeharika-Sompalli: Thanks @tinker-michaelj . Updated

services/node_stake_update.proto
```diff
@@ -112,7 +112,7 @@ message NodeStake {
   int64 reward_rate = 4;
 
   /**
-   * Consensus weight of this node for the new staking period.
+   * Total stake (rewarded and not rewarded) of this node at the beginning of the new staking period.
```
xin-hedera, line 115: it's true only when  min_stake <= stake_rewarded + stake_not_rewarded <= max_stake I believe. i.e. if the sum is less than min_stake, stake is 0; if the sum is more than max_stake, stake = max_stake

Neeharika-Sompalli: Yeah just discussed with Nick and the same field can be used as the `consensus weight` .

Neeharika-Sompalli:
`NodeStake.stake` field will be used as the consensus weight to export value to mirror node. No need for a new field. Closing this PR.

PR #273, opened 2023-05-12
Include `ContractService` state types

tinker-michaelj:
**Description**:
 - Includes protobuf types for contract storage, bytecode, and NFT state.


Neeharika-Sompalli (approved):

services/state/contract/bytecode.proto
```diff
@@ -0,0 +1,44 @@
...
+
+option java_package = "com.hederahashgraph.api.proto.java";
+// <<<pbj.java_package = "com.hedera.hapi.node.state.contract">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+/**
+ * The bytecode for a contract id, along with its hash. (Computing this hash
+ * is fairly expensive, and may be in principle be requested frequently within
+ * an EVM transaction.)
+ */
```
david-bakin-sl, line 33: Yes, expensive to compute, but is there a reason it is part of the message?  If expensive to compute can't the message receiver calculate it and keep it?  Or is it that an _instance_ of this message is _consumed_ multiple times   and so it's cheaper for the sender to compute it once, rather than each receiver once?

services/state/common.proto
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Message representing a uint256 value.
+ */
+message Int256Value {
+    /**
+     * The number low-order bytes in the 256-bits that contain ones
+     */
+    int32 non_zero_bits = 1;
+    /**
+     * The first 64 bits of a int256 value
```
david-bakin-sl, line 38: By "first" you mean big-endian of course?  (But then, why "of course"?)

services/state/contract/storage_slot.proto
```diff
@@ -0,0 +1,70 @@
...
+option java_package = "com.hederahashgraph.api.proto.java";
+// <<<pbj.java_package = "com.hedera.hapi.node.state.contract">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+/**
+ * The key of a storage slot. A slot is scoped to a specific contract number.
+ * 
+ * For each contract, its EVM storage is a mapping of 256-bit keys (or "words") to 256-bit values. 
+ */
+message SlotKey {
```
david-bakin-sl, line 36: I'm curious where these two messages are used.  Two questions come up:  Why isn't there just one message - what is the use case where you need keys without their corresponding slot or vice-versa?  And second, what is the use case where the contract number is associated with a single key, rather than having a single contract's collection of K/V pairs?  IOW, why is contract#  in the same message as slot#?

services/state/contract/storage_slot.proto
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * The number of the contract whose storage this slot belongs to.
+     */
+    int64 contract_number = 1;
+
+    /**
+     * The EVM key of this slot, when left-padded with zeros to form a 256-bit word.
+     */
+    bytes value = 2;
```
david-bakin-sl: So this single thing in one message is called all of "Slot", "key", and "value".  Three names for one thing.  ("Slot" and "key" together in the message name, "slot" in the comment @L43, and "value" is the name of the field.) 

(I get the distinction between the "slot" which holds the value and the "key" which identifies the "slot" - but I think in practice this distinction is glossed over and both the "slot" and the "key" which refers to it are called "slot".  Maybe not though.)

I guess minimum fix - and acceptable - would just be to change "value" -> "key"?

services/state/common.proto
```diff
@@ -0,0 +1,60 @@
...
+     */
+    int64 first_long = 2;
```
david-bakin-sl, line 40: I get that `uint64` types don't work well in Java, but having these be `int64` types is awkward too because representing each quarter of the `int256` (and why `int256` and not `uint256`?) with `int64` may lead to each one having random looking signs, and various confusions when writing/reading code (not to mention looking at protobufs themselves). I guess `first_long` is so you know how many of the following named fields to expect - but then why not just represent as `bytes` to be left-padded by user to 32 bytes just like for `SlotKey.value` below?  Could use either left-padding with a field to indicate whether to pad with `0` or `0xff` bytes, or could just say you need to sign-extend it to left.

david-bakin-sl (commented):
Questions about design, mainly.

services/state/token/nft.proto
```diff
@@ -0,0 +1,87 @@
...
+
+option java_package = "com.hederahashgraph.api.proto.java";
+// <<<pbj.java_package = "com.hedera.hapi.node.state.token">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+/**
+ * First-draft representation of a Hedera Token Service NFT in the network Merkle tree.
+ */
+message Nft {
+    message Id {
```
Neeharika-Sompalli, line 35: Should we have `creation_time` as well as specied in `UniqueToken`?

services/state/contract/storage_slot.proto
tinker-michaelj: We want to be able to use a single `VirtualMap` for the storage of all contracts in state; so the key has to be contract-scoped.

services/state/contract/bytecode.proto
tinker-michaelj: Removed this field.

services/state/common.proto
tinker-michaelj: Removed this type.

services/state/contract/storage_slot.proto
tinker-michaelj: Renamed this.

david-bakin-sl (approved):
LGTM

PR #280, opened 2023-06-15
add schedule proto

### LINKED ISSUE #279: Add Schedule Protobuf

povolev15:
new schedule proto for scheduler for modularization

Fixes #279

### DIFF
--- a/services/state/schedule/schedule.proto
+++ b/services/state/schedule/schedule.proto
@@ -1,0 +1,112 @@
+syntax = "proto3";
+
+package proto;
+
+/*-
+ * ‌
+ * Hedera Network Services Protobuf
+ * ​
+ * Copyright (C) 2018 - 2023 Hedera Hashgraph, LLC
+ * ​
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ‍
+ */
+
+import "basic_types.proto";
+import "timestamp.proto";
+
+option java_package = "com.hederahashgraph.api.proto.java";
+// <<<pbj.java_package = "com.hedera.hapi.node.state.file">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+/**
+ * First-draft representation of a Hedera Schedule Service in the network Merkle tree.
+ *
+ * As with all network entities, a schedule has a unique entity number, which is usually given along
+ * with the network's shard and realm in the form of a shard.realm.number id.
+ */
+message Schedule {
+
+    /**
+     * The schedule deleted flag
+     */
+    bool deleted = 1;
+
+    /**
+     * The schedule executed flag
+     */
+    bool executed = 2;
+
+    /**
+     * The memo associated with this schedule.
+     */
+    string memo = 3;
+
+    /**
+     * The file's unique entity number in the Merkle state.
+     */
+    ScheduleID id = 4;
+
+    /**
+     * The schedule account for this schedule.
+     */
+    AccountID scheduler_account = 5;
+
+    /**
+     * The payer account for this schedule.
+     */
+    AccountID payer_account = 6;
+
+    /**
+      * The admin key for this schedule.
+     */
+    Key adminKey = 7;
+
+    /**
+      * The start time of the schedule.
+     */
+    TimeStamp schedule_valid_start = 8;
+
+    /**
+      * The expiration time of the schedule as provided by the user.
+     */
+    TimeStamp expiration_time_provided = 9;
+    /**
+      * The calculated expiration time of the schedule.
+     */
+    TimeStamp calculated_expiration_time = 10;
+
+    /**
+      * The consensus timestamp of the transaction that created this schedule.
+     */
+    TimeStamp resolution_time = 11;
+
+    /**
+     * The scheduled transaction
+     */
+    SchedulableTransaction scheduled_transaction = 12;
+
+    /**
+     * The transaction body bytes
+     */
+    bytes body_bytes = 13;
+
+    /**
+    * All signature keys must sign to create schedule.
+    */
+    repeated Key signatories = 14;
+
+
+
+
+}
--- a/services/state/schedule/schedule.proto
+++ b/services/state/schedule/schedule.proto
@@ -53,7 +53,7 @@ message Schedule {
     string memo = 3;
 
     /**
-     * The file's unique entity number in the Merkle state.
+     * This schedule's unique ID within the network state.
      */
     ScheduleID id = 4;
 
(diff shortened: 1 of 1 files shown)


services/state/schedule/schedule.proto
```diff
@@ -0,0 +1,112 @@
...
+     */
+    bool executed = 2;
+
+    /**
+     * The memo associated with this schedule.
+     */
+    string memo = 3;
+
+    /**
+     * The file's unique entity number in the Merkle state.
```
jsync-swirlds, line 56:
```suggestion
     * This schedule's unique ID within the network state.
```

services/state/schedule/schedule.proto
```diff
@@ -0,0 +1,112 @@
...
+    /**
+      * The start time of the schedule.
+     */
+    TimeStamp schedule_valid_start = 8;
+
+    /**
+      * The expiration time of the schedule as provided by the user.
+     */
+    TimeStamp expiration_time_provided = 9;
+    /**
```
jsync-swirlds, line 84: newline needed here.

services/state/schedule/schedule.proto
```diff
@@ -0,0 +1,112 @@
...
+      * The calculated expiration time of the schedule.
+     */
+    TimeStamp calculated_expiration_time = 10;
+
+    /**
+      * The consensus timestamp of the transaction that created this schedule.
```
jsync-swirlds, line 90:
```suggestion
      * The consensus timestamp of the transaction that executed, deleted, or expired this schedule.
```

services/state/schedule/schedule.proto
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * The memo associated with this schedule.
+     */
+    string memo = 3;
+
+    /**
+     * The file's unique entity number in the Merkle state.
+     */
+    ScheduleID id = 4;
```
Neeharika-Sompalli, line 58:
```suggestion
    ScheduleID schedule_id = 4;
```

services/state/schedule/schedule.proto
```diff
@@ -0,0 +1,112 @@
...
+     */
+    SchedulableTransaction scheduled_transaction = 12;
+
+    /**
+     * The transaction body bytes
+     */
+    bytes body_bytes = 13;
+
+    /**
+    * All signature keys must sign to create schedule.
```
jsync-swirlds, line 105:
```suggestion
    * All the primitive keys that have already signed this schedule.  The scheduled transaction will not be executed before this list is sufficient to "activate" the required keys for the scheduled transaction.
```

jsync-swirlds:
Thanks for putting this in for me, @povolev15, it helps a lot.

services/state/schedule/schedule.proto
povolev15: done

services/state/schedule/schedule.proto
povolev15: done

services/state/schedule/schedule.proto
povolev15: done

povolev15:
> Thanks for putting this in for me, @povolev15, it helps a lot. Also, sorry about the close/reopen, github's inconsistent UI always has me clicking the wrong place.

always welcome.

services/state/schedule/schedule.proto
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * The memo associated with this schedule.
+     */
+    string memo = 3;
+
+    /**
+     * This schedule's unique ID within the network state.
+     */
+    ScheduleID id = 4;
```
Neeharika-Sompalli, line 58: To follow same pattern as other PBJ types
```suggestion
    ScheduleID schedule_id= 4;
```

services/state/schedule/schedule.proto
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * The payer account for this schedule.
+     */
+    AccountID payer_account = 6;
+
+    /**
+      * The admin key for this schedule.
+     */
+    Key adminKey = 7;
```
Neeharika-Sompalli, line 73:
```suggestion
    Key admin_key = 7;
```

services/state/schedule/schedule.proto
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * All the primitive keys that have already signed this schedule.
+     * The scheduled transaction will not be executed before this list is
+     * sufficient to "activate" the required keys for the scheduled transaction.
+     */
+    repeated Key signatories = 14;
+
+
+
+
```
Neeharika-Sompalli, line 114: nit : can remove these lines

services/state/schedule/schedule.proto
```diff
@@ -0,0 +1,115 @@
...
+    /**
+      * The admin key for this schedule.
+     */
+    Key adminKey = 7;
+
+    /**
+      * The start time of the schedule.
+     */
+    TimeStamp schedule_valid_start = 8;
```
Neeharika-Sompalli, line 78: Why do we need this field also to be serialized/deserialized when we have parent txn body_bytes already.

services/state/schedule/schedule.proto
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+      * The expiration time of the schedule as provided by the user.
+     */
+    TimeStamp expiration_time_provided = 9;
+
```
Neeharika-Sompalli, line 84: same for this. It seems we can get this field, when we parse body_bytes

services/state/schedule/schedule.proto
jsync-swirlds: body bytes are the child transaction (and not sure we need to keep it).

jsync-swirlds:
Github is broken, so trying here.
Most of the other similar types seem to use just id. We only repeat the parent type for ID when using the int64 number (which we need to fix) instead of the proper ID value.

services/state/schedule/schedule.proto
jsync-swirlds: Also here, we aren't storing the parent (I don't think) in body bytes, and might remove it.

services/state/schedule/schedule.proto
jsync-swirlds: Most of the other similar types seem to use just id. We only repeat the parent type for ID when using the int64 number (which we need to fix) instead of the proper ID value.

services/state/schedule/schedule.proto
Neeharika-Sompalli: I was looking at all the fields that are being serialized in `ScheduleVirtualValue` and seems there are more in this protobuf.

Neeharika-Sompalli (approved):
LGTM as a first draft ! But, I don't think all the fields added are needed since `ScheduleVirtualValue ` doesn't serialize/deserialize all the fields. Will defer to @jsync-swirlds

services/state/schedule/schedule.proto
jsync-swirlds: I see, I mostly looked at use rather than what's serialized.  Currently the code re-parses a lot, and that's costly.  One of the goals for this object is to avoid that where it makes sense.  Ideally we'll store less data in total, and avoid parsing the raw data repeatedly.

services/state/schedule/schedule.proto
Neeharika-Sompalli: I think we should add only the fields we are serializing in `ScheduleVirtualValue` [here](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/hedera-mono-service/src/main/java/com/hedera/node/app/service/mono/state/virtual/schedule/ScheduleVirtualValue.java#L321)

jsync-swirlds: I believe this object should be more efficient than ScheduleVirtualValue.
Ideally, we won't store the body bytes and will just store the values needed for processing directly.  That should produce an overall smaller object and also avoid double parsing cost.

tinker-michaelj (approved):
All seems very sensible to me. 👍

PR #285, opened 2023-06-21
Update file.proto to use FileID type instead of long for file identification

kimbor:
Change `long file_number` to `FileID file_id` in file.proto

**Related issue(s)**:

This is in support of [issue #7098 in the hedera-services repo](https://github.com/hashgraph/hedera-services/issues/7098)

**Notes for reviewer**:

`file.proto` is not part of any previous releases. Therefore there is no need to ensure backwards compatibility.

### DIFF
--- a/services/state/consensus/topic.proto
+++ b/services/state/consensus/topic.proto
@@ -51,9 +51,9 @@ option java_multiple_files = true;
  */
 message Topic {
     /**
-     * The topic's unique entity number in the Merkle state.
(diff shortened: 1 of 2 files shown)


iwsimon (approved):
LGTM

jsync-swirlds:
Note, Issue [7207](https://github.com/hashgraph/hedera-services/pull/7207) must merge before this.

rbair23 (approved):

PR #286, opened 2023-06-21
change topic number to topicID

povolev15:
change topic number to topicID in protobuf

https://github.com/hashgraph/hedera-services/issues/7104

### DIFF
--- a/services/state/consensus/topic.proto
+++ b/services/state/consensus/topic.proto
@@ -53,7 +53,7 @@ message Topic {
     /**
      * The topic's unique entity number in the Merkle state.
      */
-    int64 topic_number = 1;
+    TopicID id = 1;
     /**
      * The number of messages sent to the topic.
      */
--- a/services/state/consensus/topic.proto
+++ b/services/state/consensus/topic.proto
@@ -51,7 +51,7 @@ option java_multiple_files = true;
  */
 message Topic {
     /**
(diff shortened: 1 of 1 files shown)


kimbor (commented):
It should be:
TopicID topic_id
in order to follow the naming convention that other places in hedera-protobufs follows.
Also, you should update the javadoc to say id instead of number.

jsync-swirlds:
There are exactly two ID values in state prior to this PR, both are `id`.
The ID values in the public transaction protobufs often refer to an object other than the current, or are expanded to be clearer when translated to other languages (e.g. typescript where `id` can be a problem).
We should probably discuss in the daily team call which way (`name.name_id` or `name.id`) we want to go in the states objects.

services/state/consensus/topic.proto
```diff
@@ -53,7 +53,7 @@ message Topic {
     /**
      * The topic's unique entity number in the Merkle state.
```
jsync-swirlds, line 54: `s/number/id/`
Note, this PR cannot merge before the services PR that mirrors it.

iwsimon (approved):
LGTM

services/state/consensus/topic.proto
povolev15: this one https://github.com/hashgraph/hedera-services/pull/7216

povolev15:
restored the branch that it will not break develop

PR #288, opened 2023-06-27
Adjusting the state Schedule message to improve implementation.

jsync-swirlds:
* Changed body_bytes to original_create_transaction
   * Changed type to TransactionBody
 * Updated comments to clarify several items and correct language

Verified using [verify-protos](https://github.com/jsync-swirlds/public-scripts/blob/addVerifyProtos/verify-protos)

### DIFF
--- a/services/state/schedule/schedule.proto
+++ b/services/state/schedule/schedule.proto
@@ -25,13 +25,14 @@ package proto;
 import "basic_types.proto";
 import "timestamp.proto";
 import "schedulable_transaction_body.proto";
+import "transaction_body.proto";
 
 option java_package = "com.hederahashgraph.api.proto.java";
 // <<<pbj.java_package = "com.hedera.hapi.node.state.schedule">>> This comment is special code for setting PBJ Compiler java package
 option java_multiple_files = true;
 
 /**
- * First-draft representation of a Hedera Schedule Service in the network Merkle tree.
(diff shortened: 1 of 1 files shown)


services/state/schedule/schedule.proto
```diff
@@ -54,54 +57,64 @@ message Schedule {
...
     SchedulableTransactionBody scheduled_transaction = 12;
 
     /**
-     * The transaction body bytes
+     * The full transaction that created this schedule.  This is primarily used for duplicate
+     * schedule create detection.  This is also the source of the parent transaction ID, from
+     * which the child transaction ID is derived.
      */
-    bytes body_bytes = 13;
+    TransactionBody original_create_transaction = 13;
```
mhess-swl, line 117: This is probably a dumb question, but is it ok to change the type on this if we need to be backwards compatible?

jsync-swirlds: The only code that uses this is on my branch and not merged, and nothing has stored any of these objects in state, so we don't need backwards compatibility yet.
Once that code merges and there is state with these values, we will have to be immensely more careful, hence the work to make these as correct as possible at the start.

mhess-swl: Sounds good, thanks for clarifying!

mhess-swl (approved):

PR #290, opened 2023-06-29
Update common.proto, token.proto, token-relation.protp and classes that rely on it to use AccountID and TokenID

iwsimon:
**Description**:

DO NOT DELETE 7242-update-common-ids BRANCH YET

Update common.proto and classes that rely on it to use AccountID and TokenID instead of long.
Update token.proto and classes that rely on it to use AccountID and TokenID instead of long.
Update token_relation.proto and classes that rely on it to use AccountID and TokenID instead of long

**Related issue(s)**:

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/services/state/token/token.proto
+++ b/services/state/token/token.proto
@@ -38,9 +38,9 @@ option java_multiple_files = true;
 
 message Token {
     /**
-     * The unique entity number of this token.
+     * The unique entity id of this token.
      */
-    int64 token_number = 1;
+    TokenID token_id = 1;
     /**
      * The human-readable name of this token and its not necessarily unique. Maximum length allowed is 100 bytes.
      */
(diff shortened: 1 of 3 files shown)


Neeharika-Sompalli (approved):

services/state/token/token.proto
```diff
@@ -38,9 +38,9 @@ option java_multiple_files = true;
 
 message Token {
     /**
-     * The unique entity number of this token.
+     * The unique entity id of this token.
      */
-    int64 token_number = 1;
+    TokenID token_id = 1;
```
jsync-swirlds, line 43: Nit: We've been a bit inconsistent with these.
Some services have just `id`, others have `thing_id`.
I would argue that typing `Token.tokenId()` is redundant and these should be just `id`/`Token.id()`.
Might be wise to bring up in the modularization call so it can be consistent.

iwsimon: Yes, agreed. we see id in topic.proto, but file_id in file.proto, token_id in token.proto, etc. There are so many related  code changes for just this small name change.

jsync-swirlds (approved):

PR #292, opened 2023-07-05
Use NftID instead of UniqueTokenId

kimbor:
**Description**:
1) Move NftID from token_get_nft_info.proto to basic_types.proto
2) Change everywhere that uses UniqueTokenID to use NftID instead
3) Delete UniqueTokenId from common.proto

**Related issue(s)**:
Fixes [#7346 in hedera-services](https://github.com/hashgraph/hedera-services/issues/7346)

**Note**:
DO NOT MERGE UNTIL AFTER [PR # 7409](https://github.com/hashgraph/hedera-services/pull/7409) has been merged

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -114,6 +114,21 @@ message AccountID {
 }
 
 /**
+ * Represents an NFT on the Ledger
+ */
+message NftID {
+    /**
+     * The (non-fungible) token of which this NFT is an instance
+     */
+    TokenID tokenID = 1;
+
+    /**
+     * The unique identifier of this instance
+     */
+    int64 serialNumber = 2;
+}
+
+/**
  * The ID for a file 
  */
 message FileID {
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -114,18 +114,18 @@ message AccountID {
 }
 
 /**
- * Represents an NFT on the Ledger
+ * Identifier for a unique token (or "NFT"), used by both contract and token services.
  */
 message NftID {
     /**
      * The (non-fungible) token of which this NFT is an instance
      */
-    TokenID tokenID = 1;
+    TokenID token_id = 1;
 
     /**
-     * The unique identifier of this instance
+     * The serial number of this NFT within its token type
      */
-    int64 serialNumber = 2;
+    int64 serial_number = 2;
 }
 
 /**
--- a/services/state/token/account.proto
+++ b/services/state/token/account.proto
@@ -97,11 +97,11 @@ message Account {
      */
     int64 head_token_number = 13;
     /**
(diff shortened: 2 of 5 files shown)


services/state/token/account.proto
```diff
@@ -97,11 +97,11 @@ message Account {
      */
     int64 head_token_number = 13;
     /**
-     * The NftId of the head of the linked list from unique tokens map for the account.
+     * The NftID of the head of the linked list from unique tokens map for the account.
      */
     int64 head_nft_id = 14;
```
iwsimon, line 102: Should this be NftID head_nft_id=14?

kimbor: That change will be made as part of issue #7243

iwsimon: I see. Thanks.

services/basic_types.proto
```diff
@@ -113,6 +113,21 @@ message AccountID {
...
 }
 
+/**
+ * Represents an NFT on the Ledger
+ */
+message NftID {
+    /**
+     * The (non-fungible) token of which this NFT is an instance
+     */
+    TokenID tokenID = 1;
```
Neeharika-Sompalli, line 123: should be camel case token_id

services/basic_types.proto
```diff
@@ -113,6 +113,21 @@ message AccountID {
...
+
+    /**
+     * The unique identifier of this instance
+     */
+    int64 serialNumber = 2;
```
Neeharika-Sompalli, line 128: same here serial_number

services/basic_types.proto
```diff
@@ -113,6 +113,21 @@ message AccountID {
     }
 }
 
+/**
+ * Represents an NFT on the Ledger
+ */
```
Neeharika-Sompalli, line 118: Docs can be taken from the removed `UniqueTokenId` . They seem to be detailed

services/basic_types.proto
kimbor: Almost all of the names in basic_types.proto are in camel case, but a few are snake case. Do we want to standardize on snake case? If so, there are a lot more places that should be changed.

Neeharika-Sompalli: yeah, I think old ones are in camel case. But then we standardized to snake case for any further changes. It should not affect generated code afaik.

kimbor: I fixed it here. I also created issue #7423 to fix it other places.

services/basic_types.proto
kimbor: Fixed

services/basic_types.proto
kimbor: Fixed

Neeharika-Sompalli (approved):

PR #294, opened 2023-07-11
use snake case throughout protobufs

iwsimon:
**Description**:

updated *.proto to use *_number through all the services/state
Standardize to use snake case throughout all the *.proto files under services\state
Added staked_id, removed staked_number in account.proto

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

Related to hedera-services issue 7423, 7343.

**Checklist**

### DIFF
--- a/services/state/token/account.proto
+++ b/services/state/token/account.proto
@@ -78,12 +78,24 @@ message Account {
      * be set to the time when the account starts staking to a node.
      */
     int64 stake_period_start = 9;
+
     /**
-     * The node number or the account number this account is staked to.
-     * It is negative if staking to a node and positive if staking to an account and 0 if not staking to anyone.
-     * When staking to a node, it is stored as -node-1 to differentiate node 0.
+     * ID of the account or node to which this account is staking.
      */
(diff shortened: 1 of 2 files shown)


services/state/token/account.proto
```diff
@@ -78,108 +78,120 @@ message Account {
...
+         * ID of the new node this account is staked to. If set to the sentinel <tt>-1</tt>, this field
+         * removes this account's staked node ID.
+         */
+        int64 staked_node_id = 11;
+    }
     /**
      * A boolean marking if the account declines rewards.
      */
-    bool decline_reward = 11;
+    bool decline_reward = 12;
```
jsync-swirlds, line 102: renumbering is one of the few things that completely breaks protobuf.  You can do this, if there is no data, anywhere, actually using the protobuf prior to this change.  If not, then you must leave the old staked number and mark it removed, then add the new staked_id at the end of the protobuf definition.

kimbor (approved):

jsync-swirlds (approved):

PR #295, opened 2023-07-14
Add Staking Reward related fields from `MerkleNetworkContext`

Neeharika-Sompalli:
Related to https://github.com/hashgraph/hedera-services/issues/7187
Add Staking Reward related fields from `MerkleNetworkContext` to a singleton state in Token Service

### DIFF
--- a/services/state/token/staking_rewards.proto
+++ b/services/state/token/staking_rewards.proto
@@ -41,13 +41,13 @@ message StakingRewards {
    */
   bool staking_rewards_activated = 1;
   /**
-  * Total of (balance + stakedToMe) for all accounts staked to all nodes with declineReward=true, at the
+  * Total of (balance + stakedToMe) for all accounts staked to all nodes with declineReward=false, at the
   * beginning of the new staking period.
   */
   int64 total_staked_reward_start = 2;
   /**
-  * Total of (balance + stakedToMe) for all accounts staked to this node with declineReward=false, at the
-  * beginning of the new staking period.
+  * Total of (balance + stakedToMe) for all accounts staked to this node, at the beginning of the new
+  * staking period.
   */
   int64 total_staked_start = 3;
   /**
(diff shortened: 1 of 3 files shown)


services/state/token/staking_rewards.proto
```diff
@@ -0,0 +1,57 @@
...
+ * As with all network entities, staking info is per node and has an unique entity number represented as shard.realm.X.
+ */
+message StakingRewards {
+  /**
+   * Whether staking rewards are activated for this node. This is set to true when the balance of 0.0.800
+   * reaches minimum required balance.
+   */
+  bool staking_rewards_activated = 1;
+  /**
+  * Total of (balance + stakedToMe) for all accounts staked to all nodes with declineReward=true, at the
```
tinker-michaelj, line 44: `declineReward=false` not `true`, right?

Neeharika-Sompalli: Ah yes, changed.

services/state/token/staking_rewards.proto
```diff
@@ -0,0 +1,57 @@
...
+  * beginning of the new staking period.
+  */
+  int64 total_staked_reward_start = 2;
+  /**
+  * Total of (balance + stakedToMe) for all accounts staked to this node, at the beginning of the new
+  * staking period.
+  */
+  int64 total_staked_start = 3;
+  /**
+   * The total amount in tinybars that will be received in the next reward situation.
```
tinker-michaelj, line 54: I think "reward situation" is an account-specific notion; e.g., "This transaction is a reward situation for Alice but not Bob." For this field, maybe:
```
The total staking rewards in tinybars that COULD be collected by all accounts after the end of this staking period;
assuming that no account "renounces" its rewards by, for example, setting declineReward=true.
```

Neeharika-Sompalli: Yes, makes sense. Updated name of file and docs

tinker-michaelj (approved):
LGTM!

PR #297, opened 2023-07-17
Update Schedule state protobuf to add a wait_for_expiry flag.

jsync-swirlds:
* This protobuf is not yet used, so the fields were renumbered for efficient representation.
 * This is in support of resolving Issue #5518

### DIFF
--- a/services/state/schedule/schedule.proto
+++ b/services/state/schedule/schedule.proto
@@ -52,43 +52,55 @@ message Schedule {
     bool executed = 2;
 
     /**
+     * The schedule flag to wait for expiration
+     * A schedule will be executed immediately when all necessary signatures are gathered, unless
+     * this flag is set.  If this flag is set, the schedule will wait until the consensus time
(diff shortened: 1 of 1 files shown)


iwsimon (approved):
LGTM

services/state/schedule/schedule.proto
```diff
@@ -51,44 +51,56 @@ message Schedule {
...
+     * The schedule flag to wait for expiration
+     * A schedule will be executed immediately when all necessary signatures are gathered, unless
+     * this flag is set.  If this flag is set, the schedule will wait until the consensus time
+     * reaches expiration_time_provided, when signatures will again be verified, and if all
+     * required signatures are present at that time, the schedule will be executed.  Otherwise
+     * the schedule will expire without execution.
+     * Note that a schedule is always removed from state when it expires, regardless of whether it
+     * was executed or not.
+     */
+    bool wait_for_expiry = 3;
```
mhess-swl, line 64: Don't we want this at the end, and not change all of the other numbers for backwards compatibility?

PR #298, opened 2023-08-01
protobuf cleanup

kimbor:
**Description**:
Clean up files in services/state for consistency in naming
-Rename expiration_time -> expiration_second in several fields where it occurs
-Rename num -> number in several places
-Rename auto_renew_secs -> auto_renew_seconds
-Rename some id fields. All id fields are now of the form [message name]_id
-Remove unused imports
-Edits to comments

**Related issue(s)**:

Fixes #7684 in hedera-services

**Notes for reviewer**:
DO NOT MERGE until after [PR # 7986](https://github.com/hashgraph/hedera-services/pull/7986) in hedera-services is merged

### DIFF
--- a/services/state/blockrecords/block_info.proto
+++ b/services/state/blockrecords/block_info.proto
@@ -39,12 +39,12 @@ message BlockInfo {
   /**
    * The consensus time of the first transaction of the last block, this is the last completed immutable block.
    */
-  Timestamp first_cons_time_of_last_block = 2;
+  Timestamp first_cons_time_of_last_block = 3;
   /**
    * SHA384 48 byte hashes of the last 256 blocks in single byte array.
    * First 48 bytes is the oldest block.
    * Last 48 bytes is the newest block, which is the last fully completed immutable block.
    * If we are shortly after genesis and there are less than 256 blocks then this could contain less than 256 hashes.
    */
-  bytes block_hashes = 3;
+  bytes block_hashes = 2;
 }
--- a/services/state/blockrecords/block_info.proto
+++ b/services/state/blockrecords/block_info.proto
@@ -39,12 +39,12 @@ message BlockInfo {
   /**
    * The consensus time of the first transaction of the last block, this is the last completed immutable block.
    */
-  Timestamp first_cons_time_of_last_block = 3;
+  Timestamp first_cons_time_of_last_block = 2;
   /**
    * SHA384 48 byte hashes of the last 256 blocks in single byte array.
    * First 48 bytes is the oldest block.
    * Last 48 bytes is the newest block, which is the last fully completed immutable block.
    * If we are shortly after genesis and there are less than 256 blocks then this could contain less than 256 hashes.
    */
-  bytes block_hashes = 2;
+  bytes block_hashes = 3;
 }
--- a/services/state/consensus/topic.proto
+++ b/services/state/consensus/topic.proto
@@ -61,7 +61,7 @@ message Topic {
     /**
      * The topic's consensus expiration time in seconds since the epoch.
      */
-    int64 expiry = 3;
+    int64 expiration_seconds = 3;
     /**
      * The number of seconds for which the topic will be automatically renewed 
      * upon expiring (if it has a valid auto-renew account).
--- a/services/state/consensus/topic.proto
+++ b/services/state/consensus/topic.proto
@@ -29,7 +29,7 @@ option java_package = "com.hederahashgraph.api.proto.java";
 option java_multiple_files = true;
 
 /**
- * First-draft representation of a Hedera Consensus Service topic in the network Merkle tree. 
+ * Representation of a Hedera Consensus Service topic in the network Merkle tree.
  * 
  * As with all network entities, a topic has a unique entity number, which is usually given along 
  * with the network's shard and realm in the form of a shard.realm.number id.
@@ -51,9 +51,9 @@ option java_multiple_files = true;
  */
 message Topic {
     /**
-     * The topic's unique id number in the Merkle state.
+     * The topic's unique id in the Merkle state.
      */
-    TopicID id = 1;
+    TopicID topic_id = 7;
     /**
      * The number of messages sent to the topic.
      */
@@ -71,11 +71,11 @@ message Topic {
      * The id of the account (if any) that the network will attempt to charge for the
      * topic's auto-renewal upon expiration.
      */
-    AccountID auto_renew_account_id = 5;
+    AccountID auto_renew_account_id = 8;
     /**
      * Whether this topic is deleted.
      */
-    bool deleted = 6;
+    bool deleted = 1;
     /**
      * When a topic is created, its running hash is initialized to 48 bytes of binary zeros.
      * For each submitted message, the topic's running hash is then updated to the output
@@ -84,11 +84,11 @@ message Topic {
      * See the TransactionReceipt.proto documentation for an exact description of the
      * data included in the SHA-384 digest used for the update.
      */
-    bytes running_hash = 7;
+    bytes running_hash = 6;
     /**
      * An optional description of the topic with UTF-8 encoding up to 100 bytes.
      */
-    string memo = 8;
+    string memo = 5;
     /**
      * If present, enforces access control for updating or deleting the topic.
      * A topic without an admin key is immutable.
--- a/services/state/consensus/topic.proto
+++ b/services/state/consensus/topic.proto
@@ -61,7 +61,7 @@ message Topic {
     /**
      * The topic's consensus expiration time in seconds since the epoch.
      */
-    int64 expiration_seconds = 3;
+    int64 expiration_second = 3;
     /**
      * The number of seconds for which the topic will be automatically renewed 
      * upon expiring (if it has a valid auto-renew account).
--- a/services/state/consensus/topic.proto
+++ b/services/state/consensus/topic.proto
@@ -53,7 +53,7 @@ message Topic {
     /**
      * The topic's unique id in the Merkle state.
      */
-    TopicID topic_id = 7;
+    TopicID topic_id = 1;
     /**
      * The number of messages sent to the topic.
      */
@@ -71,11 +71,11 @@ message Topic {
      * The id of the account (if any) that the network will attempt to charge for the
      * topic's auto-renewal upon expiration.
      */
-    AccountID auto_renew_account_id = 8;
+    AccountID auto_renew_account_id = 5;
     /**
      * Whether this topic is deleted.
      */
-    bool deleted = 1;
+    bool deleted = 6;
     /**
      * When a topic is created, its running hash is initialized to 48 bytes of binary zeros.
      * For each submitted message, the topic's running hash is then updated to the output
@@ -84,11 +84,11 @@ message Topic {
      * See the TransactionReceipt.proto documentation for an exact description of the
      * data included in the SHA-384 digest used for the update.
      */
-    bytes running_hash = 6;
+    bytes running_hash = 7;
     /**
      * An optional description of the topic with UTF-8 encoding up to 100 bytes.
      */
-    string memo = 5;
+    string memo = 8;
     /**
      * If present, enforces access control for updating or deleting the topic.
      * A topic without an admin key is immutable.
--- a/services/state/file/file.proto
+++ b/services/state/file/file.proto
@@ -23,7 +23,6 @@ package proto;
  */
 
 import "basic_types.proto";
-import "timestamp.proto";
 
 option java_package = "com.hederahashgraph.api.proto.java";
 // <<<pbj.java_package = "com.hedera.hapi.node.state.file">>> This comment is special code for setting PBJ Compiler java package
--- a/services/state/file/file.proto
+++ b/services/state/file/file.proto
@@ -42,7 +42,7 @@ message File {
     /**
       * The file's consensus expiration time in seconds since the epoch.
       */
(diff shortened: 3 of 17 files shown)


services/state/blockrecords/block_info.proto
```diff
@@ -39,12 +39,12 @@ message BlockInfo {
...
-  Timestamp first_cons_time_of_last_block = 2;
+  Timestamp first_cons_time_of_last_block = 3;
   /**
    * SHA384 48 byte hashes of the last 256 blocks in single byte array.
    * First 48 bytes is the oldest block.
    * Last 48 bytes is the newest block, which is the last fully completed immutable block.
    * If we are shortly after genesis and there are less than 256 blocks then this could contain less than 256 hashes.
    */
-  bytes block_hashes = 3;
+  bytes block_hashes = 2;
```
jsync-swirlds, line 49: When reordering, while not _necessary_, it is helpful to actually move the fields within the file.

services/state/schedule/schedule.proto
```diff
@@ -100,7 +100,7 @@ message Schedule {
      * The requested expiration time of the schedule as provided by the user.
      * The actual calculated expiration time may be "earlier" than this, but will not be later.
      */
-    Timestamp expiration_time_provided = 10;
+    Timestamp expiry_provided = 10;
```
jsync-swirlds, line 103: There is no good reason to shorten this, it's clearer _and more accurate_ spelled out (particularly as it *is* an expiration time, not an expiry).
Same applies to the calculated value.

services/state/schedule/schedule.proto
```diff
@@ -71,19 +71,19 @@ message Schedule {
...
      */
-    AccountID scheduler_account = 6;
+    AccountID scheduler_account_id = 6;
 
     /**
      * The explicit payer account for the scheduled transaction.
      * This account is added to the accounts that must sign the schedule before it will execute.
      */
-    AccountID payer_account = 7;
+    AccountID payer_account_id = 7;
```
jsync-swirlds, line 86: Why add the unnecessary and redundant `_id` value here (and for scheduler_account above)?

services/state/schedule/schedule.proto
```diff
@@ -71,19 +71,19 @@ message Schedule {
     /**
      * This schedule's unique ID within the global network state.
      */
-    ScheduleID id = 5;
+    ScheduleID schedule_id = 5;
```
jsync-swirlds, line 74: Not a fan of `Schedule.scheduleId()` as a pattern.  `Schedule.id()` reads better in almost all cases.

services/state/schedule/schedule.proto
kimbor: It is called expiry in other places. I was renaming for consistency. I could rename expiry -> expiration_time instead, if you think that's better.

kimbor: Actually it's mostly called expiry outside of services/state, where it can't be renamed.

services/state/schedule/schedule.proto
kimbor: I'm agnostic on which is better, but IMO consistency is what's most important and scheduleId() is what's currently in use in more places.

services/state/blockrecords/block_info.proto
jsync-swirlds: That's reasonable.  It's not necessary, but it still may be helpful to follow up with a reorder PR after this, just to make working with these easier.

rbair23: Speaking of which, what is the criteria for deciding what order to put the fields in?

services/state/schedule/schedule.proto
rbair23: In this case, I think the _id is good because it refers to an AccountID rather than an Account.

services/state/schedule/schedule.proto
rbair23: This is a good question, generally I'd also omit the foo from fooId so it is foo.id() vs. foo.fooId() in the code -- but I'm not sure if it is too late for that. In all our other protobuf I think we have the redundant name.

services/state/blockrecords/block_info.proto
kimbor: Ordering methodology is explained in the issue associated with this PR. https://github.com/hashgraph/hedera-services/issues/7684

services/state/schedule/schedule.proto
jsync-swirlds: This PR is where we can change it for state objects.
There are a few with ID and a few with thingID, so making them consistent at just ID for state objects is doable now, but would be very hard to do later.

david-bakin-sl: I agree with `id` for your _own_ id, `foo_id` for someone else's.  SImiliar to how you'd name primary keys vs foreign keys in a table.

services/state/schedule/schedule.proto
kimbor: Per discussion, I will change _expiry_ to _expiration_seconds_

jsync-swirlds:
Did you accidentally amend an earlier commit from the main branch?

jsync-swirlds (approved):

services/state/blockrecords/block_info.proto
rbair23: Overall I'm skeptical of the reasoning. I like that we're thinking about how to best order things. In the end it won't matter that much because when we add new fields (which we will no doubt do), they will completely break this ordering. We cannot change the field orders without breaking compatibility. So we will add new booleans at the end, for example, which breaks the reason for reordering fields in the first place.

Maybe it makes sense to order based on what is most likely to be set. Future fields will most likely be left at their default value (since there will be many existing objects with no knowledge of it, so the default will be quite prevalent).

kimbor: We could leave gaps to be filled in by possible future values, i.e. reserve 1-8 for booleans, etc.

jsync-swirlds: Later adds will be less efficient, certainly, but starting from a well packed structure saves space regardless, as that portion of the structure on disk will still be efficient, even if we add less efficiently packed elements later.
For very small object it's probably not a big difference, but for things like NFTs and accounts it's enough that it is worth making things as efficiently stored as we reasonably can.

services/state/primitives.proto
```diff
@@ -24,8 +24,8 @@ package proto;
  * type instead.
  */
 
-option java_package = "com.hederahashgraph.api.proto.java.primitive";
-// <<<pbj.java_package = "com.hedera.hapi.node.state.primitive">>> This comment is special code for setting PBJ Compiler java package
+option java_package = "com.hederahashgraph.api.proto.java";
+// <<<pbj.java_package = "com.hedera.hapi.node.state.primitives">>> This comment is special code for setting PBJ Compiler java package
```
jsync-swirlds, line 28: This should be singular, to match the rest of the codebase which doesn't really use plural for package names.

services/state/schedule/schedule.proto
```diff
@@ -100,7 +100,7 @@ message Schedule {
      * The requested expiration time of the schedule as provided by the user.
      * The actual calculated expiration time may be "earlier" than this, but will not be later.
      */
-    Timestamp expiration_time_provided = 10;
+    int64 provided_expiration_seconds = 10;
```
jsync-swirlds, line 103: This should be singular, it is not a duration, it is a point in time.

services/state/schedule/schedule.proto
```diff
@@ -109,7 +109,7 @@ message Schedule {
      * The schedule will be removed from global network state after the network reaches a consensus
      * time greater than or equal to this value.
      */
-    Timestamp calculated_expiration_time = 11;
+    int64 calculated_expiration_seconds = 11;
```
jsync-swirlds, line 112: This should be singular, it is not a duration, it is a point in time.

services/state/token/account.proto
```diff
@@ -30,167 +30,165 @@ option java_package = "com.hederahashgraph.api.proto.java";
...
      * This key will not be set for hollow accounts until the account is finalized.
      * This key should be set on all the accounts, except for immutable accounts (0.0.800 and 0.0.801).
      */
-    Key key = 3;
+    Key key = 26;
     /**
      * The expiration time of the account, in seconds since the epoch.
      */
-    int64 expiry = 4;
+    int64 expiration_seconds = 12;
```
jsync-swirlds, line 57: I believe we agreed to use singular for point-in-time and plural only for durations?

services/state/file/file.proto
```diff
@@ -30,34 +30,34 @@ option java_package = "com.hederahashgraph.api.proto.java";
...
     /**
      * The file's unique file identifier in the Merkle state.
      */
-    FileID file_id = 1;
+    FileID file_id = 5;
     /**
       * The file's consensus expiration time in seconds since the epoch.
       */
-    int64 expiration_time = 2;
+    int64 expiration_seconds = 2;
```
jsync-swirlds, line 46: `expiration_second`?

services/state/consensus/topic.proto
```diff
@@ -51,17 +51,17 @@ option java_multiple_files = true;
...
+    TopicID topic_id = 7;
     /**
      * The number of messages sent to the topic.
      */
     int64 sequence_number = 2;
     /**
      * The topic's consensus expiration time in seconds since the epoch.
      */
-    int64 expiry = 3;
+    int64 expiration_seconds = 3;
```
jsync-swirlds, line 64: `expiration_second`?

services/state/token/token.proto
```diff
@@ -40,123 +40,123 @@ message Token {
...
      * If this is not provided in a allowed range on token creation, the transaction will fail with INVALID_AUTO_RENEWAL_PERIOD.
      * The default values for the minimum period and maximum period are 30 days and 90 days, respectively.
      */
-    int64 auto_renew_secs = 19;
+    int64 auto_renew_secs = 8;
     /**
      * The expiration time of the token, in seconds since the epoch.
      */
-    int64 expiry = 20;
+    int64 expiration_seconds = 9;
```
jsync-swirlds, line 139: `expiration_second`?

services/state/primitives.proto
kimbor: I agree with you, but changing this is outside the scope of this PR. It was changed in PR #302 and is only showing up here because I merged that change in.

jsync-swirlds: It should not have changed in that PR, then.

services/state/token/token.proto
```diff
@@ -136,7 +136,7 @@ message Token {
     /**
      * The expiration time of the token, in seconds since the epoch.
      */
-    int64 expiry = 20;
+    int64 expiration_seconds = 20;
```
jsync-swirlds, line 139: `expiration_second`?

services/state/token/account.proto
```diff
@@ -55,7 +54,7 @@ message Account {
     /**
      * The expiration time of the account, in seconds since the epoch.
      */
-    int64 expiry = 4;
+    int64 expiration_seconds = 4;
```
jsync-swirlds, line 57: `expiration_second`?

services/state/blockrecords/block_info.proto
kimbor: Done

services/state/schedule/schedule.proto
kimbor: Fixed.

services/state/schedule/schedule.proto
kimbor: Fixed.

services/state/schedule/schedule.proto
kimbor: Fixed.

services/state/schedule/schedule.proto
kimbor: Fixed.

services/state/token/account.proto
kimbor: Fixed.

services/state/file/file.proto
kimbor: Fixed.

services/state/consensus/topic.proto
kimbor: Fixed.

services/state/token/account.proto
```diff
@@ -30,167 +30,165 @@ option java_package = "com.hederahashgraph.api.proto.java";
...
-    AccountID auto_renew_account_id = 25;
+    AccountID auto_renew_account_id = 29;
     /**
      * The number of seconds the network should automatically extend the account's expiration by, if the
      * account has a valid auto-renew account, and is not deleted upon expiration.
      * If this is not provided in a allowed range on account creation, the transaction will fail with INVALID_AUTO_RENEWAL_PERIOD.
      * The default values for the minimum period and maximum period are 30 days and 90 days, respectively.
      */
-    int64 auto_renew_secs = 26;
+    int64 auto_renew_secs = 21;
```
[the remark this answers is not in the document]
kimbor: Fixed.

services/state/token/token.proto
kimbor: Fixed.

services/state/token/account.proto
kimbor: Fixed.

jsync-swirlds (approved):

rbair23 (approved):

iwsimon (approved):
LGTM

PR #299, opened 2023-08-03
Add Protobuf messages for primitive values and a list of Schedules.

jsync-swirlds:
* Added primitive.proto which includes ProtoLong, ProtoInteger, ProtoBoolen, ProtoString, and ProtoBytes.
 * Added ScheduleList message to schedule.proto

This is half of the [related issue](https://github.com/hashgraph/hedera-services/issues/7779).  The other half (code changes to use the new protobuf messages and remove Codec implementations) will follow.

### DIFF
--- a/services/state/primitives.proto
+++ b/services/state/primitives.proto
@@ -1,0 +1,65 @@
+syntax = "proto3";
+
+package proto;
+/*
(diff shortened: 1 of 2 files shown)


iwsimon (approved):
LGTM

PR #311, opened 2023-12-08
Add `BlockInfo#first_cons_time_of_current_block`

tinker-michaelj:
**Description**:
 - We must keep the provisional consensus time of the current block in state so reconnecting nodes can deterministically decide when to close the current block and begin the next one.


Neeharika-Sompalli (approved):
LGTM !

kimbor (approved):

PR #313, opened 2024-01-16
Protobuf changes for implementation of metadata hips 646, 657 and 765

thomas-swirlds-labs:
**Description**:

This PR modifies the token_create, token_update 
and creates a new token_nft_update.proto in order to support 
the 3 metadata related HIPs - [646, 657 and 765]

**Related issue(s)**:
https://github.com/hashgraph/hedera-services/issues/10931

**Fixes**:
https://github.com/hashgraph/hedera-services/issues/10932

**Checklist**

### DIFF
--- a/services/token_update.proto
+++ b/services/token_update.proto
@@ -142,4 +142,15 @@ message TokenUpdateTransactionBody {
      * transaction will resolve to TOKEN_HAS_NO_PAUSE_KEY
      */
     Key pause_key = 15;
+
+    /**
+     * Metadata of the created token definition
+     */
+    google.protobuf.BytesValue metadata = 16;
+
+    /**
+     * The key which can change the metadata of a token
+     * (token definition, partition definition, and individual NFTs).
+     */
+    Key metadata_key = 17;
 }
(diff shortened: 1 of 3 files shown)


services/token_update.proto
```diff
@@ -142,4 +142,15 @@ message TokenUpdateTransactionBody {
      * transaction will resolve to TOKEN_HAS_NO_PAUSE_KEY
      */
     Key pause_key = 15;
+
+    /**
+     * Metadata of the created token definition
+     */
+    google.protobuf.BytesValue metadata = 16;
```
kimbor, line 149: Why is this google.protobuf.BytesValue instead of bytes?

thomas-swirlds-labs: I believe google.protobuf.BytesValue was chosen for metadata in order to make this field optional and easier to extend this transaction type later

kimbor: It's `bytes` in token_create but `google.protobuf.BytesValue` in token_update. I'm not sure which is more appropriate but at a minimum they should match.

Neeharika-Sompalli: We have always used wrappers in `*Update` transaction bodies to differentiate between the value being set or not.  Just like `StringValue` for memo in `token_update` and `string`  in token_create.

PR #321, opened 2024-01-18
Use contract id in slot key

kimbor:
**Description**:
Change long to ContractID in the SlotKey message. This is necessary to support sharding in the future.

**Related issue(s)**:
This PR + https://github.com/hashgraph/hedera-services/pull/10958 together fix https://github.com/hashgraph/hedera-services/issues/10746


iwsimon (approved):
LGTM

PR #323, opened 2024-01-26
feat: Add `Comparable` to State Messages

jsync-swirlds:
* Added the PBJ comparable comment to all non-enum basic types that do not have repeated fields
   * Repeated fields are not currently supported, and enum types are `Comparable` by default
   * Also removed a lot of random non-semantic trailing spaces
* Moved TokenID definition earlier in the file to work around an order-dependency issue in PBJ compiler.
   * This stll fails, but only in the CI environment.  Objects _in basic_types_ that use TokenID are not comparable for now.
* Added the PBJ comparable comment to timestamp, duration, and custom_fees
* Added the PBJ comparable comment to state objects where possible
   * Anything with a Key type field cannot be `Comparable` due to KeyList having a repeated field
   * Anything with a repeated field cannot be `Comparable` currently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,14 @@ must be paid in ℏ from a cryptocurrency account. The payer authorizes a
 fee by signing an appropriate transaction with a sufficient subset of the 
 Ed25519 key(s) associated to their account.
 
+## Overview of state
+State directory and its subdirectories contain the protobuf files that define the state of the network.
+The state is divided into the following subdirectories, based on the service modules:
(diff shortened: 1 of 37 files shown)


kimbor (dismissed):
I assume there will be a follow-on PR in services to remove the one-off comparators in HapiUtils?

jsync-swirlds:
Stale.  Will be redone when/if PBJ comparable support is fixed.

PR #325, opened 2024-02-02
HIP 796: Update TokenLock, TokenUnlock TransactionBody Definitions

thomas-swirlds-labs:
**Description**:

Change the definitions of TokenLockTransactionBody and TokenUnlockTransactionBody
to allow multiple NFTs to be locked/unlocked with a single call

The format of the new transactions is more aligned with the rest of Token Service.

**Related issue(s)**:
https://github.com/hashgraph/hedera-services/issues/9636

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -115,8 +115,6 @@ message AccountID {
 
 message LockableSerialNumbers {
     repeated int64 serial_numbers = 1;
-    // Work around a PBJ bug for now
-    bool tbd = 2;
 }
 
 /**
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -113,10 +113,6 @@ message AccountID {
     }
 }
 
-message LockableSerialNumbers {
-    repeated int64 serial_numbers = 1;
-}
-
 /**
  * Identifier for a unique token (or "NFT"), used by both contract and token services.
  */
--- a/services/token_lock.proto
+++ b/services/token_lock.proto
@@ -34,9 +34,7 @@ import "basic_types.proto";
  */
  message TokenLockTransactionBody {
     AccountID account_id = 1;
-    TokenID token_id = 2; // token-definition-id or partition-definition-id
-    oneof amountOrSerialNumbers {
-        int64 amount = 3; // if token-definition is FUNGIBLE_COMMON
-        LockableSerialNumbers serial_numbers = 4; // if the token-definition is NON_FUNGIBLE_UNIQUE
-    }
+    TokenID token_id = 2;       // token-definition-id or partition-definition-id
+    int64 amount = 3;           // if token-definition is FUNGIBLE_COMMON
+    LockableSerialNumbers serial_numbers = 4; // if the token-definition is NON_FUNGIBLE_UNIQUE
 }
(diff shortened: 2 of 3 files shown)


services/token_lock.proto
```diff
@@ -34,9 +34,7 @@ import "basic_types.proto";
  */
  message TokenLockTransactionBody {
     AccountID account_id = 1;
-    TokenID token_id = 2; // token-definition-id or partition-definition-id
-    oneof amountOrSerialNumbers {
-        int64 amount = 3; // if token-definition is FUNGIBLE_COMMON
-        LockableSerialNumbers serial_numbers = 4; // if the token-definition is NON_FUNGIBLE_UNIQUE
-    }
+    TokenID token_id = 2;       // token-definition-id or partition-definition-id
+    int64 amount = 3;           // if token-definition is FUNGIBLE_COMMON
```
iwsimon, line 38: Not sure why you changed oneof here. @Neeharika-Sompalli? A token can be either fungible or non_fungible.

netopyr: We decided to make this change consistent with the rest of the API. It is certainly not ideal. Using a `oneof` would be the better choice. However, in API design, consistency is usually considered more important.

services/token_lock.proto
```diff
@@ -34,9 +34,7 @@ import "basic_types.proto";
...
+    LockableSerialNumbers serial_numbers = 4; // if the token-definition is NON_FUNGIBLE_UNIQUE
```
netopyr, line 39:
```suggestion
    repeated int64 serialNumbers = 4; // if the token-definition is NON_FUNGIBLE_UNIQUE
```

I do not think `LockableSerialNumbers` is needed at all.

services/token_lock.proto
```diff
@@ -34,9 +34,7 @@ import "basic_types.proto";
  */
  message TokenLockTransactionBody {
     AccountID account_id = 1;
```
netopyr, line 36: In the TokenService we usually use the names `account` and `token`.

services/token_unlock.proto
```diff
@@ -34,9 +34,7 @@ import "basic_types.proto";
  */
  message TokenUnlockTransactionBody {
     AccountID account_id = 1;
-    TokenID token_id = 2; // token-definition-id or partition-definition-id
-    oneof amountOrSerialNumbers {
-        int64 amount = 3; // if token-definition is FUNGIBLE_COMMON
-        LockableSerialNumbers serial_numbers = 4; // if the token-definition is NON_FUNGIBLE_UNIQUE
-    }
+    TokenID token_id = 2;     // token-definition-id or partition-definition-id
```
netopyr, line 37: Same as above. Names `account` and `token` are more consistent. `LockableSerialNumbers` is not needed anymore.

services/basic_types.proto
```diff
@@ -115,8 +115,6 @@ message AccountID {
 
 message LockableSerialNumbers {
```
netopyr, line 116: I think we do not need this message anymore.

netopyr (changes requested):

netopyr (approved):
LGTM

iwsimon (approved):
LGTM

PR #326, opened 2024-02-14
Update for all Metadata HIPS with support for NFTs

thomas-swirlds-labs:
**Description**:

* Addition of metadata and metadataKey fields to TokenInfo and new service endpoint for updating an NFT with metadata
* Modifies the token_create, token_update and creates a new token_nft_update.proto in order to support
the 3 metadata related HIPs - [646, 657 and 765]
* Updates basic_types to add new UpdateNft to HederaFunctionality 
* Addition of TokenUpdateNftTransactionBody to transaction_body.proto
* Support for additional response codes 

**Related issue(s)**:
https://github.com/hashgraph/hedera-services/issues/10931 

**Fixes #**
https://github.com/hashgraph/hedera-services/issues/10932

**Notes for reviewer**:
This PR will need to be merged in tandem together with the changes in Hedera Services 

**Checklist**


services/token_update_nft.proto
```diff
@@ -34,19 +34,34 @@ import "google/protobuf/wrappers.proto";
...
  *
+ * Updating the metadata of an NFT does not affect its ownership or transferability.
+ * This operation is intended for updating attributes of individual NFTs in a collection.
+ *
+ * --- Signing Requirements ---
+ * 1. Whether or not a token has an admin key, its expiry can be extended with only the transaction
+ *    payer's signature.
+ * 2. To update an NFT, the account that owns the NFT must sign the transaction.
+ * 3. The metadata key is not required to be included in the TokenUpdateNftTransactionBody as it is
+ *      not required to set the metadata. The Admin key is required to sign the metadata update.
```
jascks, line 50: It makes sense the metadata key is not included, but is it the Admin key or the metadata key that is required to sign the metadata update?

services/basic_types.proto
```diff
@@ -1171,6 +1171,11 @@ enum HederaFunctionality {
      * Get a record for a transaction.
      */
     TransactionGetFastRecord = 87;
+
+    /**
+    * Update one of more NFTS in a Collection
```
tinker-michaelj, line 1176: Maybe,
```
Update the metadata of one or more NFT's of a specific token type.
```

services/response_code.proto
```diff
@@ -1373,4 +1373,19 @@ enum ResponseCodeEnum {
    * An alias that is assigned to an account or contract cannot be assigned to another account or contract.
    */
   ALIAS_ALREADY_ASSIGNED = 332;
+
+  /**
+   * A provided metadata key was invalid. Verify the bytes for an Ed25519 public key are exactly 32 bytes; and the bytes for a compressed ECDSA(secp256k1) key are exactly 33 bytes, with the first byte either 0x02 or 0x03..
```
tinker-michaelj, line 1378: Keys can have complex structures; so the verification listed here might be inapplicable or at least incomplete. 

Maybe just:
```
A provided metadata key was invalid. Verification includes, for example, checking the size of Ed25519 and ECDSA(secp256k1) public keys. 
```

services/token_update_nfts.proto
```diff
@@ -34,19 +34,34 @@ import "google/protobuf/wrappers.proto";
...
+     * The token for which to update NFTs.
+     */
+    TokenID token = 1;
+
     /**
-     * The NftID of the NFT
+     * The list of serial numbers to be updated.
      */
-    NftID nft_id = 1;
+    repeated int64 serialNumbers = 2;
```
steven-sheehy, line 61:
```suggestion
    repeated int64 serial_numbers = 2;
```

steven-sheehy (changes requested):

services/transaction_body.proto
```diff
@@ -350,5 +351,10 @@ message TransactionBody {
      * Generates a pseudorandom number.
      */
     UtilPrngTransactionBody util_prng = 52;
+
+    /**
+     * Updates individual NFTs in a collection.
+     */
+    TokenUpdateNftsTransactionBody tokenUpdateNfts = 53;
```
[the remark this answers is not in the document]
thomas-swirlds-labs: Yes we should add it. It won't be enabled yet, but if needed in the future it will already be present in the protobufs. 👍

services/token_update_nfts.proto
```diff
@@ -34,19 +34,34 @@ import "google/protobuf/wrappers.proto";
 /**
  * At consensus, updates an already created Non Fungible Token to the given values.
  * 
- * If no value is given for a field, that field is left unchanged. For an immutable tokens (that is,
- * a token without an admin key), only the expiry may be updated. Setting any other field in that
- * case will cause the transaction status to resolve to TOKEN_IS_IMMUTABLE.
+ * If no value is given for a field, that field is left unchanged.
+ * Only certain fields such as metadata can be updated. For an immutable token (that is, a token
+ * without an admin key), only the expiry may be updated. Setting any other field in that case will
+ * cause the transaction status to resolve to TOKEN_IS_IMMUTABLE.
```
xin-hedera, line 40: is this true? Can we not change token metadata / nft metadata if the token has a metadata key but no admin key?

services/token_update_nfts.proto
netopyr: Please do not do this. TokenService uses camel case almost everywhere. We discussed this and decided that consistency is more important.

services/transaction_body.proto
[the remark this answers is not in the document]
netopyr: This is so bad. Almost all transaction bodies use camel case and suddenly we are switching to snake case?

We should use the pattern that is more widespread in this file.

services/token_update_nfts.proto
steven-sheehy: Disagree. We have agreed as a company to use snake_case everywhere going forward as recommended by the [Protobuf style guide](https://protobuf.dev/programming-guides/style/#message-field-names). We have not done a mass conversion everywhere but that does not mean we should introduce new areas using camel case.

services/transaction_body.proto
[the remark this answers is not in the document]
steven-sheehy: Disagree. See https://github.com/hashgraph/hedera-protobufs/pull/326#discussion_r1504760240

services/token_update_nfts.proto
steven-sheehy: Once modularization is complete and this pbj branch is merged we should look into that mass conversion so we don't have to keep having these discussions in every PR. 😄

netopyr: When was this decided? I recall a quite recent discussion where we decided that consistency is more important, even if the outcome is sub-optimal.

Neeharika-Sompalli: @netopyr I think initially we were using `camelCase`. But there was a decision at some point to use `snake_case`. 
You can see the latest protobufs eg : `node_stake_update.proto` all use snake_case.
Since we did not change all old ones to snake_case after that decision was made it is confusing.

steven-sheehy: Believe it was in 2021 after a SDK team member pushed for some standardization changes via one or more [PRs](https://github.com/hashgraph/hedera-protobufs/pull/13). The decision was made on a call with leadership, though we should've documented it somewhere in a repo developer guide to make it official.

services/token_update_nfts.proto
Neeharika-Sompalli: We will be able to change metadata on the Nft if there is `metadata_key` on the token. Admin key is required only to update `metadata_key` on the token, similar to other keys on token.

xin-hedera: Then the comment is wrong and should be removed, i.e., the part starting from `For an immutable token`, believe it's copied from token_update.proto

Neeharika-Sompalli: Agreed. Fixed @xin-hedera

services/response_code.proto
Neeharika-Sompalli: Done. Thanks @tinker-michaelj

services/basic_types.proto
Neeharika-Sompalli: Done. Thank @tinker-michaelj

services/token_create.proto
```diff
@@ -200,8 +200,8 @@ message TokenCreateTransactionBody {
     Key pause_key = 22;
 
     /**
-   * Metadata of the created token definition.
-   */
+     * Metadata of the created token definition.
+     */
     bytes metadata = 23;
```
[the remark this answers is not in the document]
Neeharika-Sompalli: Updated javadoc in `token_update` @steven-sheehy . Thanks

services/token_update_nfts.proto
```diff
@@ -34,19 +34,28 @@ import "google/protobuf/wrappers.proto";
...
-    NftID nft_id = 1;
+    repeated int64 serial_numbers = 2;
 
     /**
-     * The new metadata of the NFT
+     * The new metadata of the NFT(s)
      */
-    google.protobuf.BytesValue metadata = 2;
-}
+    google.protobuf.BytesValue metadata = 3;
```
xin-hedera, line 60: Q: Should this be a list as well? in token mint, it's a list so newly mint nfts can have different metadata. I don't know how important it is to batch update multiple nft's metadata to be the same.

Neeharika-Sompalli: yeah as per the [HIP](https://hips.hedera.com/hip/hip-657) this transaction updates _same_ `metadata` for a batch of NFTs. 
@netopyr thoughts?

services/token_update_nfts.proto
jascks: My understanding is that when updating metadata (token level or some number of NFTs by serial number), the transaction must be signed by the metadata key. With no metadata key defined, metadata is effectively immutable.

But, it sounds like the admin key can be used to later set a metadata key, making metadata mutable once again. The only way to guarantee metadata immutability is to first remove the metadata key, then the admin key. Just removing the admin key will leave metadata forever mutable since the metadata key is now immutable.

If I'm not just misunderstanding all this, then it seems we need to be clear on the order of operations required to reach a certain result. That is, if I bought a specific NFT because its metadata plays a part in defining its value and I expect that to be unchanging (not the dynamic stone sword to diamond sword scenario), the token may be technically immutable once the admin key is cleared, but if the metadata key still exists it seems there is a problem. In the sword upgrade scenario, then the metadata key is needed regardless of the admin key.

netopyr: HIP-540, which we are going to implement for the metadata key, too, describes this in detail. If there is no admin key, the metadata key can remove itself making the token immutable.

services/token_update_nfts.proto
netopyr: Right, the same metadata should be set for all NFTs.

steven-sheehy (dismissed):

jascks (approved):
LGTM

xin-hedera (approved):
LGTM

Neeharika-Sompalli (approved):
LGTM !

iwsimon (approved):
LGTM

povolev15 (approved):
LGTM

PR #327, opened 2024-03-07
Modularization protobuf changes

netopyr:
This PR merges the branch with all the changes the modularized Services implementation requires. We have tested it thoroughly because the branch was used exclusively while developing modularization.

Closes https://github.com/hashgraph/hedera-services/issues/11937

### DIFF
--- a/services/state/blockrecords/block_info.proto
+++ b/services/state/blockrecords/block_info.proto
@@ -29,7 +29,7 @@ option java_package = "com.hederahashgraph.api.proto.java";
 option java_multiple_files = true;
 
 /**
- * Information about the most recently completed and last 256 blocks.
+ * Information about ongoing, most recently completed, and last 256 blocks.
  */
 message BlockInfo {
   /**
@@ -47,4 +47,16 @@ message BlockInfo {
    * If we are shortly after genesis and there are less than 256 blocks then this could contain less than 256 hashes.
    */
   bytes block_hashes = 3;
+  /**
+   * The consensus time of the last transaction that was handled by the node. This property is how we 'advance the
+   * consensus clock', i.e. continually setting this property to the latest consensus timestamp (and thus transaction)
+   * handled by the node.
+   */
+  Timestamp cons_time_of_last_handled_txn = 4;
+  /**
+   * A flag indicating whether migration records have been published. This property should be marked 'false'
+   * immediately following a node upgrade, and marked 'true' once the migration records (if any) are published, which
+   * should happen during the first transaction handled by the node.
+   */
+  bool migration_records_streamed = 5;
 }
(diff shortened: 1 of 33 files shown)


iwsimon (approved):
LGTM

services/state/blockrecords/block_info.proto
```diff
@@ -0,0 +1,67 @@
...
+ */
+message BlockInfo {
+  /**
+   * The last block number, this is the last completed immutable block.
+   */
+  int64 last_block_number = 1;
+  /**
+   * The consensus time of the first transaction of the last block, this is the last completed immutable block.
+   */
+  Timestamp first_cons_time_of_last_block = 2;
```
mhess-swl, line 42: To my knowledge, I haven't talked to anyone that can explain why this property is necessary, or if it's even intentional. I thought it was a mistaken duplication of `first_cons_time_of_current_block`.

services/state/common.proto
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * A single 64-bit number identifying a Hedera native entity.
+ */
+message EntityNumber {
+    int64 number = 1;
+}
+
+/**
+ * Pair of AccountID and TokenID to represent TokenRelation
```
mhess-swl, line 38: Nit: if we're calling this message `EntityIDPair`, we should change this javadoc to reflect that it can be a pair of any entity IDs. If this is only for token relations, why not call it `TokenRelationID` or similar? 

Not a blocker

Neeharika-Sompalli (approved):
LGTM !

services/state/common.proto
netopyr: @jsync-swirlds is updating the documentation. Please send suggestions to him.

jsync-swirlds: The items in this file aren't in use.  There is a follow-on change in progress that removes this file entirely (or it could be done here).

services/token_update_nft.proto
```diff
@@ -0,0 +1,52 @@
+syntax = "proto3";
+
+package proto;
+
+/*-
+ * ‌
+ * Hedera Network Services Protobuf
+ * ​
+ * Copyright (C) 2018 - 2024 Hedera Hashgraph, LLC
```
steven-sheehy, line 9: In general, new files should only show the years for which they were active and not the years in which the entire repository was active.

steven-sheehy (approved):
LGTM for everything outside the `state` directory. One comment about copyright which I won't hold things up for.

services/token_update_nft.proto
netopyr: Hey @jsync-swirlds , are you fixing this as part of your documentation work?

nathanklick: This can be easily fixed by running spotless with Git history analysis enabled.

PR #331, opened 2024-03-19
definitions for event payloads

lpetrovic05:
**Description**:
As part of the process of moving to protobuf for serialization and the blockstream effort, we start with the definitions of event payloads (formerly transactions).

**Related issue(s)**:
- https://github.com/hashgraph/hedera-services/issues/12210

### DIFF
--- a/services/state/token/token_relation.proto
+++ b/services/state/token/token_relation.proto
@@ -55,19 +55,15 @@ message TokenRelation {
      */
     bool kyc_granted = 5;
     /**
-     * The flag indicating if the token relationship was deleted.
-     */
-    bool deleted = 6;
-    /**
      * The flag indicating if the token relationship was created using automatic association.
      */
-    bool automatic_association = 7;
+    bool automatic_association = 6;
     /**
      * The previous token id of account's association linked list
      */
-    TokenID previous_token = 8;
+    TokenID previous_token = 7;
     /**
      * The next token id of account's association linked list
      */
-    TokenID next_token = 9;
+    TokenID next_token = 8;
 }


alittley (dismissed):

platform/event/state_signature_payload.proto
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * the hash that was signed
+   */
+  bytes hash = 3;
+
+  /**
+   * the hash of the epoch to which this signature corresponds to
+   */
+  bytes epochHash = 4;
```
Neeharika-Sompalli, line 52: We use snake case for all the variables
```suggestion
  bytes epoch_hash = 4;
```

platform/event/event_payload.proto
```diff
@@ -0,0 +1,49 @@
...
+   */
+  oneof payload {
+      /**
+       * application payload, the encoding of the payload is defined by the application
+       */
+      bytes applicationPayload = 1;
+      /**
+       * state signature payload
+       */
+      StateSignaturePayload stateSignaturePayload = 2;
```
Neeharika-Sompalli, line 46: Same here 

```suggestion
      StateSignaturePayload state_signature_payload = 2;
```

platform/event/event_payloads.proto
```diff
@@ -0,0 +1,39 @@
...
+import "event_payload.proto";
+
+option java_package = "com.swirlds.proto.legacy.event";
+// <<<pbj.java_package = "com.swirlds.proto.event">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+/**
+ * A list payloads of an event
+ */
+message EventPayloads {
```
nickpoorman, line 34: I'm curious what this is used for. We could save some bytes by directly using `repeated EventPayload payloads = 1;` in a message instead of the wrapper `EventPayloads`.

platform/event/event_payload.proto
```diff
@@ -0,0 +1,49 @@
...
+// <<<pbj.java_package = "com.swirlds.proto.event">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+/**
+ * The payload of an event, which is either generated by the application or the platform
+ */
+message EventPayload {
+  /**
+   * the payload
+   */
```
jsync-swirlds, line 37: comments on the `oneof` are generally not retained in specification/documentation output.  Prefer to specify the fields instead.

platform/event/event_payloads.proto
lpetrovic05: We need it for payload retraction. The hash of an event will consist of the hash of all payloads plus other event data. In order to get that hash independently, I think we need to have the payloads as a separate message, unless I'm missing something.

platform/event/event_payload.proto
```diff
@@ -0,0 +1,49 @@
...
+  oneof payload {
+      /**
+       * application payload, the encoding of the payload is defined by the application
+       */
+      bytes applicationPayload = 1;
```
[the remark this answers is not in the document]
lpetrovic05: done

platform/event/event_payload.proto
lpetrovic05: done

platform/event/state_signature_payload.proto
lpetrovic05: done

alittley (dismissed):

platform/event/event_payload.proto
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * The payload of an event, which is either generated by the application or the platform
+ */
+message EventPayload {
+  /**
+   * the payload
+   */
+  oneof payload {
+      /**
+       * application payload, the encoding of the payload is defined by the application
```
jsync-swirlds, line 40: Field documentation is markdown, except that line breaks are removed and there must not be any blank lines (the processor breaks on those).
HTML tags should be used for block quotes, lists, paragraph breaks and line breaks.
markdown still works for style (e.g. `_italic_`, `**bold**`, etc...)
```suggestion
       * An application payload.<br/>
       * The contents of this payload SHALL be defined by the application or platform
       * subsystem that created the event.<br/>
       * The contents SHOULD be a serialized protobuf message.
```

platform/event/event_payload.proto
```diff
@@ -0,0 +1,49 @@
...
+       */
+      bytes applicationPayload = 1;
+      /**
+       * state signature payload
```
jsync-swirlds, line 44:
```suggestion
       * A state signature.<br/>
       * This payload MUST be a valid state signature for an event.
```

jsync-swirlds (commented):
The comments of proto files should form a specification and (soon) will be processed into a formal document during builds.
I've offered suggestions for the `event_payload.proto`, and it would be helpful to update all of the proto files here, and any added going forward, to a similar standard.

More examples are available on the [work-in-progress branch](https://github.com/hashgraph/hedera-protobufs/tree/pbj-storage-spec-review) where everything is being updated to these expectations.

Neeharika-Sompalli (approved):
Agree with other javadoc style comments. Otherwise LGTM !

platform/event/event_payload.proto
```diff
@@ -0,0 +1,49 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
lpetrovic05: done

platform/event/event_payload.proto
lpetrovic05: ok, removed this

PR #334, opened 2024-03-21
HIP-540: add should_verify_key enum field to TokenUpdateTransactionBody

### LINKED ISSUE #332: HIP-540: add a new field in `TokenUpdateTransactionBody` to whether or not verify passed keys to be changed
**Description**
In [HIP-540](https://hips.hedera.com/hip/hip-540) is stated that a `shouldVerifyKey` boolean should be introduced in `TokenUpdateTransactionBody` to indicate whether or not to verify a passed key to be updated for a token. However, after some considerations about the current design which is: if we don’t want to verify a key in `TokenUpdateTx` we will pass false but this will immediately mean that we will remove the verification of any other passed key to `TokenUpdateTx` (we can update several keys simultaneously with one `TokenUpdateTx`) and we might want to change the behaviour in the future?

The final decision is to be an `enum` field 
cc: @netopyr

**Related issue(s)**:

Fixes [#12157](https://github.com/hashgraph/hedera-services/issues/12157)

**Solution**
Add `shouldVerifyKey` enum field to `TokenUpdateTransactionBody`, initially with two values which will indicate whether or not to verify keys and keep the possibility to extend it in the future if we want to change the behaviour

cc: @tinker-michaelj @MiroslavGatsanoga @Neeharika-Sompalli

petreze:
**Description**:
Add `shouldVerifyKey` enum field to `TokenUpdateTransactionBody`, initially with two values which will indicate whether or not to verify keys and keep the possibility to extend it in the future if we want to change the behaviour

**Related issue(s)**:

Fixes #332 

**Notes for reviewer**:
Adding the following enum will 

**Checklist**

### DIFF
--- a/services/token_update.proto
+++ b/services/token_update.proto
@@ -155,4 +155,9 @@ message TokenUpdateTransactionBody {
      * transaction will resolve to TOKEN_HAS_NO_METADATA_KEY
      */
     Key metadata_key = 17;
+
+    /**
+     * Determines whether the system should check the validity of the passed keys for update.
+     */
+    TokenKeyValidation should_verify_key = 18;
 }
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -488,6 +488,21 @@ enum TokenSupplyType {
 }
 
 /**
+ * Types of validation strategies for token keys.
+ */
+enum TokenKeyValidation {
+    /**
+     * Currently the default behaviour. It will perform all token key validations.
+     */
+    FULL_VALIDATION = 0;
+
+    /**
+     * Perform no validations at all for all passed token keys.
+     */
+    NO_VALIDATION = 1;
+}
+
+/**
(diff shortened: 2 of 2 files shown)


services/token_update.proto
```diff
@@ -155,4 +155,9 @@ message TokenUpdateTransactionBody {
      * transaction will resolve to TOKEN_HAS_NO_METADATA_KEY
      */
     Key metadata_key = 17;
+
+    /**
+     * Determines whether the system should check the validity of the passed keys for update.
+     */
+    TokenKeyValidation should_verify_key = 18;
```
netopyr, line 162: I would expect a yes/no-answer when a property is names like this. What do you think about renaming it to `key_verification_mode`?

petreze: I get your point and I totally agree!

netopyr (approved):

SimiHunjan (approved):

steven-sheehy (approved):

PR #338, opened 2024-04-01
feat: define event protobufs

alittley:
- closes https://github.com/hashgraph/hedera-services/issues/12473


platform/event/core_event_data.proto
```diff
@@ -0,0 +1,45 @@
...
+ * Contains information about an event, and its parents.
+ *
+ * ### Keywords
+ * The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+ * "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
+ * document are to be interpreted as described in [RFC2119](https://www.ietf.org/rfc/rfc2119).
+ */
+syntax = "proto3";
+
+package proto;
```
jsync-swirlds, line 12: We need to start using proper packages.
`package com.hedera.hapi.platform.event;`
Reference other protos like C++ namespaces (elide common prefixes).
The "old" messages are all prefixed with `proto` until we're able to start fixing those.

jsync-swirlds:
A couple suggestions to help get these files closer to where we would like to be in terms of specification and structure.

platform/event/core_event_data.proto
```diff
@@ -0,0 +1,45 @@
...
+import "event_descriptor.proto";
+import "timestamp.proto";
+import "basic_types.proto";
+
+option java_package = "com.swirlds.proto.legacy.event";
+// <<<pbj.java_package = "com.swirlds.proto.event">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+message CoreEventData {
+  EventDescriptor event_descriptor = 1;
```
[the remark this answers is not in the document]
alittley: You're right, it doesn't make sense. Fixing.

platform/event/event_descriptor.proto
```diff
@@ -0,0 +1,64 @@
...
+// <<<pbj.java_package = "com.hedera.hapi.platform.event">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+/**
+ * Unique identifier for an event.
+ */
+message EventDescriptor {
+  /**
+   * The hash of the event. <br/>
+   * Future work: the precise hash methodology hasn't been determined yet.
```
jsync-swirlds, line 42: For commentary like this, blockquotes are often useful (it's not as obvious in code form, but when rendered to markdown this stands out very clearly):
```suggestion
   * <blockquote>Future work<br/><blockquote>
   * The precise hash methodology has not been determined yet.</blockquote></blockquote>
```
example:
<blockquote>Future work<br/><blockquote>
The precise hash methodology has not been determined yet.</blockquote>

platform/event/event_descriptor.proto
```diff
@@ -0,0 +1,64 @@
...
+ * document are to be interpreted as described in [RFC2119](https://www.ietf.org/rfc/rfc2119).
+ */
+syntax = "proto3";
+
+package com.hedera.hapi.platform.event;
+
+/*
+ * Hedera Network Services Protobuf
+ *
+ * Copyright (C) 2018 - 2024 Hedera Hashgraph, LLC
```
jsync-swirlds, line 17: These are new files, the copyright is just `2024` for all of them, not a range of years.

platform/event/event_descriptor.proto
```diff
@@ -0,0 +1,64 @@
...
+  bytes hash = 1;
+
+  /**
+   * The ID of the node that created the event.
+   */
+  int64 creator_id = 2;
+
+  /**
+   * The generation of the event. <br/>
+   * The generation SHALL be equal to the maximum generation of all parents + 1. <br/>
```
jsync-swirlds, line 53: Could be a little clearer (and the space before the break is unnecessary)
```suggestion
   * A consensus event generation.<br/>
   * This value SHALL be **one greater** than the _maximum_ generation of all parents.<br/>
```

platform/event/core_event_data.proto
```diff
@@ -0,0 +1,74 @@
+/**
+ * # Core Event Data
+ * Contains information about an event and its parents.
```
jsync-swirlds, line 3:
```suggestion
 * A message that describes the metadata for an event.
 *
 * The `CoreEventData` links an event and it's list of parent events, and also includes
 * the software version and identifier for the node that created this event as well as the
 * birth round and creation timestamp for the event.
```

platform/event/core_event_data.proto
```diff
@@ -0,0 +1,74 @@
...
+
+  /**
+   * The ID of the node that created the event.
+   */
+  int64 creator_id = 2;
+
+  /**
+   * The birth round of the event. <br/>
+   * The birth round SHALL be equal to pending consensus round at the time the event was created. <br/>
+   * The pending consensus round is defined as the latest round to have reached consensus + 1.
```
jsync-swirlds, line 58: Could be a bit more of an imperative
```suggestion
   * The birth round SHALL be the pending consensus round at the time the event was created. <br/>
   * The pending consensus round SHALL be **one greater** than the latest round to reach consensus.
```

platform/event/core_event_data.proto
```diff
@@ -0,0 +1,74 @@
...
+   * The time created SHALL be strictly after the time created of the previous event created by the same node, if the
+   * node has created a previous event.
+   */
+  proto.Timestamp time_created = 4;
+
+  /**
+   * A list of EventDescriptors representing the parents of this event. <br/>
+   * The list of parents SHALL include zero or one self parents, and zero or more other parents.
+   */
+  repeated EventDescriptor parents = 5;
```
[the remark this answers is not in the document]
alittley: > not permitted to have more than one parent from the same creator

I'll add this.

@jasperpotts Could you weigh in here, regarding the parent ordering?

jsync-swirlds (commented):
A couple questions around the split between `CoreEventData` and `EventDescriptor` (as well as some naming questions).

The bulk of comments here are just suggestions for possible improved wording.

platform/event/core_event_data.proto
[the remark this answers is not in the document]
jasperpotts: I think it should be fine to have self parent first. Don't love it but can see advantages. The more documentation on restrictions/constraints the better. In the future we would like to start using a protobuf validation library that will let you specify a lot of those constraints in annotation that can then be auto validated on parsing.

platform/event/event_descriptor.proto
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Unique identifier for an event.
+ */
+message EventDescriptor {
+  /**
+   * The hash of the event. <br/>
+   * Future work: the precise hash methodology hasn't been determined yet.
+   */
+  bytes hash = 1;
```
jasperpotts, line 44: If we go with `bytes` type for hashes then that is minimal wasted bytes. We will need to specify that it is a 48 byte SHA384 hash in documentation. It means if we want to change it in the future we will need to add a new field with new number for that hash like: `bytes hash_sha512 = 10;` which will work fine. The other option is to have a Hash message with a enum for hash type. We already have one but unfortunately the enum is frustrating as the default value is UNKNOWN.  Default values are special as they require 0 bytes to be sent on wire. So if we use existing hash object we are adding 2 bytes for the message wrapper and 2 bytes for enum. So it costs us 4 extra bytes every time, a small amount of extra parsing serialization effort and 2 extra Java object creations.

platform/event/event_descriptor.proto
```diff
@@ -0,0 +1,64 @@
...
+   * The ID of the node that created the event.
+   */
+  int64 creator_id = 2;
+
+  /**
+   * The generation of the event. <br/>
+   * The generation SHALL be equal to the maximum generation of all parents + 1. <br/>
+   *
+   */
+  int64 generation = 3;
```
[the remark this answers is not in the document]
jasperpotts: I thought it was more but doing the math, I think the first 16 fields are special that the field tag only requires 1 bytes encoded. After that 2 bytes is good till 2048 fields. I had in my head that it was 64 but guess I was wrong or did the math wrong. See https://protobuf.dev/programming-guides/encoding/#structure . It says field tag is varint encoded which means 7 bits of data plus continuation bit. Then it is `(field_number << 3) | wire_type` so you get 7 bits in 1 byte of which 3 are used for `wire_type` that leaves you with `2^4` for field number , hence 16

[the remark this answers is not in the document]
jasperpotts: if that is the case then protecting those first 16 field numbers is a good idea. You could just make `int64 generation = 17;` that way it is 2 bytes and in the future we will not need to use it. Field numbers can have gaps.

platform/event/gossiped_event.proto
```diff
@@ -0,0 +1,61 @@
...
+   */
+  CoreEventData core_event_data = 1;
+
+  /**
+   * The signature of the event creator on the hash of the event. <br/>
+   * The signed hash MUST have been created using the same methodology as the hash in an EventDescriptor. <br/>
+   * The signature MUST have been created with the private key corresponding to the public key belonging to the event
+   * creator, as it exists in the current address book.
+   */
+  bytes signature = 2;
```
jasperpotts, line 54: We need to define what kind of signature this is, RSA, HinTS etc, similar problem to Hashes commented above.

platform/event/core_event_data.proto
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * The software version of the node at the time the event was created.<br/>
+   * This software version SHALL exactly match the software version passed into the platform at construction.
+   */
+  proto.SemanticVersion version = 1;
+
+  /**
+   * The ID of the node that created the event.
+   */
+  int64 creator_id = 2;
```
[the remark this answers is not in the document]
alittley: Looking through the other protobufs, it seems node ID is always represented by an `int64`

I'm not sure whether this should be addressed now, since these *are* completely new definitions. @jasperpotts Do you have any thoughts on this?

platform/event/gossiped_event.proto
```diff
@@ -0,0 +1,61 @@
...
+import "core_event_data.proto";
+
+option java_package = "com.hedera.hapi.platform.event";
+// <<<pbj.java_package = "com.hedera.hapi.platform.event">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+/**
+ * An event that is sent and received via gossip
+ */
+message GossipedEvent {
```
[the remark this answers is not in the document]
alittley: I believe `GossipedEvent` was chosen to avoid a collision with the existing platform `GossipEvent` class. I agree, though, that `Gossiped` is a bit awkward

@lpetrovic05 Could you weigh in here?

platform/event/gossiped_event.proto
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * The core event data
+   */
+  CoreEventData core_event_data = 1;
+
+  /**
+   * The signature of the event creator on the hash of the event. <br/>
```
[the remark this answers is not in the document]
alittley: Last I heard, there were still ongoing discussions regarding what exactly is being signed. 

I'll leave a future work comment to clarify, once a precise definition has been settled on

platform/event/core_event_data.proto
```diff
@@ -0,0 +1,74 @@
...
+import "basic_types.proto";
+
+option java_package = "com.hedera.hapi.platform.event";
+// <<<pbj.java_package = "com.hedera.hapi.platform.event">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+/**
+ * Contains information about an event and its parents.
+ */
+message CoreEventData {
```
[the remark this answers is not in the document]
alittley: I'd be fine with `EventDetail`, I've no strong opinion either way

@lpetrovic05 thoughts?

platform/event/gossiped_event.proto
```diff
@@ -0,0 +1,61 @@
...
+   * The signed hash MUST have been created using the same methodology as the hash in an EventDescriptor. <br/>
+   * The signature MUST have been created with the private key corresponding to the public key belonging to the event
+   * creator, as it exists in the current address book.
+   */
+  bytes signature = 2;
+
+  /**
+   * The event payloads. <br/>
+   * This field MAY contain zero payloads.
```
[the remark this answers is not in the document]
alittley: I don't know what the upper limit is. @lpetrovic05 @jasperpotts Has this already been considered?

platform/event/gossiped_event.proto
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * The core event data
+   */
+  CoreEventData core_event_data = 1;
+
+  /**
+   * The signature of the event creator on the hash of the event. <br/>
+   * The signed hash MUST have been created using the same methodology as the hash in an EventDescriptor. <br/>
+   * The signature MUST have been created with the private key corresponding to the public key belonging to the event
+   * creator, as it exists in the current address book.
```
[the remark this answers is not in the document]
alittley: I'll add a future work comment for now, since the discussion surrounding what type of signature to use is still ongoing

platform/event/event_descriptor.proto
```diff
@@ -0,0 +1,64 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+option java_package = "com.hedera.hapi.platform.event";
+// <<<pbj.java_package = "com.hedera.hapi.platform.event">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+/**
+ * Unique identifier for an event.
```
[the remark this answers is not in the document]
alittley: The `EventDescriptor` (as currently defined) cannot be part of the gossip event, since the `EventDescriptor` contains the event hash.

Including the event hash in the gossip event would be a waste, since every node must independently generate that hash anyway.

platform/event/core_event_data.proto
alittley: Added this extra description, with some modifications (in particular, I removed the phrase "links an event", since this expression collides with the concept of event linking that is internal to the platform)

platform/event/gossiped_event.proto
alittley: Adding a `future work` comment to specify signature type once the decision has been made

platform/event/event_descriptor.proto
alittley: My personal preference is to use the raw `bytes` field, and explicitly state what kind of hash it is

@lpetrovic05 Any preference here?

platform/event/event_descriptor.proto
jsync-swirlds: This is why we're working to improve the specification, to make these things clear.
Might be worth noting in the event descriptor message comment that this may be used for both consensus and non-consensus events (do we have specific types on non-consensus events?).

platform/event/event_descriptor.proto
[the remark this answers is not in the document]
jsync-swirlds: Perhaps detail that in the file comment. A few sentences about the design and where event hashes are _expected_ to be and why things are structured the way they are.

The file comment is a great place for larger blocks of text that discuss design decisions and document the use or purpose for various messages. The file comment may also include multiple sections with headings and other text styles not necessarily appropriate for individual field descriptions.

platform/event/core_event_data.proto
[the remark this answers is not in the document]
alittley: I think we do need these fields in both places.

I suppose the alternative would be to put an `EventDescriptor` in `CoreEventData`. But it doesn't really fit there, since `EventDescriptor` has a `generation` and a `hash`, both of which would have to be blank since they don't belong in the `CoreEventData`.

platform/event/event_descriptor.proto
```diff
@@ -0,0 +1,64 @@
+/**
+ * # Event Descriptor
+ * Unique identifier for an event.
```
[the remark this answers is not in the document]
alittley: Added a description as requested, to clarify why `EventDescriptor ` is separate

platform/event/core_event_data.proto
```diff
@@ -0,0 +1,74 @@
...
+   * The birth round of the event. <br/>
+   * The birth round SHALL be equal to pending consensus round at the time the event was created. <br/>
+   * The pending consensus round is defined as the latest round to have reached consensus + 1.
+   */
+  int64 birth_round = 3;
+
+  /**
+   * The wall clock time at which the event was created, according to the node creating the event. <br/>
+   * The time created SHALL be strictly after the time created of the previous event created by the same node, if the
+   * node has created a previous event.
```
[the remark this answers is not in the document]
alittley: I went with the first option, with some edits to fix a correctness problem with my original description.

platform/event/gossiped_event.proto
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * A node signature.<br/>
+   * The signed hash MUST be created using the same methodology as the hash in an EventDescriptor.<br/>
+   * The signature MUST verify using the public key belonging to the `event_creator`.<br/>
+   * The `event_creator` public key SHALL be read from the address book that corresponds to the event's birth round.<br/>
+   * <blockquote>Future work<br/><blockquote>
+   * The precise hash methodology hasn't been determined yet. This documentation must be updated once it has.</blockquote></blockquote>
+   * <blockquote>Future work<br/><blockquote>
+   * The type of signature to use hasn't been determined yet. This documentation must be updated once it has.</blockquote></blockquote>
```
jsync-swirlds, line 56: (very minor) It's possible to simplify when there are multiple items as follows
```suggestion
   * <blockquote>Future work<br/><blockquote>
   * The precise hash methodology hasn't been determined yet. This documentation must be updated once it has.</blockquote></blockquote>
   * <p>
   * The type of signature to use hasn't been determined yet. This documentation must be updated once it has.</blockquote></blockquote>
```

platform/event/event_descriptor.proto
alittley: The latest plan is to *not* have separate types for consensus / non-consensus events. 

I'll tweak the event descriptor message comment, to make it clear that the event descriptor can be constructed immediately, without needing to wait for the event to reach consensus

platform/event/gossiped_event.proto
alittley: Noted that the second `Future work` line is redundant, but is your suggestion correct in having 4 `</blockquote>` close tags, but only 2 `<blockquote>` open tags?

jsync-swirlds: Oops, missed that.  Should only be one pair of closing tags.

platform/event/core_event_data.proto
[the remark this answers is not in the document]
lpetrovic05: I think the platform's concept of a node ID is different from Hedera's. Sharding does not seem like something that we will be rolling out soon, so it would be hard to define if that should be part of the ID

platform/event/core_event_data.proto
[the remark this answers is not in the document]
lpetrovic05: @cody-littley is the one that suggested this name. I actually like it more than `EventDetail`. How about `CoreEventInfo`?

platform/event/core_event_data.proto
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Contains information about an event and its parents.
+ */
+message CoreEventData {
+  /**
+   * The software version of the node at the time the event was created.<br/>
+   * This software version SHALL exactly match the software version passed into the platform at construction.
+   */
+  proto.SemanticVersion version = 1;
```
lpetrovic05, line 52: A reminder that all non Hedera apps will have to switch to semantic version

platform/event/event_descriptor.proto
lpetrovic05: Previously we have decided to go with bytes for hashes and signatures, I think for performance reasons. This is how I did it for the state signature payload. So just mention that is should be a SHA384 hash.

platform/event/event_descriptor.proto
```diff
@@ -0,0 +1,77 @@
...
+   * The birth round SHALL be the pending consensus round at the time the event is created.<br/>
+   * The pending consensus round SHALL be **one greater** than the latest round to reach consensus.
+   */
+  int64 birth_round = 3;
+
+  /**
+  * The generation of the event.<br/>
+  * This value SHALL be **one greater** than the _maximum_ generation of all parents.<br/>
+  */
+  int64 generation = 17;
```
lpetrovic05, line 76: 17?

platform/event/gossiped_event.proto
lpetrovic05: currently we use SHA384withRSA

platform/event/gossiped_event.proto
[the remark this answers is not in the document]
lpetrovic05: Right now in the platform this is a configuration option. Don't know what the plan is going forward

platform/event/event_descriptor.proto
alittley: See the comment from Jasper above https://github.com/hashgraph/hedera-protobufs/pull/338#discussion_r1550128991

I'm not sure if it makes sense to document this reason in the comment for the field since their style seems so... formal

platform/event/core_event_data.proto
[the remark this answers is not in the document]
alittley: Of the suggestions so far, I think `EventCore` is most simple and to the point.

platform/event/event_descriptor.proto
jsync-swirlds: You could add a line-end comment.  Perhaps something like `// This field is temporary until birth_round is complete.`

platform/event/gossiped_event.proto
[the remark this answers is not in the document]
jsync-swirlds: If there is an upper limit, even if configured, we should probably document here. Something like this, perhaps.
```proto
 * This field MUST NOT exceed `configurationName` entries, initially `NN`.
```

platform/event/core_event_data.proto
jsync-swirlds: Do we know what happened to the HIP to change to a date-based version?
That would reuse the `SemanticVersion` object, but change the meaning of the fields.

platform/event/core_event_data.proto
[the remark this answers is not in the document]
alittley: I've preliminarily made the change to `EventCore`. We can change it again if the conversation continues to develop

platform/event/event_descriptor.proto
alittley: Specified the hash type

platform/event/gossiped_event.proto
jsync-swirlds: Any thought to perhaps using edDSA instead of RSA at some point?  RSA signatures are much larger (and hence cost much more bandwidth) than edDSA.

alittley: I went ahead and specified `SHA384withRSA` in the doc, since it is the status quo, and none of the alternatives are a clear improvement:

- a BLS signature is small but extremely expensive to verify, and not quantum secure
- ed25519 is moderately smaller than RSA, but also not quantum secure

platform/event/core_event_data.proto
alittley: Is there anything you think needs doing in this regard?

platform/event/event_descriptor.proto
alittley: Done

platform/event/gossiped_event.proto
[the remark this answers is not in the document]
alittley: Added

platform/event/gossiped_event.proto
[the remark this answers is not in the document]
alittley: Thinking some more on this, I think we ought to call the protobuf `GossipEvent`, and rename the Platform class currently called by that name as soon as convenient.

At the end of the day, THIS *is* a gossip event.. Whatever we wrap around the protobuf for internal usage is by definition *not* a gossip event.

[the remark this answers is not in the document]
cody-littley: I also prefer `GossipEvent`. @lpetrovic05 can you remember why the name `GossipedEvent` was chosen over `GossipEvent`? I remember the conversation but not the conclusion.

platform/event/gossiped_event.proto
```diff
@@ -0,0 +1,66 @@
...
+   * The signed hash MUST be created using the same methodology as the hash in an EventDescriptor.<br/>
+   * <blockquote>Future work<br/><blockquote>
+   * The precise hash methodology hasn't been determined yet. This documentation must be updated once it has.</blockquote></blockquote>
+   */
+  bytes signature = 2;
+
+  /**
+   * The event payloads.<br/>
+   * This field MAY contain zero payloads.<br/>
+   * This field MUST NOT exceed `maxTransactionCountPerEvent` entries, initially `245760`.<br/>
```
jsync-swirlds, line 62: count and bytes are the same?

alittley: Indeed

platform/event/gossiped_event.proto
jsync-swirlds: In fairness, RSA isn't quantum secure either.  I don't know of any public key algorithm currently that's quantum secure (which might be a problem, except practical quantum computing is still quite a ways off, in my opinion).

jsync-swirlds (approved):

platform/event/gossiped_event.proto
alittley: 👍 My main argument for RSA is just status quo.

jsync-swirlds: Understood.
I wasn't arguing for an immediate change.
The signature size has a significant impact on block stream data volume, however, so it might be worth _considering_ a future change.

jsync-swirlds (dismissed):

platform/event/gossiped_event.proto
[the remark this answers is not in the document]
lpetrovic05: GossipEvent makes sense to me

lpetrovic05 (dismissed):

platform/event/gossiped_event.proto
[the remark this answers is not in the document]
alittley: Thusly renamed

platform/event/core_event_data.proto
alittley: I don't see any hips about a date-based version

jsync-swirlds: It has mostly been in slack so far, I'll send you the link.

jsync-swirlds (dismissed):

cody-littley (dismissed):

jsync-swirlds (dismissed):

edward-swirldslabs (approved):

platform/event/event_core.proto
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Contains information about an event and its parents.
+ */
+message EventCore {
+  /**
+   * The software version of the node at the time the event was created.<br/>
+   * This software version SHALL exactly match the software version passed into the platform at construction.
+   */
+  proto.SemanticVersion version = 1;
```
rbair23, line 52: At the moment, software version and specification version are the same thing. But imagine a future where there were multiple software releases that were completely compatible (just suspend disbelief for the moment to consider this :-)). Should this really be a specification version?

platform/event/event_core.proto
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * A Creator Identifier.<br/>
+   * This SHALL be the unique identifier for the node that created the event.
+   */
```
rbair23, line 57: Is it worth also documenting this ID is the same as the node ID in the address book for this node? Or is it something different? creator_node_id maybe?

platform/event/event_descriptor.proto
```diff
@@ -0,0 +1,79 @@
...
+   * The birth round SHALL be the pending consensus round at the time the event is created.<br/>
+   * The pending consensus round SHALL be **one greater** than the latest round to reach consensus.
+   */
+  int64 birth_round = 3;
+
+  /**
+  * The generation of the event.<br/>
+  * This value SHALL be **one greater** than the _maximum_ generation of all parents.<br/>
+  */
+  int64 generation = 17; // This field is temporary until birth_round migration is complete. Field number 17 chosen to avoid polluting cheaper 1 byte field numbers 1-16
```
rbair23, line 78: I love that comment! Very helpful

platform/event/event_descriptor.proto
```diff
@@ -0,0 +1,79 @@
...
+   * This SHALL be the unique identifier for the node that created the event.
+   */
+  int64 creator_id = 2;
+
+  /**
+   * The birth round of the event.<br/>
+   * The birth round SHALL be the pending consensus round at the time the event is created.<br/>
+   * The pending consensus round SHALL be **one greater** than the latest round to reach consensus.
+   */
+  int64 birth_round = 3;
```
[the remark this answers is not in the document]
edward-swirldslabs: in a dynamic address book world, the birth round is necessary to know so that we know which address book was active during event creation.  In order to validate the signature, we have to use the address book that contains the node that the event came from.    It is also going to be the means of determining when an event becomes ancient instead of using the generation of the event to determine ancient.

platform/event/event_core.proto
alittley: It is the same. I modified the documentation, and the name of the field, to reflect this

jsync-swirlds (dismissed):

edward-swirldslabs (dismissed):

jsync-swirlds (approved):

platform/event/event_core.proto
alittley: I changed it to a `specification_version`, and also moved the field into `GossipEvent`, since the spec covers all of the event objects

edward-swirldslabs (approved):

jsync-swirlds (approved):

lpetrovic05 (dismissed):

rbair23 (approved):

edward-swirldslabs (dismissed):

jsync-swirlds (approved):

cody-littley (approved):

edward-swirldslabs (approved):

PR #340, opened 2024-04-09
HIP-540: add new response code to support

### LINKED ISSUE #339: HIP-540: Add new Response Code `TOKEN_HAS_NO_ADMIN_KEY`
Add new `TOKEN_HAS_NO_ADMIN_KEY` response code to support HIP-540

petreze:
**Description**:
Create a `TOKEN_HAS_NO_ADMIN_KEY` response code needed after HIP-540

**Related issue(s)**:

Fixes #339 

**Notes for reviewer**:
Previously, when a token has no admin key on initialization, it is considered to be immutable and when a TokenUpdate is performed to those kind of tokens, we return `TOKEN_IS_IMMUTABLE` error.

After HIP-540, a token without an admin key on initialization is **not** considered immutable when it has a low priority 
on create. In this case a low priority key can update itself even if the token has no admin key.

We need a new response code - `TOKEN_HAS_NO_ADMIN_KEY` for the cases where a token does not have an admin key on create and in TokenUpdate we try to update the admin key. Previously, this resulted in `TOKEN_IS_IMMUTABLE` but after the HIP this is not the case anymore so we need new response status to indicate the reason of the failure

**Checklist**

### DIFF
--- a/services/response_code.proto
+++ b/services/response_code.proto
@@ -1392,4 +1392,9 @@ enum ResponseCodeEnum {
    * NFT serial numbers are missing in the TokenUpdateNftsTransactionBody
    */
   MISSING_SERIAL_NUMBERS = 336;
+
+  /**
+   * Admin key is not set on token
+   */
+  TOKEN_HAS_NO_ADMIN_KEY = 337;
 }


steven-sheehy (approved):

Neeharika-Sompalli (approved):
LGTM !

xin-hedera (approved):
LGTM

PR #346, opened 2024-04-27
feat: Update specification text for address book service

jsync-swirlds:
* Updated specification/API comments

### Rendered documentation
A rendered version of the documentation in these files is _temporarily_ available on [another branch](https://github.com/hashgraph/hedera-protobufs/tree/temp-DAB-Update-Spec-Backup/documents/api/services) to assist with visualizing what these changes accomplish.

### DIFF
--- a/services/address_book_service.proto
+++ b/services/address_book_service.proto
@@ -1,10 +1,8 @@
 syntax = "proto3";
 
-package proto;
+package com.hedera.hapi.node.addressbook;
 
 /*
- * Hedera Network Services Protobuf
- *
  * Copyright (C) 2024 Hedera Hashgraph, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -29,37 +27,140 @@ import "transaction_response.proto";
 import "transaction.proto";
 
 /**
- * Transactions for the AddressBook Service, those HAPI APIs facilitate changes to the nodes used across the Hedera network.
- * All those transactions needs to be signed by the Hedera Council. Steps needed below:
- * 1. The node operator creates and signs the transaction with their key (the key on the node operator account)
- * 2. The node operator hands this transaction to Alex, who then gives it to the council to sign
- * 3. When signed and submitted, the server will verify that account 2 keys have signed, and the keys on the operator account have signed.
- * Hedera council should have ability to make all edits in addition to add/remove nodes
+ * The Address Book service provides the ability for Hedera network node
+ * administrators to add, update, and remove consensus nodes. This addition,
+ * update, or removal of a consensus node requires governing council approval,
+ * but each node operator may update their own operational attributes without
+ * additional approval, reducing overhead for routine operations.
+ *
+ * Most operations are `privileged operations` and require governing council
+ * approval.
+ *
+ * ### For a node creation transaction.
+ * - The node operator SHALL create a `createNode` transaction.
+ *    - The node operator SHALL sign this transaction with the active `key` for
+ *      the account to be assigned as the "node account".
+ *    - The node operator MUST deliver the signed transaction to the Hedera
+ *      council representative.
+ *    - The Hedera council representative SHALL arrange for council members to
+ *      review and sign the transaction.
+ *    - Once sufficient council members have signed the transaction, the
+ *      Hedera council representative SHALL submit the transaction to the
+ *      network.
+ * - Upon receipt of a valid and signed node creation transaction the network
+ *   software SHALL
+ *    - Validate the threshold signature for the Hedera governing council
+ *    - Validate the signature of the active `key` for the account to be
+ *      assigned as the "node account".
+ *    - Create the new node in state, this new node SHALL NOT be active in the
+ *      network at this time.
+ *    - When executing the next `freeze` transaction with `freeze_type` set to
+ *      `PREPARE_UPGRADE`, update network configuration and bring the
+ *      new node to an active status within the network. The node to be added
+ *      SHALL be active in the network following this upgrade.
+ *
+ * ### For a node deletion transaction.
+ * - The node operator or Hedera council representative SHALL create a
+ *   `deleteNode` transaction.
+ *    - If the node operator creates the transaction
+ *       - The node operator MUST sign this transaction with the active `key`
+ *         for the account assigned as the "node account".
+ *       - The node operator SHALL deliver the signed transaction to the Hedera
+ *         council representative.
+ *    - The Hedera council representative SHALL arrange for council members to
+ *      review and sign the transaction.
+ *    - Once sufficient council members have signed the transaction, the
+ *      Hedera council representative SHALL submit the transaction to the
+ *      network.
+ * - Upon receipt of a valid and signed node deletion transaction the network
+ *   software SHALL
+ *    - Validate the threshold signature for the Hedera governing council
+ *    - Remove the existing node from network state. The node SHALL still
+ *      be active in the network at this time.
+ *    - When executing the next `freeze` transaction with `freeze_type` set to
+ *      `PREPARE_UPGRADE`, update network configuration and remove the
+ *      node to be deleted from the network. The node to be deleted SHALL NOT
+ *      be active in the network following this upgrade.
+ *
+ * ### For a node update transaction.
+ * - The node operator or Hedera council representative SHALL create an
+ *   `updateNode` transaction.
+ *    - If the node operator creates the transaction
+ *       - The node operator MUST sign this transaction with the active `key`
+ *         for the account assigned as the current "node account".
+ *       - If the transaction changes the value of the "node account" the
+ *         node operator MUST _also_ sign this transaction with the active `key`
+ *         for the account to be assigned as the new "node account".
+ *       - The node operator SHALL submit the transaction to the
+ *         network.  Hedera council approval SHALL NOT be sought for this
+ *         transaction
+ *    - If the Hedera council representative creates the transaction
+ *       - The Hedera council representative SHALL arrange for council members
+ *         to review and sign the transaction.
+ *       - Once sufficient council members have signed the transaction, the
+ *         Hedera council representative SHALL submit the transaction to the
+ *         network.
+ * - Upon receipt of a valid and signed node update transaction the network
+ *   software SHALL
+ *    - If the transaction is signed by the Hedera governing council
+ *       - Validate the threshold signature for the Hedera governing council
+ *    - If the transaction is signed by the active `key` for the node account
+ *       - Validate the signature of the active `key` for the account assigned
+ *         as the "node account".
+ *    - If the transaction modifies the value of the "node account",
+ *       - Validate the signature of the _new_ `key` for the account to be
+ *         assigned as the new "node account".
+ *    - Modify the node information held in network state with the changes
+ *      requested in the update transaction. The node changes SHALL NOT be
+ *      applied to network configuration, and SHALL NOT affect network
+ *      operation at this time.
+ *    - When executing the next `freeze` transaction with `freeze_type` set to
+ *      `PREPARE_UPGRADE`, update network configuration according to the
+ *      modified information in network state. The requested changes SHALL
+ *      affect network operation following this upgrade.
  */
 service AddressBookService {
     /**
-      * Prepare to add a new node to the network.
-      * When a valid council member initiates a HAPI transaction to add a new node,
-      * then the network should acknowledge the transaction and update the network’s Address Book within 24 hours.
-      * The added node will not be active until the network is upgraded.
-      */
-    rpc createNode (Transaction) returns (TransactionResponse);
+     * A transaction to create a new consensus node in the network.
+     * address book.
+     * <p>
+     * This transaction, once complete, SHALL add a new consensus node to the
+     * network state.<br/>
+     * The new consensus node SHALL remain in state, but SHALL NOT participate
+     * in network consensus until the network updates the network configuration.
+     * <p>
+     * Hedera governing council authorization is REQUIRED for this transaction.
+     */
+    rpc createNode (proto.Transaction) returns (proto.TransactionResponse);
 
     /**
-     * Prepare to delete the node to the network.
-     * The deleted node will not be deleted until the network is upgraded.
-     * Such a deleted node can never be reused.
+     * A transaction to remove a consensus node from the network address
+     * book.
+     * <p>
+     * This transaction, once complete, SHALL remove the identified consensus
+     * node from the network state.
+     * <p>
+     * Hedera governing council authorization is REQUIRED for this transaction.
      */
-    rpc deleteNode (Transaction) returns (TransactionResponse);
+    rpc deleteNode (proto.Transaction) returns (proto.TransactionResponse);
 
     /**
-     * Prepare to update the node to the network.
-     * The node will not be updated until the network is upgraded.
+     * A transaction to update an existing consensus node from the network
+     * address book.
+     * <p>
+     * This transaction, once complete, SHALL modify the identified consensus
+     * node state as requested.
+     * <p>
+     * This transaction MAY be authorized by either the node operator OR the
+     * Hedera governing council.
      */
-    rpc updateNode (Transaction) returns (TransactionResponse);
+    rpc updateNode (proto.Transaction) returns (proto.TransactionResponse);
 
     /**
-     * Retrieves the node information by node Id.
+     * A transaction to query the current state of a consensus node in the
+     * network address book state.
+     * <p>
+     * Hedera governing council authorization is REQUIRED for this transaction.
      */
-    rpc getNodeInfo (Query) returns (Response);
+    rpc getNodeInfo (proto.Query) returns (proto.Response);
 }
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -67,7 +67,7 @@ message RealmID {
 }
 
 /**
- * The ID for an a cryptocurrency account 
+ * The ID for an a cryptocurrency account
  */
 message AccountID {
     /**
@@ -130,7 +130,7 @@ message NftID {
 }
 
 /**
- * The ID for a file 
+ * The ID for a file
  */
 message FileID {
     /**
@@ -150,7 +150,7 @@ message FileID {
 }
 
 /**
- * The ID for a smart contract instance 
+ * The ID for a smart contract instance
  */
 message ContractID {
     /**
@@ -170,21 +170,21 @@ message ContractID {
         int64 contractNum = 3;
 
         /**
-        * The 20-byte EVM address of the contract to call. 
-        * 
+        * The 20-byte EVM address of the contract to call.
+        *
         * Every contract has an EVM address determined by its <tt>shard.realm.num</tt> id.
         * This address is as follows:
         *   <ol>
         *     <li>The first 4 bytes are the big-endian representation of the shard.</li>
         *     <li>The next 8 bytes are the big-endian representation of the realm.</li>
         *     <li>The final 8 bytes are the big-endian representation of the number.</li>
-        *   </ol>  
-        * 
-        * Contracts created via CREATE2 have an <b>additional, primary address</b> that is 
-        * derived from the <a href="https://eips.ethereum.org/EIPS/eip-1014">EIP-1014</a> 
-        * specification, and does not have a simple relation to a <tt>shard.realm.num</tt> id. 
-        * 
-        * (Please do note that CREATE2 contracts can also be referenced by the three-part 
+        *   </ol>
+        *
+        * Contracts created via CREATE2 have an <b>additional, primary address</b> that is
+        * derived from the <a href="https://eips.ethereum.org/EIPS/eip-1014">EIP-1014</a>
+        * specification, and does not have a simple relation to a <tt>shard.realm.num</tt> id.
+        *
+        * (Please do note that CREATE2 contracts can also be referenced by the three-part
         * EVM address described above.)
         */
         bytes evm_address = 4;
@@ -225,13 +225,13 @@ message TransactionID {
     bool scheduled = 3;
 
     /**
-     * The identifier for an internal transaction that was spawned as part 
-     * of handling a user transaction. (These internal transactions share the 
+     * The identifier for an internal transaction that was spawned as part
+     * of handling a user transaction. (These internal transactions share the
      * transactionValidStart and accountID of the user transaction, so a
      * nonce is necessary to give them a unique TransactionID.)
-     * 
-     * An example is when a "parent" ContractCreate or ContractCall transaction 
-     * calls one or more HTS precompiled contracts; each of the "child" 
+     *
+     * An example is when a "parent" ContractCreate or ContractCall transaction
+     * calls one or more HTS precompiled contracts; each of the "child"
      * transactions spawned for a precompile has a id with a different nonce.
      */
     int32 nonce = 4;
@@ -426,19 +426,19 @@ enum TokenType {
 }
 
 /**
- * Allows a set of resource prices to be scoped to a certain type of a HAPI operation. 
- * 
+ * Allows a set of resource prices to be scoped to a certain type of a HAPI operation.
+ *
  * For example, the resource prices for a TokenMint operation are different between minting fungible
  * and non-fungible tokens. This enum allows us to "mark" a set of prices as applying to one or the
  * other.
- * 
+ *
  * Similarly, the resource prices for a basic TokenCreate without a custom fee schedule yield a
  * total price of $1. The resource prices for a TokenCreate with a custom fee schedule are different
  * and yield a total base price of $2.
  */
 enum SubType {
     /**
-     * The resource prices have no special scope 
+     * The resource prices have no special scope
      */
     DEFAULT = 0;
 
@@ -567,15 +567,15 @@ enum TokenPauseStatus {
  * A Key can be a public key from either the Ed25519 or ECDSA(secp256k1) signature schemes, where
  * in the ECDSA(secp256k1) case we require the 33-byte compressed form of the public key. We call
  * these public keys <b>primitive keys</b>.
- * 
- * If an account has primitive key associated to it, then the corresponding private key must sign 
- * any transaction to transfer cryptocurrency out of it. 
  *
- * A Key can also be the ID of a smart contract instance, which is then authorized to perform any 
- * precompiled contract action that requires this key to sign. 
+ * If an account has primitive key associated to it, then the corresponding private key must sign
+ * any transaction to transfer cryptocurrency out of it.
  *
- * Note that when a Key is a smart contract ID, it <i>doesn't</i> mean the contract with that ID 
- * will actually create a cryptographic signature. It only means that when the contract calls a 
+ * A Key can also be the ID of a smart contract instance, which is then authorized to perform any
+ * precompiled contract action that requires this key to sign.
+ *
+ * Note that when a Key is a smart contract ID, it <i>doesn't</i> mean the contract with that ID
+ * will actually create a cryptographic signature. It only means that when the contract calls a
  * precompiled contract, the resulting "child transaction" will be authorized to perform any action
  * controlled by the Key.
  *
@@ -639,10 +639,10 @@ message Key {
          */
         bytes ECDSA_secp256k1 = 7;
 
-        /**  
+        /**
          * A smart contract that, if the recipient of the active message frame, should be treated
-         * as having signed. (Note this does not mean the <i>code being executed in the frame</i> 
-         * will belong to the given contract, since it could be running another contract's code via 
+         * as having signed. (Note this does not mean the <i>code being executed in the frame</i>
+         * will belong to the given contract, since it could be running another contract's code via
          * <tt>delegatecall</tt>. So setting this key is a more permissive version of setting the
          * contractID key, which also requires the code in the active message frame belong to the
          * the contract with the given id.)
@@ -683,7 +683,7 @@ message KeyList {
 }
 
 /**
- * This message is <b>DEPRECATED</b> and <b>UNUSABLE</b> with network nodes. It is retained 
+ * This message is <b>DEPRECATED</b> and <b>UNUSABLE</b> with network nodes. It is retained
  * here only for historical reasons.
  *
  * Please use the SignaturePair and SignatureMap messages.
@@ -726,7 +726,7 @@ message Signature {
 }
 
 /**
- * This message is <b>DEPRECATED</b> and <b>UNUSABLE</b> with network nodes. It is retained 
+ * This message is <b>DEPRECATED</b> and <b>UNUSABLE</b> with network nodes. It is retained
  * here only for historical reasons.
  *
  * Please use the SignaturePair and SignatureMap messages.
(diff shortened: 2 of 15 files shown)


services/address_book_service.proto
```diff
@@ -29,37 +27,127 @@ import "transaction_response.proto";
...
 
 /**
- * Transactions for the AddressBook Service, those HAPI APIs facilitate changes to the nodes used across the Hedera network.
- * All those transactions needs to be signed by the Hedera Council. Steps needed below:
- * 1. The node operator creates and signs the transaction with their key (the key on the node operator account)
- * 2. The node operator hands this transaction to Alex, who then gives it to the council to sign
- * 3. When signed and submitted, the server will verify that account 2 keys have signed, and the keys on the operator account have signed.
- * Hedera council should have ability to make all edits in addition to add/remove nodes
+ * The Address Book service provides the ability for Hedera network node
+ * administrators to add and remove consensus nodes. The addition or removal
```
iwsimon, line 31: administrators to add, update and remove consensus nodes. The addition, updating or removal

services/address_book_service.proto
```diff
@@ -29,37 +27,127 @@ import "transaction_response.proto";
...
+ *      council representative.
+ *    - The Hedera council representative SHALL arrange for council members to
+ *      review and sign the transaction.
+ *    - Once sufficient council members have signed the transaction, the
+ *      Hedera council representative SHALL submit the transaction to the
+ *      network.
+ * - Upon receipt of a valid and signed node creation transaction the network
+ *   software SHALL
+ *    - Validate the threshold signature for the Hedera governing council
+ *    - Validate the signature of the "admin key" for the account to be assigned
```
iwsimon, line 53: There is no "admin key" in these transactions.

jsync-swirlds: The idea is to specify which key must sign.  Account only has one key attribute, however, so I'll reword and clarify.

services/address_book_service.proto
```diff
@@ -29,37 +27,127 @@ import "transaction_response.proto";
...
+ *      as the "node account".
+ *    - Create the new node in a pending state.
+ *    - When executing the next `freeze` transaction with `freeze_type` set to
+ *      `PREPARE_UPGRADE`, update network configuration and transition the
```
iwsimon, line 57: When executing the next `freeze` transaction with `freeze_type` set to`PREPARE_UPGRADE`, config.txt, .pem file per each node, and a marker file will be generated.

services/address_book_service.proto
```diff
@@ -29,37 +27,127 @@ import "transaction_response.proto";
...
+ *      Hedera council representative SHALL submit the transaction to the
+ *      network.
+ * - Upon receipt of a valid and signed node deletion transaction the network
+ *   software SHALL
+ *    - Validate the threshold signature for the Hedera governing council
+ *    - Validate the signature of the "admin key" for the account assigned
+ *      as the "node account", _if present_.
+ *    - Modify the existing node to a pending delete state.
+ *    - When executing the next `freeze` transaction with `freeze_type` set to
+ *      `PREPARE_UPGRADE`, update network configuration and transition the
```
iwsimon, line 80: Same as above.

services/address_book_service.proto
```diff
@@ -29,37 +27,127 @@ import "transaction_response.proto";
...
     /**
-      * Prepare to add a new node to the network.
-      * When a valid council member initiates a HAPI transaction to add a new node,
-      * then the network should acknowledge the transaction and update the network’s Address Book within 24 hours.
-      * The added node will not be active until the network is upgraded.
-      */
-    rpc createNode (Transaction) returns (TransactionResponse);
+     * A transaction to create a new consensus node in the network.
+     * address book.<br/>
+     * This transaction, once complete, SHALL add a new consensus node to the
```
iwsimon, line 118: Even this transaction is complete, the new node is still not joining the network until next freeze upgrade for phase 2 and not joining the network until consensus roster sent in the next few round for phase 3 (more details still pending to be narrowed down.)

services/node_create.proto
```diff
@@ -27,62 +25,90 @@ option java_multiple_files = true;
...
  * to join the network, and requires governing council authorization.
  *
- * A `NodeCreateTransactionBody` MUST be signed by the governing council.<br/>
- * The newly created node information will be used to generate config.txt and
- * a-pulbic-NodeAlias.pem file per each node during phase 2,<br>
- * not active until next freeze upgrade.
+ * - A `NodeCreateTransactionBody` MUST be signed by the governing council.
+ * - The newly created node information SHALL be added to the network address
+ *   book information in the network state.
+ * - The new entry SHALL be created in a "pending" state.
```
iwsimon, line 39: We should remove "pending" state here.

services/address_book_service.proto
```diff
@@ -29,37 +27,127 @@ import "transaction_response.proto";
...
- * All those transactions needs to be signed by the Hedera Council. Steps needed below:
- * 1. The node operator creates and signs the transaction with their key (the key on the node operator account)
- * 2. The node operator hands this transaction to Alex, who then gives it to the council to sign
- * 3. When signed and submitted, the server will verify that account 2 keys have signed, and the keys on the operator account have signed.
- * Hedera council should have ability to make all edits in addition to add/remove nodes
+ * The Address Book service provides the ability for Hedera network node
+ * administrators to add and remove consensus nodes. The addition or removal
+ * of a consensus node requires governing council approval, but each node
+ * operator may update their own operational attributes without additional
+ * approval, reducing overhead for routine operations.
```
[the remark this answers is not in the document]
jsync-swirlds: HIP text _appears to_ claim a node operator can update their own node *without* council approval. Is council approval required for everything (in which case node operators are never actually required), or are node operators permitted to update their own nodes independently?

services/address_book_service.proto
jsync-swirlds: I prefer not to include internal functional details (when _completely_ invisible to external operations) in specification.
Stating what observable transaction causes a state change is helpful to an implementor or client; but including files written to local disk seems an unnecessary detail (another implementation may not do the same thing, it's not observable, and clients shouldn't care).

services/node_create.proto
jsync-swirlds: What state is the internal (in the merkle tree) entry created, if not "pending"?  I just copied that from the previous text and reworded.

services/address_book_service.proto
jsync-swirlds: I'll clarify that the node is added to _state_ (which it must be, or the consensus roster will never be updated in a later process).

services/node_create.proto
```diff
@@ -27,62 +25,90 @@ option java_multiple_files = true;
...
+     * consensus nodes.<br/>
+     * All other entries SHALL be reserved for future use.
      */
-    repeated ServiceEndpoint gossip_endpoint = 3;
+    repeated proto.ServiceEndpoint gossip_endpoint = 3;
 
     /**
-     * A node's grpc service IP addresses and ports, IP:Port is mandatory,
-     * fully qualified domain name is optional. Maximum number of these endpoints is 8.
+     * A list of service endpoints.<br/>
```
iwsimon, line 82: A list of grpc service endpoints. "grpc" needs to be specified in the previous notion discuss.

services/node_create.proto
```diff
@@ -27,62 +25,90 @@ option java_multiple_files = true;
...
+     * These endpoints SHALL represent the published endpoints to which clients
+     * may submit transactions.<br/>
+     * These endpoints MUST specify address and port.<br/>
```
iwsimon, line 85: ip address instead of address

services/node_create.proto
```diff
@@ -27,62 +25,90 @@ option java_multiple_files = true;
...
+     * These endpoints MUST NOT specify a DNS name.<br/>
```
iwsimon, line 86: DNS name is optional, not "MUST NOT".

services/node_create.proto
jsync-swirlds: These are service endpoints.  gRPC should be specified in the requirements lines below if that is strictly required (not sure it truly is, but no harm adding it for now).

I'll add the gRPC requirement.

services/node_create.proto
jsync-swirlds: And yet, in another location we state the system will fail if DNS is specified.
We have to decide if this is allowed or not and make it clear.  I assumed the more restrictive of the two, as "this is optional, but your transaction will fail" is not what I call optional.

services/node_create.proto
```diff
@@ -27,62 +25,90 @@ option java_multiple_files = true;
...
 
     /**
-     * The node's X509 certificate used to sign stream files (e.g., record stream
-     * files). Precisely, this field is the DER encoding of gossip X509 certificate.
-     * This is a mandatory field.
+     * A certificate used to sign gossip events.<br/>
+     * This value MUST be a certificate of a type permitted for gossip
+     * signatures.<br/>
+     * The value presented MUST be a UTF-8 NFKD encoding of the TLS
+     * certificate.<br/>
```
iwsimon, line 97: Keep this "Precisely, this field is the DER encoding of gossip X509 certificate."

jsync-swirlds: That sentence does not make sense.
Also, the latest update in channel is that we must use PEM encoding, not DER.
Note, that this won't merge before the next discussion where additional updates for the HIP are likely.

services/node_create.proto
jsync-swirlds: I believe `address` is sufficient here.
Adding `IP` is unnecessary as the sub-type of address is defined by the `ServiceEndpoint` message.

services/node_create.proto
iwsimon: For gossip_endpoint, fully qualified domain name is not allowed; for service_endpoint, fully qualified domain name is allowed.

jsync-swirlds: Definitely some vagueness in HIP text for that.
I'll adjust the `service_endpoint` text to state the endpoints `MAY` specify a DNS name.

services/node_create.proto
jsync-swirlds: I've updated this text with a possible reword; it's a bit of a guess, so please review carefully.

services/node_create.proto
jsync-swirlds: Wording updated to flow better and be a bit more clear on formats and encoding.

services/address_book_service.proto
```diff
@@ -29,37 +36,132 @@ import "transaction_response.proto";
...
- * All those transactions needs to be signed by the Hedera Council. Steps needed below:
- * 1. The node operator creates and signs the transaction with their key (the key on the node operator account)
- * 2. The node operator hands this transaction to Alex, who then gives it to the council to sign
- * 3. When signed and submitted, the server will verify that account 2 keys have signed, and the keys on the operator account have signed.
- * Hedera council should have ability to make all edits in addition to add/remove nodes
+ * The Address Book service provides the ability for Hedera network node
+ * administrators to add, update, and remove consensus nodes. This addition,
+ * update, or removal of a consensus node requires governing council approval,
+ * but each node operator may update their own operational attributes without
+ * additional approval, reducing overhead for routine operations.
```
iwsimon, line 43: All updates need approval.

services/node_create.proto
iwsimon: Address could have different meanings. We have Address and AddressBook concepts in the code. We can use IP instead of address here.

services/node_create.proto
iwsimon: DER is the encoding format to store the certificate. PEM is the file extension.

services/state/addressbook/node.proto
```diff
@@ -27,81 +34,164 @@ option java_package = "com.hederahashgraph.api.proto.java";
...
+ *   as full consensus nodes.
+ * - If a `deleteNode` transaction is processed for a consensus node, the node
+ *   status SHALL change to the `PENDING_DELETION` state.
+ * - Following the next `network upgrade` every node with the status
+ *   `PENDING_DELETE` SHALL be removed from the live address book, the status
+ *   SHALL change to `DELETED`, and the node SHALL no longer participate in
+ *   network consensus.
+ * - The `DELETED` state SHALL be a terminal state. Node identifiers SHALL NOT
+ *   be re-used and SHALL NOT be "revived".
+ */
```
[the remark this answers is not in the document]
jsync-swirlds: The entire structure makes no sense if deletion has no status.  What you're describing is a non-functional half-state, which should not be specified.
What is the intention to manage deleted nodes; that is what we need to specify and we'll adjust that _if_ future phases choose a different approach.

services/address_book_service.proto
jsync-swirlds: Repeating my earlier response:
HIP text appears to claim a node operator can update their own node without council approval. Is council approval required for everything (in which case node operators are never actually required), or are node operators permitted to update their own nodes independently?

services/node_create.proto
jsync-swirlds: PEM is absolutely a format, and it's not identical to DER (although the two _can be_ combined).
DER is a binary format (ASN.1 "Distinguished Encoding Rules") for encoding data, often used for PKI certificates (of various forms).
PEM is an ASCII format (S/MIME "Privacy Enhanced Mail") that Base64 encodes binary data as ASCII text with specific delimiters.
When the two are combined (which is unambiguous here) as a "PEM encoded certificate" you get a PKI certificate encoded to binary and then Base64 encoded to ASCII text and stored in a file with specified delimiters.

Specifying DER here with a "PEM" file extension would leave us with an unusable binary file that cannot be properly processed as a PEM format file and might result in data corruption if the file is assumed to be actual PEM format and treated as 7-bit-safe ASCII.

iwsimon: "Precisely, this field is the DER encoding of gossip X509 certificate." This was Nathan's words to keep during our team meeting together. We parse .PEM file and get X509Certificate and store it's getEncoded() which is DER.

services/node_create.proto
jsync-swirlds: Context is already clear here, a ServiceEndpoint (the type of _this_ field) only holds IP addresses, and the wording "address and port" is the common way of differentiating from FQDN.

services/state/addressbook/node.proto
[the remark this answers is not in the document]
jsync-swirlds: Based on discussion, DELETED state is removed (the node will be removed entirely, although we'll still need to ensure node ID is never reused).
We still need the two `PENDING` status, without that we have no clear marker for what to change in the active configuration at the next upgrade.

This entity is internal network state (and part of the implementation details), so adding a status to enable managing the items modified but not yet active in network configuration is reasonable, in my opinion.

services/address_book_service.proto
iwsimon: From today's meeting, update don't need council approval.

services/node_create.proto
```diff
@@ -27,62 +25,100 @@ option java_multiple_files = true;
...
+     * This list MUST NOT contain more than `10` entries.<br/>
+     * The first entry in this list SHALL be the endpoint published to all
+     * consensus nodes.<br/>
+     * All other entries SHALL be reserved for future use.
+     * <p>
+     * Some networks restrict these endpoints further.<br/>
+     * A network may require that these endpoints must not specify a DNS name,
+     * may permit a DNS name, or may require a DNS name. The specific
+     * requirements for each network MUST be public, and SHALL apply to all
+     * interactions with the network address book for that network.
```
iwsimon, line 86: Don't want to loss these details:
The first in the list is used as the Internal IP address in config.txt,
 the second in the list is used as the External IP address in config.txt,
  the rest of IP addresses are ignored for DAB phase 2.

jsync-swirlds: Those are specific to our implementation, and would belong in a design document, but not here in the protobuf.
I can note the internal/external part (not config.txt, however) if we believe it will remain a _requirement_ for all implementations.

services/query.proto
```diff
@@ -200,8 +193,17 @@ message Query {
...
+         * This query is a privileged action that requires signature from the
+         * Hedera governing council.<br/>
+         * The response, when successful, SHALL contain the full information
+         * currently available, including current status, for the single
+         * consensus node identifier provided.<br/>
+         * The node queried MAY be active, inactive, deleted, or pending a
+         * change at the next network upgrade.
          */
-        NodeGetInfoQuery nodeGetInfo = 59;
+        com.hedera.hapi.node.addressbook.NodeGetInfoQuery nodeGetInfo = 59;
```
iwsimon, line 207: Don't need the whole package name here.

services/state/addressbook/node.proto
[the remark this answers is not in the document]
iwsimon: We should remove this enum NodeStatus. It's not used in phase 2. It was designed for phase 3 to deal with Roster. These comments do not make sense. We can add them later for phase 3 if needed.

services/node_create.proto
iwsimon: I will copy them to java file comments.

services/query.proto
jsync-swirlds: It is required now; as the package directive in the files is updated.

services/state/addressbook/node.proto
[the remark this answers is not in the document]
jsync-swirlds: Not sure how the comments don't make sense; they describe the behavior specified elsewhere, just specific to the status.
We aren't going to be able to implement, even phase 2, without some sort of status for pending items.
It might have been missed, but something has to track that a node is added in state but not written to configuration yet, and that something has to be _in_ state.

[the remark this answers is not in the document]
iwsimon: In phase 2, we use a store to track those hapi transactions and they are not active until next freeze update. Here is the design [diagram](https://lucid.app/lucidchart/01f7f0ae-a6e3-4e44-9768-9220936ecec6/edit?invitationId=inv_bb15f582-8437-496f-875c-3139d5bdadfe&page=m-5o7ONTd-nK#).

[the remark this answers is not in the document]
jsync-swirlds: From that diagram the "store" (which is a proxy for the merkle tree) is storing `Node` objects (exactly the message that we're defining here).
Nothing in that diagram there explains how the freeze upgrade process will know _which_ `Node` objects need to be _updated_, _added_, or _deleted_.
Pending status values (and one non-pending value) seem, to me, to be the most sensible mechanism for such indicator.

services/state/addressbook/node.proto
```diff
@@ -27,81 +25,202 @@ option java_package = "com.hederahashgraph.api.proto.java";
...
+     * The consensus weight of a node SHALL be calculated based on the amount
+     * of HBAR staked to that node.<br/>
+     * Consensus SHALL be calculated based on agreement of greater than `2/3`
+     * of the total `weight` value of all nodes on the network.
      */
     uint64 weight = 8;
 
-    /**
-     * an enum to indicate the status of the node, not used in phase 2
+    reserved "status";
```
netopyr, line 143: We should not define changes for an incomplete design.

services/state/addressbook/node.proto
```diff
@@ -27,81 +25,202 @@ option java_package = "com.hederahashgraph.api.proto.java";
...
+    /* [Placed AFTER reserved, and not doc-comment due to PBJ issue #217]
+     * Reserved as `NodeStatus status = 9;`
+     * A node status value.<br/>
+     * This value is not yet used and design is incomplete.
      */
-    //NodeStatus status = 9;
 
 }
 
+/**
```
netopyr, line 153: We should not define `NodeStatus` at this point. The design for the next phase is incomplete, and one can easily imagine a solution that does not need a `NodeStatus`.

netopyr (changes requested):
Please remove the `status` property and `NodeStatus`. The design for this functionality is incomplete and the usage of `status` and `NodeStatus` is not a given.

services/state/addressbook/node.proto
jsync-swirlds: What you are requesting is a non-functional design.  Without *something* we have no pending state and much of the rest of the design is impossible to implement.  That suggests that *this phase* design is incomplete, in which case we need more work and should consider the need for some form of status(or alternative) as part of the current phase.

services/transaction_body.proto
```diff
@@ -93,287 +89,296 @@ import "node_delete.proto";
...
+
+        /**
+         * Unpauses the Token
+         */
+        TokenUnpauseTransactionBody token_unpause = 47;
+
+        /**
+         * Update the metadata of one or more NFT's of a specific token type.
+         */
+        TokenUpdateNftsTransactionBody token_update_nfts = 53;
```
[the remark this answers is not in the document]
jsync-swirlds: This reorder moves the transactions for each service to be grouped together (so token service transactions are together, crypto service transaction are together, etc...); otherwise services end up scattered about, with some transactions for the same service quite far apart.

It could be returned to number order, but that, in my opinion, makes the specification less readable for an outside contributor.

Perhaps something to discuss more widely, as number order is something more helpful when _adding_ items, but service-grouped is more helpful when _finding_ items.

services/address_book_service.proto
```diff
@@ -29,37 +27,132 @@ import "transaction_response.proto";
...
+ *    - The Hedera council representative SHALL arrange for council members to
+ *      review and sign the transaction.
+ *    - Once sufficient council members have signed the transaction, the
+ *      Hedera council representative SHALL submit the transaction to the
+ *      network.
+ * - Upon receipt of a valid and signed node deletion transaction the network
+ *   software SHALL
+ *    - Validate the threshold signature for the Hedera governing council
+ *    - Validate the signature of the active `key` for the account assigned
+ *      as the "node account", _if present_.
```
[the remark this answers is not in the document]
jsync-swirlds: Validate the signature on the transaction?
Not sure what isn't possible here...

services/basic_types.proto
```diff
@@ -1385,9 +1385,9 @@ message ServiceEndpoint {
 /**
  * The data about a node, including its service endpoints and the Hedera account to be paid for
  * services provided by the node (that is, queries answered and transactions submitted.)
- * 
- * If the <tt>serviceEndpoint</tt> list is not set, or empty, then the endpoint given by the
- * (deprecated) <tt>ipAddress</tt> and <tt>portno</tt> fields should be used.
+ *
+ * If the `serviceEndpoint` list is not set, or empty, then the endpoint given by the
```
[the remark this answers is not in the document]
jsync-swirlds: I believe so, as this is a field name in the same message, so it's a code reference.
The guideline I'm following is to use inline code styling (\`) for references to field or message names.

services/basic_types.proto
```diff
@@ -1364,19 +1364,19 @@ message CurrentAndNextFeeSchedule {
...
     /**
-     * The port of the node
+     * The port of the service endpoint
      */
     int32 port = 2;
 
     /**
-     * A node domain name
-     * This MUST be the fully qualified domain name of the node.
+     * A node domain name.<br/>
```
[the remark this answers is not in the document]
jsync-swirlds: To add a line break after the sentence (otherwise markdown will merge the lines).
It could be a `<p>` tag (technically `<p/>` is more correct, but that confuses javadoc), and in future (much more thorough edits) that is under consideration for the standard separation between description and constraint.

services/state/addressbook/node.proto
jsync-swirlds: Note, if we have a solid design that will definitely work _as specified_ for phase 2; I have no issue to remove this pre-existing definition entirely. If we intend to keep this in the protobuf definitions, however, it needs to be well specified.

Side note. If we have a design that works but is _different_ from specification, then, in my opinion, we need to **publicly document** that design and then we can update specification accordingly.

services/state/addressbook/node.proto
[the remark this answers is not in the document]
jsync-swirlds: Updated based on discussion.

services/state/addressbook/node.proto
jsync-swirlds: Status field and enum are removed.

services/node_create.proto
```diff
@@ -27,62 +25,111 @@ option java_multiple_files = true;
...
+     * all consensus nodes.<br/>
+     * All other entries SHALL be reserved for future use.
+     * <p>
+     * Each network may have additional requirements for these endpoints.
+     * A client MUST check network-specific documentation for those
+     * details.<br/>
+     * <blockquote>Example<blockquote>
+     * Hedera Mainnet _requires_ that address be specified, and does not
+     * permit DNS name (FQDN) to be specified.<br/>
+     * Mainnet also requires that the first entry be an "internal" IP
```
edward-swirldslabs, line 87: I don't know if this impacts what is written here, in the code the effective difference between "internal" and "external" is the order in which the connection attempts are made.  If the "internal" address resolves, we attempt to connect to that address first.  Once we start having consensus rosters, the generalized treatment will be to attempt connection in order of the list and not make a categorical discrepancy between "internal" and "external".

services/node_create.proto
```diff
@@ -27,62 +25,111 @@ option java_multiple_files = true;
...
     /**
-     * The node's X509 certificate used to sign stream files (e.g., record stream
-     * files). Precisely, this field is the DER encoding of gossip X509 certificate.
-     * This is a mandatory field.
+     * A certificate used to sign gossip events.
+     * <p>
+     * This value MUST be a certificate of a type permitted for gossip
+     * signatures.<br/>
+     * This value MUST be the DER encoding of the certificate presented.<br/>
+     * This field is REQUIRED and MUST NOT be empty.
```
edward-swirldslabs, line 119: Right now I believe all signing certs are self-signed.  I don't know if the self-signed property makes a difference on its ability to be used in the trust store for the agreement keys in mutual TLS.  We will be automatically rendering the mTLS agreement keys on boot with the signing cert as a parent certificate.  If the self-signed property is necessary for it to function as the parent cert of the mTLS key, then this should be documented here.

edward-swirldslabs (approved):

services/node_create.proto
jsync-swirlds: Using a self-signed signing certificate as parent for the mTLS agreement key setup is neither required nor disbarred, it is just a matter of how you establish trust of the certificate and what authenticity guarantees the network is willing to accept.

That's just standard mTLS considerations.

services/node_create.proto
jsync-swirlds: That will be, as I understand it, something visible in config.txt; and does not affect what is stored here in this phase.
Each `Node` from the in-state store will become an entry in `config.txt` as part of the freeze/upgrade process, and certificates (with their public keys) will be written to PEM files on disk.
As to later, the network address book (as distinct from the internal platform concept of the "roster") will need to track more than just consensus nodes, and will be used by more than just consensus nodes. Changes and/or adjustments needed for those later requirements are deferred to later consideration (c.f. Michael Heinrichs' comments above).

services/address_book_service.proto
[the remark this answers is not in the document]
netopyr: I am not sure if validating a key _if present_ is possible. We can define the account key as optional, but this will only tell us whether there is a valid signature or not. We cannot distinguish between an invalid signature and a missing signature. I believe this distinction would be required here.

services/basic_types.proto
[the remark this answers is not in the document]
netopyr: I find the mixture of html and markdown for formatting confusing. But if this is the standard we want to follow...

netopyr (commented):
Please note that we have to pull merged changes for the dynamic address book out of `main` and keep them in a branch. They need to be brought back once we are ready to release the functionality.

services/address_book_service.proto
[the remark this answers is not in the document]
jsync-swirlds: If the signature is present, we can certainly validate that, and if not present not validate the signature.  I think I see your point; the wording is a bit confusing, then again the process is a bit confusing too (because it might be initiated in two different ways).
I'll rework the wording a bit to make the necessary actions a bit clearer.

services/basic_types.proto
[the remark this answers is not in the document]
jsync-swirlds: It's all markdown; it's just that base markdown has two large gaps normally filled with html tags.
1. Standard markdown has inconsistent support for a line break.  It adds a paragraph break for each blank line, but single line break is recommended to use `<br/>` for best compatibility.
1. Markdown requires the use of HTML tags for anything multi-line in tables because tables are an afterthought in markdown so a table row cannot contain a newline.
What the processor does is remove all line breaks in these comment blocks before building the table row, so we have to use tags a bit more in field comments than in message comments.
The main alternative would be to use html exclusively, but that's much harder to read, in my opinion, than markdown.

jsync-swirlds:
> Please note that we have to pull merged changes for the dynamic address book out of `main` and keep them in a branch. We need to bring them back once we are ready to release the functionality. We should retarget this PR against this branch once it was created.

Why would we remove these from `main`?  This is an API definition, the definition is complete for phase 2, and there is no risk to publishing the API prior to completing the code design.

services/address_book_service.proto
```diff
@@ -29,37 +27,132 @@ import "transaction_response.proto";
...
+ *      review and sign the transaction.
+ *    - Once sufficient council members have signed the transaction, the
+ *      Hedera council representative SHALL submit the transaction to the
+ *      network.
+ * - Upon receipt of a valid and signed node creation transaction the network
+ *   software SHALL
+ *    - Validate the threshold signature for the Hedera governing council
+ *    - Validate the signature of the active `key` for the account to be
+ *      assigned as the "node account".
+ *    - Create the new node in a pending state.
```
iwsimon, line 55: Maybe remove '"pending state' here, same below 'transition the new pending node to an active status'? We can say "the new node is not active until next freeze updrade' or something like that.

jsync-swirlds: Updated here, and also for the similar items in update and delete.

iwsimon (dismissed):
LGTM. Thanks @jsync-swirlds

jsync-swirlds:
Note, merge for this PR is delayed to NET 2024-05-31 to accommodate a requirement for release `0.51`.

Mark-Swirlds (commented):
(small comment) Can we change the name of node_create to node_add?  I understand the CRUD structure, but node_add is more descriptive of the task.

jsync-swirlds:
> (small comment) Can we change the name of node_create to node_add? I understand the CRUD structure, but node_add is more descriptive of the task.

1. @iwsimon would be the best individual to do that in the PR to add all these files back into the protobuf repo (everything related to DAB was removed from both protobuf and services codebase on Wednesday).
1. The HIP would need to be rewritten to change all the `create` to `add` wording.
1. That would be inconsistent with every other service use of `create` for adding or creating an entity
1. It is creating an entity in state, so `create` seems entirely reasonable to my mind.

jsync-swirlds:
Updated following the removal and re-add of all address book changes.
That bounce required squashing these commits to make rebase sane.

jsync-swirlds:
> Please remove the `status` property and `NodeStatus`. The design for this functionality is incomplete and the usage of `status` and `NodeStatus` is not a given.

@netopyr This is done, along with matching the most recent discussions and updates.  Please re-review.

netopyr (commented):
LGTM - thanks @jsync-swirlds

netopyr (approved):

JeffreyDallas (approved):

povolev15 (approved):
LGTM

PR #347, opened 2024-05-06
Added NodeGetInfo

iwsimon:
**Description**:

Added NodeGetInfo to HederaFunctionality, added NodeGetInfoQuery to query.proto

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1206,6 +1206,11 @@ enum HederaFunctionality {
      * Delete a node
      */
     NodeDelete = 91;
+
+    /**
+   * Get Node information
+   */
+    NodeGetInfo = 92;
 }
 
 /**
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1208,8 +1208,8 @@ enum HederaFunctionality {
     NodeDelete = 91;
 
     /**
-   * Get Node information
-   */
+     * Get Node information
+     */
     NodeGetInfo = 92;
 }
 
--- a/services/query.proto
+++ b/services/query.proto
@@ -60,6 +60,7 @@ import "token_get_nft_info.proto";
 import "token_get_nft_infos.proto";
 
 import "get_account_details.proto";
+import "node_get_info.proto";
 
 /**
  * A single query, which is sent from the client to a node. This includes all possible queries. Each
@@ -197,5 +198,10 @@ message Query {
          * Gets all information about an account including allowances granted by the account
          */
         GetAccountDetailsQuery accountDetails = 58;
+
+        /**
+         * Get all information about a node
+         */
+        NodeGetInfoQuery nodeGetInfo = 59;
     }
 }


services/basic_types.proto
```diff
@@ -1206,6 +1206,11 @@ enum HederaFunctionality {
      * Delete a node
      */
     NodeDelete = 91;
+
+    /**
+   * Get Node information
+   */
```
jsync-swirlds, line 1212: Indentation is slightly off here.

jsync-swirlds (commented):
Just a slight error in spacing.

services/basic_types.proto
iwsimon: Updated Indentation.

jsync-swirlds (approved):

jsync-swirlds (approved):

jsync-swirlds (approved):

Neeharika-Sompalli (approved):
LGTM! Thanks @iwsimon

steven-sheehy (approved):

services/query.proto
```diff
@@ -197,5 +198,10 @@ message Query {
          * Gets all information about an account including allowances granted by the account
          */
         GetAccountDetailsQuery accountDetails = 58;
+
+        /**
+         * Get all information about a node
+         */
+        NodeGetInfoQuery nodeGetInfo = 59;
```
jsync-swirlds, line 205: Found post merge, this doesn't compile (because we used bad package names in these legacy files)...
fix:
```suggestion
        com.hedera.hapi.node.addressbook.NodeGetInfoQuery nodeGetInfo = 59;
```

iwsimon: I have another PR in hedera-service to compile all the protobuf changes. This change matches others. We do have import "node_get_info.proto"; at the top.

jsync-swirlds: It fails to compile as a protobuf.
No idea why your services PR works (might be a bug in PBJ), but it doesn't compile in protoc.

mhess-swl: @jsync-swirlds that's bizarre... 🤯

iwsimon: You can see this [PR](https://github.com/hashgraph/hedera-services/pull/13090) is compiled. Query.java and NodeGetInfoQuery.java were generated correctly.

jsync-swirlds: PBJ doesn't really used a lot of the directives in the proto file as it only handles Java, and there is much simplification (which sometimes leads to bugs or overly-relaxed compile as it does here).
protoc is strict about all of the directives, including the `package` keyword (which it treats as a C++ namespace for the messages, which has interesting semantics).
protoc is used for everything except the consensus node software, so we have to continue working with protoc.

iwsimon: Is this a new guideline for services?

jsync-swirlds: We have always required protoc, it's what builds all our SDKs.
It's just not always been well communicated.

jsync-swirlds: Disregard.

There were a couple cross-merged branches that caused a false error.
The issue has been fixed.

jsync-swirlds: To item (3):
Services should not run protoc.
This protobuf repo should, however.  That's the more critical missing piece, there are no CI checks in this repo, so it's far too easy to accidentally commit a change here that doesn't compile everywhere it needs to (i.e. Javascript, Go, C#, C++, Rust, Java, etc...).

When you build in services, it runs PBJ, no protoc.  Sometimes PBJ will succeed when protoc will not, and it only generates the java sources (not all the other languages).  Again, not a services issue, but an issue to resolve in _this_ repo.

PR #351, opened 2024-05-13
feat: add token reject transaction

MiroslavGatsanoga:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/services/response_code.proto
+++ b/services/response_code.proto
@@ -1437,4 +1437,10 @@ enum ResponseCodeEnum {
    * grpc_certificate_hash is invalid
    */
   INVALID_GRPC_CERTIFICATE = 345;
+
+  /**
+   * The transaction attempted to use duplicate `TokenReference`.<br/>
+   * This affects `TokenReject` attempting to reject same token reference more than once.
+   */
+  TOKEN_REFERENCE_REPEATED = 346;
 }
--- a/services/response_code.proto
+++ b/services/response_code.proto
@@ -1449,4 +1449,9 @@ enum ResponseCodeEnum {
    * This affects `TokenReject` attempting to reject same token reference more than once.
    */
   TOKEN_REFERENCE_REPEATED = 347;
+
+  /**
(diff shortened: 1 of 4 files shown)


jsync-swirlds (approved):
Looks good.
Would be nice to see _comments_ word-wrap at column 80, but that's a nitpick.

Neeharika-Sompalli (approved):
LGTM! For `0.52` release changes

services/response_code.proto
```diff
@@ -1437,4 +1437,10 @@ enum ResponseCodeEnum {
    * grpc_certificate_hash is invalid
    */
   INVALID_GRPC_CERTIFICATE = 345;
+
+  /**
+   * The transaction attempted to use duplicate `TokenReference`.<br/>
+   * This affects `TokenReject` attempting to reject same token reference more than once.
+   */
+  TOKEN_REFERENCE_REPEATED = 346;
```
MrValioBg, line 1445: @Neeharika-Sompalli suggested [here](https://github.com/hashgraph/hedera-services/pull/13117#discussion_r1598711139) we could also add INVALID_OWNER_ID as status.

Maybe we could consider such a status, since for approve allowance we return INVALID_ALLOWANCE_SPENDER_ID instead of INVALID_ACCOUNT_ID

What are your thoughts on that ?

cc: @Nana-EC @MiroslavGatsanoga

MiroslavGatsanoga: I was thinking we can use the existing `INVALID_ACCOUNT_ID` for `TokenReject` because we have only one field (`owner`) of type `AccountID` i.e. it's not ambiguous.

Nana-EC: @MiroslavGatsanoga I like Neeha's suggestion as it's more clear on reading and to someone who's unfamiliar with HAPI it immediately says what the issue is

MiroslavGatsanoga: Added.

PR #352, opened 2024-05-15
Reserve removed fields to preserve SDK compatibility

jsync-swirlds:
Several fields were removed from protobuf without reserving the field numbers.
Some of these removals are temporary, others are permanent.
Here we reserve the field numbers for those removed fields to ensure the field numbers are not accidentally reused, breaking compatibility.

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -111,6 +111,10 @@ message AccountID {
          */
         bytes alias = 4;
     }
+
+    // Removed evm_address February 2023 to compress HIP 583/631
+    //     into the single immutable alias field.
+    reserved 5;
 }
 
 /**
--- a/services/crypto_create.proto
+++ b/services/crypto_create.proto
@@ -46,7 +46,7 @@ import "duration.proto";
  * used to extend its expiration as long as possible. If it is has a zero balance when it expires,
  * then it is deleted. This transaction must be signed by the payer account. If receiverSigRequired
  * is false, then the transaction does not have to be signed by the keys in the keys field. If it is
- * true, then it must be signed by them, in addition to the keys of the payer account. If the 
+ * true, then it must be signed by them, in addition to the keys of the payer account. If the
  * auto_renew_account field is set, the key of the referenced account must sign.
  *
  * An entity (account, file, or smart contract instance) must be created in a particular realm. If
@@ -62,6 +62,13 @@ import "duration.proto";
  * multiple shards.
  */
 message CryptoCreateTransactionBody {
+    // Removed auto_renew_account_id "temporarily" in January 2023.
+    // Removed evm_address February 2023 to compress HIP 583/631
+    //     into the single immutable alias field.
+    reserved 19,20;
+    // Removed prior to oldest available history
+    reserved 4,5;
+
     /**
      * The key that must sign each transfer out of the account. If receiverSigRequired is true, then
      * it must also sign any transfer into the account.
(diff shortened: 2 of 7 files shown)


services/basic_types.proto
```diff
@@ -111,6 +111,10 @@ message AccountID {
          */
         bytes alias = 4;
     }
+
+    // Removed evm_address February 2023 to compress HIP 583/631
+    //     into the single immutable alias field.
```
mhess-swl, line 116: This gap looks intentional, yes?

mhess-swl (approved):
LGTM. Thanks @jsync-swirlds !

services/basic_types.proto
jsync-swirlds: Yes, it's there to improve readability.

steven-sheehy (commented):
Why do we need to reserve fields for HIPs that never made it to any environment? HIPs 583/631 virtual addresses was removed in favor of a different approach and is not expected to come back. Likewise the autorenew fields never made it to production. The only reason we would need to reserve them is if there is data in any bucket that might have these fields present, which is not the case. If this functionality is ever resurrected we can assign them whatever is the newest field number at the time.

jsync-swirlds:
> Why do we need to reserve fields for HIPs that never made it to any environment? HIPs 583/631 virtual addresses was removed in favor of a different approach and is not expected to come back. Likewise the autorenew fields never made it to production. The only reason we would need to reserve them is if there is data in any bucket that might have these fields present, which is not the case. If this functionality is ever resurrected we can assign them whatever is the newest field number at the time.

Those versions were published on the main branch, so external code may have been built with them, and we cannot, as far as I know, guarantee nothing was serialized with those definitions.  Once anything is serialized with a definition, the field number _must not_ be reused with a different type definition, it breaks compatibility.

If we have clear evidence that there is _no chance_ that the definition with those fields included was ever used, we can drop the reserved, but currently I am not aware of such.  The safe option, then, is to reserve the field numbers and either restore the fields exactly as originally defined, or just not use those numbers going forward.

Reserving a field number doesn't hurt us in any significant manner, and avoids any possibility of serialization errors due to reusing field numbers with a different definition.

steven-sheehy (changes requested):
> If we have clear evidence that there is no chance that the definition with those fields included was ever used, we can drop the reserved, but currently I am not aware of such. The safe option, then, is to reserve the field numbers and either restore the fields exactly as originally defined, or just not use those numbers going forward.

We do know 100% that it didn't get serialized and make its way into any bucket.

> Reserving a field number doesn't hurt us in any significant manner, and avoids any possibility of serialization errors due to reusing field numbers with a different definition.

It bloats the protobuf definition unnecessarily and somewhat defeats the purpose of removing those changes before they made it into an artifact. It may not functionally cause any harm but it makes the API less clear and concise. As such I am against this change without explicit proof this is causing customer's any harm.

jsync-swirlds:
> > If we have clear evidence that there is no chance that the definition with those fields included was ever used, we can drop the reserved, but currently I am not aware of such. The safe option, then, is to reserve the field numbers and either restore the fields exactly as originally defined, or just not use those numbers going forward.
> 
> We do know 100% that it didn't get serialized and make its way into any bucket.

If we're certain these changes never made it into a third-party SDK nor were ever used submit transactions ("buckets" are only part of the picture, not everything), then I'm fine with documenting that and moving on.  The purpose of this PR is that we had a _public_ version (which means we don't control if or how it's used) of the codebase that included fields that were later removed without any clear documentation of what was done or why (and when discovered, nobody involved could remember why things were removed).

I'll close this PR based on your statements, and document the original removal PR with the claims presented.

jsync-swirlds:
As claimed above, that despite being on the main branch for 4-6 months, the versions of these proto files with the removed fields were never used in any deployed software (regardless of whether the individual fields were filled in).

Closing as unnecessary based on that assertion.

PR #354, opened 2024-05-17
feat: add TokenRejectTransactionBody to transaction_body proto

MrValioBg:
**Description**:
This PR adds missing TokenRejectTransactionBody to transaction_body proto

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/services/transaction_body.proto
+++ b/services/transaction_body.proto
@@ -378,7 +378,7 @@ message TransactionBody {
     NodeDeleteTransactionBody nodeDelete = 56;
 
     /**
-     * A transaction body for a `rejectToken` request.
+     * Rejects undesired tokens from an account by returning them to the treasury.
      */
     TokenRejectTransactionBody tokenReject = 57;
   }
--- a/services/transaction_body.proto
+++ b/services/transaction_body.proto
@@ -378,7 +378,21 @@ message TransactionBody {
     NodeDeleteTransactionBody nodeDelete = 56;
 
     /**
-     * Rejects undesired tokens from an account by returning them to the treasury.
+     * A transaction body to "reject" undesired tokens.<br/>
+     * This transaction will transfer one or more tokens or token
+     * balances held by the requesting account to the treasury
+     * for each token type.
+     * <p>
+     * Each transfer MUST be one of the following:
+     * <ul>
+     *   <li>A single non-fungible/unique token.</li>
+     *   <li>The full balance held for a fungible/common
+     *       token type.</li>
+     * </ul>
+     * When complete, the requesting account SHALL NOT hold the
+     * rejected tokens.<br/>
+     * Custom fees and royalties defined for the tokens rejected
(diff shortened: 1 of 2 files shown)


services/transaction_body.proto
```diff
@@ -375,5 +376,10 @@ message TransactionBody {
      * A transaction body for a `deleteNode` request.
      */
     NodeDeleteTransactionBody nodeDelete = 56;
+
+    /**
+     * A transaction body for a `rejectToken` request.
```
jsync-swirlds, line 381: This specification comment is not particularly helpful; it merely restates the type of a variable in more words.
A better option (which may still not be perfect, but is much more useful):
```suggestion
     * A transaction body to "reject" undesired tokens.<br/>
     * This transaction will transfer one or more tokens or token
     * balances held by the requesting account to the treasury
     * for each token type.
     * <p>
     * Each transfer MUST be one of the following:
     * <ul>
     *   <li>A single non-fungible/unique token.</li>
     *   <li>The full balance held for a fungible/common
     *       token type.</li>
     * </ul>
     * When complete, the requesting account SHALL NOT hold the
     * rejected tokens.<br/>
     * Custom fees and royalties defined for the tokens rejected
     * SHALL NOT be charged for this transaction.
```

services/transaction_body.proto
```diff
@@ -375,5 +376,10 @@ message TransactionBody {
...
+     */
+    TokenRejectTransactionBody tokenReject = 57;
```
Neeharika-Sompalli, line 383: You might need to add to `schedule_transaction_body.proto` and also for `HederaFunctionality`

MrValioBg: I've added it to schedule_transaction_body.proto.  🙂

For HederaFunctionality - its already [included in basic_types.proto](https://github.com/hashgraph/hedera-protobufs/pull/351/files#diff-636b14bb9e5952ee6f3253a1bba9a76bde7e8a9a27a78c0eda2d6303ebb3bde6R1219).

services/transaction_body.proto
MrValioBg: I've adjusted it, but I haven't included this whole comment, as the whole description already exists in the token_reject.proto itself, and all the comments for other transaction bodies in transaction_body.proto are short and concise.

If you think it's still better to have the full description here and also in the TokenRejectTransactionBody itself (token_reject.proto), I could add it. 🙂

jsync-swirlds: We should have the more full description.  The short (minimal, really) descriptions currently present in most cases in this file are all scheduled to be replaced with thorough _specification_.

Neeharika-Sompalli (approved):
LGTM!

services/transaction_body.proto
MrValioBg: Got it. I've adjusted it.

jsync-swirlds (approved):
LGTM.

PR #358, opened 2024-05-28
feat: Back out protobuf-changes to dynamic address book for release 0.51

### LINKED ISSUE #357: Back out protobuf-changes related to dynamic address book for release 0.51
### Problem

Back out protobuf-changes related to dynamic address book for release 0.51

### Solution

Back out protobuf-changes related to dynamic address book for release 0.51

### Alternatives

_No response_

iwsimon:
**Description**:

**Related issue(s)**:

Fixes #357


services/response_code.proto
```diff
@@ -1398,49 +1398,9 @@ enum ResponseCodeEnum {
...
-  /**
-   * The maximum automatic associations value is not valid.<br/>
-   * The most common cause for this error is a value less than `-1`.
-   */
-  INVALID_MAX_AUTO_ASSOCIATIONS = 346;
+    /**
+     * The maximum automatic associations value is not valid.<br/>
+     * The most common cause for this error is a value less than `-1`.
+     */
+    INVALID_MAX_AUTO_ASSOCIATIONS = 338;
```
jsync-swirlds, line 1405: I would strongly recommend we keep the number `346` here.  Changing the enumeration number for any published enum value can break a lot of things.

jsync-swirlds (commented):
Just one item that probably shouldn't change.

services/response_code.proto
iwsimon: I don't know if we should skip the number here. I was planning to add back rest of node code after this. INVALID_MAX_AUTO_ASSOCIATIONS is added after 0.50 release, ie. it will be in 0.51. I think we should be ok here.

jsync-swirlds: We should not issue a release with a number that we plan to change.  We'll be _changing_ this unrelated field to `338`, then possibly _changing_ back to `446` (or else we break a bunch of stuff that's building on the current `main` version by changing all the DAB numbers).

Skipping numbers is pretty common all through the current protobuf (because we have fields and enums that were removed historically without reserving the numbers), so that should be fine.

kimbor (approved):
lgtm!

services/response_code.proto
iwsimon: Changed it back to 346, original value.

kimbor (approved):

jsync-swirlds (approved):

povolev15 (approved):

Neeharika-Sompalli (approved):

PR #367, opened 2024-06-06
HIP 904 proto updates for token reject

MiroslavGatsanoga:
**Description**:

Adding protobuf updates related to TokenReject transaction implementation planned for 0.52.

**Notes for reviewer**:

Merge this PR after https://github.com/hashgraph/hedera-services/pull/13685, otherwise the build in Services `develop` branch will fail.

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [ ] Tested (unit, integration, etc.)

PR #371, opened 2024-06-10
feat: Protobuf changes for registering airdrop

ibankov:
**Description**:

This PR modifies the protobufs in order to support creating Airdrops

**Related issue(s)**:

https://github.com/hashgraph/hedera-services/issues/13698

Fixes #

**Notes for reviewer**:

Merge this PR after https://github.com/hashgraph/hedera-protobufs/pull/367 is merged in main and this branch is synced with main, otherwise the build in Services develop branch will fail.

**Checklist**

### DIFF
--- a/services/basic_types.proto
+++ b/services/basic_types.proto
@@ -1211,6 +1211,11 @@ enum HederaFunctionality {
      * Transfer one or more token balances held by the requesting account to the treasury for each token type.
      */
     TokenReject = 92;
+
+    /**
+     * Airdrop one or more tokens to one or more accounts.
+     */
+    TokenAirdrop = 93;
 }
 
 /**
@@ -1668,4 +1673,70 @@ message StakingInfo {
          */
         int64 staked_node_id = 6;
     }
+}
+
+/**
+ * A unique, composite, identifier for a pending airdrop.
+ *
+ * Each pending airdrop SHALL be uniquely identified by a PendingAirdropId.
+ * A PendingAirdropId SHALL be recorded when created and MUST be provided in any transaction
+ * that would modify that pending airdrop (such as a `claimAirdrop` or `cancelAirdrop`).
+ */
+message PendingAirdropId {
+    /**
+     * A sending account.<br/>
+     * This is the account that initiated, and SHALL fund, this pending airdrop.<br/>
+     * This field is REQUIRED.
+     */
+    AccountID sender_id = 1;
+
+    /**
+     * A receiving account.<br/>
+     * This is the ID of the account that SHALL receive the airdrop.<br/>
+     * This field is REQUIRED.
+     */
+    AccountID receiver_id = 2;
+
+    oneof token_reference {
+        /**
+         * A token ID.<br/>
+         * This is the type of token for a fungible/common token airdrop.<br/>
+         * This field is REQUIRED for a fungible/common token and MUST NOT be used for a
+         * non-fungible/unique token.
+         */
+        TokenID fungible_token_type = 3;
+
+        /**
+         * The id of a single NFT, consisting of a Token ID and serial number.<br/>
+         * This is the type of token for a non-fungible/unique token airdrop.<br/>
+         * This field is REQUIRED for a non-fungible/unique token and MUST NOT be used for a
+         * fungible/common token.
+         */
+        NftID non_fungible_token = 4;
+    }
+}
+
+/**
+ * A single pending airdrop value.
+ *
+ * This message SHALL record the airdrop amount for a fungible/common token.<br/>
+ * This message SHOULD be null for a non-fungible/unique token.<br/>
+ * If a non-null `PendingAirdropValue` is set for a non-fungible/unique token, the amount
+ * field MUST be `0`.
+ *
+ * It is RECOMMENDED that implementations store pending airdrop information as a key-value map
+ * from `PendingAirdropId` to `PendingAirdropValue`, with a `null` value used for non-fungible
+ * pending airdrops.
+ */
+message PendingAirdropValue {
+    /**
+     * An amount to transfer for fungible/common tokens.<br/>
+     * This is expressed in the smallest available units for that token
+     * (i.e. 10<sup>-`decimals`</sup> whole tokens).<br/>
+     * This amount SHALL be transferred from the sender to the receiver, if claimed.<br/>
+     * If the token is a fungible/common token, this value MUST be strictly greater than `0`.
+     * If the token is a non-fungible/unique token, this message SHOULD NOT be set, and if
+     * set, this field MUST be `0`.
+     */
+    uint64 amount = 1;
 }
--- a/services/token_airdrop.proto
+++ b/services/token_airdrop.proto
@@ -1,0 +1,104 @@
+/**
+ * # Token Airdrop
+ * Messages used to implement a transaction to "airdrop" tokens.<br/>
+ * An "airdrop" is a distribution of tokens from a funding account
+ * to one or more recipient accounts, ideally with no action required
+ * by the recipient account(s).
+ *
+ * ### Keywords
+ * The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+ * "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
+ * document are to be interpreted as described in [RFC2119](https://www.ietf.org/rfc/rfc2119).
+ */
+syntax = "proto3";
+
+package proto;
+
+/*-
+ * ‌
+ * Hedera Network Services Protobuf
+ * ​
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ * ​
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * ‍
+ */
+
+option java_package = "com.hederahashgraph.api.proto.java";
+// <<<pbj.java_package = "com.hedera.hapi.node.token">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+import "basic_types.proto";
+
+/**
+ * Airdrop one or more tokens to one or more accounts.
+ *
+ * ### Effects
+ * This distributes tokens from the balance of one or more sending account(s) to the balance
+ * of one or more recipient accounts. Accounts MAY receive the tokens in one of four ways.
+ *
+ *  - An account already associated to the token to be distributed SHALL receive the
+ *    airdropped tokens immediately to the recipient account balance.<br/>
+ *    The fee for this transfer SHALL include the transfer, the airdrop fee, and any custom fees.
+ *  - An account with available automatic association slots SHALL be automatically
+ *    associated to the token, and SHALL immediately receive the airdropped tokens to the
+ *    recipient account balance.<br/>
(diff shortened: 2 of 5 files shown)


vtronkov (approved):

Nana-EC (approved):

services/basic_types.proto
```diff
@@ -1216,6 +1216,11 @@ enum HederaFunctionality {
      * Transfer one or more token balances held by the requesting account to the treasury for each token type.
      */
     TokenReject = 93;
+
+    /**
+     * Airdrops one or more tokens to one or more accounts.
```
jsync-swirlds, line 1221:
```suggestion
     * Airdrop one or more tokens to one or more accounts.
```

services/token_airdrop.proto
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+option java_package = "com.hederahashgraph.api.proto.java";
+// <<<pbj.java_package = "com.hedera.hapi.node.token">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+import "basic_types.proto";
+
+/**
+ * Airdrops one or more tokens to one or more accounts.
```
jsync-swirlds, line 44:
```suggestion
 * Airdrop one or more tokens to one or more accounts.
```

services/token_airdrop.proto
```diff
@@ -0,0 +1,104 @@
...
+ *    associated to the token, and SHALL immediately receive the airdropped tokens to the
+ *    recipient account balance.<br/>
+ *    The fee for this transfer SHALL include the transfer, the association, the cost to renew
+ *    that association once, the airdrop fee, and any custom fees.
+ *  - An account with "receiver signature required" set SHALL have a "Pending Airdrop" created
+ *    and must claim that airdrop with a `claimAirdrop` transaction.<br/>
+ *    The fee for this transfer SHALL include the transfer, the association, the cost to renew
+ *    that association once, the airdrop fee, and any custom fees. If the pending airdrop is not
+ *    claimed immediately, the `sender` SHALL pay the cost to renew the token association, and
+ *    the cost to maintain the pending airdrop, until the pending airdrop is claimed or cancelled.
```
jsync-swirlds, line 63: I do not think we create the association until the "claim" transaction.
```suggestion
 *    The fee for this transfer SHALL include the transfer, the airdrop fee,
 *    and any custom fees. If the pending airdrop is not claimed immediately,
 *    the `sender` SHALL pay the cost to maintain the pending airdrop, until
 *    the pending airdrop is claimed or cancelled.
```

services/token_airdrop.proto
```diff
@@ -0,0 +1,104 @@
...
+ *  - An account with no available automatic association slots SHALL have a "Pending Airdrop"
+ *    created and must claim that airdrop with a `claimAirdrop` transaction.<br/>
+ *    The fee for this transfer SHALL include the transfer, the association, the cost to renew
+ *    that association once, the airdrop fee, and any custom fees. If the pending airdrop is not
+ *    claimed immediately, the `sender` SHALL pay the cost to renew the token association, and
+ *    the cost to maintain the pending airdrop, until the pending airdrop is claimed or cancelled.
```
jsync-swirlds, line 69: Same here, I do not think we create or charge for the association until the
airdrop is claimed.
```suggestion
 *  - An account with no available automatic association slots SHALL have a
 *    "Pending Airdrop" created and must claim that airdrop with a
 *    `claimAirdrop` transaction.<br/>
 *    The fee for this transfer SHALL include the transfer, the airdrop fee,
 *    and any custom fees. If the pending airdrop is not claimed immediately,
 *    the `sender` SHALL pay the cost to maintain the pending airdrop, until
 *    the pending airdrop is claimed or cancelled.
```

services/token_airdrop.proto
```diff
@@ -0,0 +1,104 @@
...
+ * for the same sender, receiver, and token already exists, the existing pending airdrop
+ * SHALL be updated to add the new amount to the existing airdrop, rather than creating a new
+ * pending airdrop.
+ *
+ * Any airdrop that completes immediately SHALL be irreversible. Any airdrop that results in a
+ * "Pending Airdrop" MAY be canceled via a `cancelAirdrop` transaction.
+ *
+ * All transfer fees (including custom fees and royalties), as well as the rent cost for the
+ * first auto-renewal period for any automatic-association slot occupied by the airdropped
+ * tokens, SHALL be charged to the account paying for this transaction.
```
jsync-swirlds, line 81: It isn't clear how the automatic-association slot and renewal fees will be charged
to the payer currently.  Something to bring up in the design discussions, perhaps.

jsync-swirlds (commented):
Some small wording updates, and possible corrections for what fees are charged at the time of the airdrop transaction versus the time of the claim transaction.

services/token_airdrop.proto
IvanKavaldzhiev: The association fee should be pre-paid as per the HIP specifications.

services/basic_types.proto
ibankov: Fixed

services/token_airdrop.proto
ibankov: Fixed

services/token_airdrop.proto
IvanKavaldzhiev: If the recipient has available auto association slots, they will be used directly during an airdrop. In this case the payer of the airdrop will pay:
- the usage of this auto association
- the first rental auto-renew period for the token

In this case the token will be directly transferred.

If the recipient has no associations with token and no available auto association slots, then the payer of the airdrop will pay:
- prepayment for the association
- the first rental auto-renew period for the token
- the fee for transfer operation
- the airdrop fee itself
- all custom fees required to be paid by the sender/payer

In this case the token will end up in pending state.

services/token_airdrop.proto
jsync-swirlds: I believe the specification is flawed here, as we aren't creating an association yet (that's been discussed), and we shouldn't charge for something we haven't done yet.
We really don't have any way, technically, to accomplish what's written, as I understand things.

services/token_airdrop.proto
jsync-swirlds: I am aware of how the association is intended.  What I meant is that we have no mechanism, currently, to charge the payer for the renewal fee, as we cannot "store" fees, and we have nowhere in state to record that a renewal fee is "pre-paid".
I think that there is a design gap here, and I do not believe that we _can_ do this under current design and structure.

services/token_airdrop.proto
jsync-swirlds: There is no mechanism to do so.  We could charge, surely, but we have no way to record that this fee (or any fee) is pre-paid.  We cannot simply extend expiration, as that value is limited to a single renewal duration.

Also, we don't actually know, until claim, that an association will be needed.  Charging for a "potential" event is asking for problems as we charge fees for things that never happen.
In my opinion, timing assumptions for these "pre payment" items are significantly flawed.

services/token_airdrop.proto
IvanKavaldzhiev: I think I understood your point. Yes, we don't have a way to store paid fees and check them later but I believe this is an overcomplication. I think the idea is that if an airdrop transaction was performed successfully, then we know for sure (unless we have a bug), that all of the needed fees were collected. Then, when we perform airdrop claim - directly we will associate the token to the recipient, without charging for the association - just for the claim operation.

In the case where a token is in pending state and its first auto-renew period expires - we know that we should charge the payer of the airdrop for the next auto-renew fee.

In the case where an association was pre-paid by a payer and the token is in pending state and then the recipient associates themselves with the token explicitly, before claim - the recipient will also pay for the association. In this case the association fee will be paid two times, but probably we can live with this and this scenario might be more of a rare occasion.

@Nana-EC can you confirm this is correct?

jsync-swirlds: When you say `token` is in a pending state, I assume you mean `airdrop` here (and it's not "in a pending state" more that a "pending airdrop" entry exists; we should be removing those entries, immediately, when claimed or canceled.
If I think like a customer wishing to conduct an "airdrop", I get some concerns.
- If I conduct an airdrop and I expect a very low claim rate, I would prefer to pay for associations only when claimed; do we want to address this concern?
- If I conduct an airdrop I may wish to also conduct a "associate your account" campaign to my customers; should I end up with both of us charged for association when I do that?  I suspect many in the community (often with very loud "voices") will not appreciate the double charging, even if the amount is quite small.
- The description you offer leaves out the first renewal fee for direct transfers. Are we excluding those from a "benefit" of the airdrop?

As to pending airdrops as auto-renew flag:
- The pending airdrop should not be in state by the time the first renewal period expires.  We shouldn't, therefore, have anything to look up to determine the first auto renewal was pre-paid.  Also, looking up the pending airdrop is quite costly (either we have an extra virtual map or we have to scan all pending airdrops, *and* we must keep them for over 90 days after they're claimed), we shouldn't do that if we can avoid it.
- The token association renewal times do not (really must not) start until the airdrop is claimed (there is no association to renew before then).  We charge separately for renewing the pending airdrop entry itself (or should).  We shouldn't be renewing the association until one period (90 days) _after_ the airdrop _claim_ transaction.

IvanKavaldzhiev: Yes, I mean that we have an actual airdrop, which is saved in state.

Regarding the airdrop concerns:
- When you send an airdrop you should be conscious and aware that an association fee will be charged and that will additionally discourage malicious actors or airdrop senders that try to send spam tokens
- Ideally the association will be paid only during airdrop. Then the user should be aware that they have airdrops in their wallet and if they are notified and informed properly, it will be a rare case to manually associate the token while you can just claim your token. The user will still have the freedom to ignore the airdrop and manually associate to the token or associate to the token and then claim the token, but that's up to the users and we can't cover all of their behaviour. We want to have a frictionless airdrop and hide the association component from the receivers. They should just claim the airdrop and pay for the claim operation, without paying additional for something else
- I'm not sure I fully understand your comment but I agree that the start period for the expiration can be counted when the recipient claims the token and the actual association happens. But when we pre-pay the first auto-renew period the recipient will be free of charge for this association after claiming. IMO, recipients shouldn't be overloaded with paying multiple fees when they collect something like a "gift", so pre-payment for auto-renew happens both in direct transfer (when airdropping to a recipient with existing association) or when we have an airdrop in state

Regarding the auto-renew section:
- Why pending airdrop shouldn't be in state by the time the auto renew period expires? I think having the airdrop fields we are interested in - like sender, token, recipient we can use the `ReadableAirdropStore.exists()` method , defined in this [doc](https://github.com/hashgraph/hedera-services/pull/12792/files), and then perform auto-renew charging of the token (in case of token rent) or token association, that have expired.

- I agree that association auto-renew period can start to be counted on claim and this first period will already be pre-paid by the airdrop sender.

Something that might be missing here is when we have a pending airdrop in state - will it actually expire? Probably only if we have a token rent implemented or if the recipient has manually associated themselves with the token and the auto-renew period started counting. Otherwise, as you pointed out we don't have an association in place that can expire.

jsync-swirlds: I don't believe the quick/simple approach is going to work here.
My thoughts:
1. I was in no way suggesting that a _recipient_ would pay fees.  Only that if we want to keep this design we need to find a way to charge the payer of the _airdrop_ for the association fees without double charging (double charging, even rarely, is a very high risk event for the network).  We cannot, without problems, just tack on fees up front, we have to figure out a _correct_ mechanism to defer that to the proper time.
1. We don't have expiration or rent enabled (or entirely implemented) yet, so this isn't necessarily a "right now" issue, but we need to think through the implications and make sure we don't tie ourselves to an unworkable approach.
1. The pending airdrop needs to be removed when claimed or canceled because we do not have expiration implemented, there is no expiration defined for the pending airdrops (because they're supposed to be removed when claimed or canceled), and the cost of carrying these things in state indefinitely is unsustainable.
1. That simple method, `ReadableAirdropStore.exists()`, when the association needs to be renewed won't find your airdrop (we don't have an airdrop ID at that point).  You could try to synthesize one, but we don't know sender, and even if we did, that entry (as noted) no longer exists.

Unfortunately I did not know about the meeting this morning until too late, but I believe a short conversation would resolve the confusion here.  Feel free to ping me in slack if you would like to discuss.

PR #376, opened 2024-06-18
feat: add Services compilation check

tinker-michaelj:
**Description**:
 - Adds a GH action to validate Services will still compile with the protobufs on a PR to `main`.

### DIFF
--- a/.github/workflows/validate-protobufs.yml
+++ b/.github/workflows/validate-protobufs.yml
@@ -1,0 +1,29 @@
+name: Validate Protobufs Compilation
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  validate-protobufs:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Get PR branch name
+        id: pr_branch
+        run: echo "branch=${GITHUB_HEAD_REF}" >> $GITHUB_ENV
+
+      - name: Checkout hedera-services
+        uses: actions/checkout@v3
+        with:
+          repository: hashgraph/hedera-services
+          path: hedera-services
+
+      - name: Update hedera-services branch
+        run: |
+          sed -i 's/branch = "main"/branch = "${{ env.branch }}"/' hedera-services/hapi/build.gradle.kts
+
+      - name: Run Gradle assemble
+        run: ./gradlew assemble
+        working-directory: hedera-services


.github/workflows/validate-protobufs.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Checkout hedera-services
+        uses: actions/checkout@v3
+        with:
+          repository: hashgraph/hedera-services
+          path: hedera-services
+
+      - name: Update hedera-services branch
+        run: |
+          sed -i 's/branch = "main"/branch = "${{ env.branch }}"/' hedera-services/hapi/build.gradle.kts
```
jsync-swirlds, line 25: This might have issues with the occasional branch modification.  Here's a possible fix:
```suggestion
          sed -i 's/branch = "[a-zA-Z\-\/0-9]+"/branch = "${{ env.branch }}"/' hedera-services/hapi/build.gradle.kts
```

.github/workflows/validate-protobufs.yml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Run Gradle assemble
+        run: ./gradlew assemble
```
jsync-swirlds, line 28: This may be OK, but we might need to add the cgroup controls Nathan worked out for services to make sure the compile doesn't starve out the github agent on the runner.
Something to watch for.

.github/workflows/validate-protobufs.yml
Neeharika-Sompalli: Agreed. We might not have `main` as branch name always in services.

.github/workflows/validate-protobufs.yml
```diff
@@ -22,7 +18,7 @@ jobs:
 
       - name: Update hedera-services branch
         run: |
-          sed -i 's/branch = "main"/branch = "${{ env.branch }}"/' hedera-services/hapi/build.gradle.kts
+          sed -i.bak "s/branch = \".*\"/branch = \"${GITHUB_HEAD_REF}\"/" hedera-services/hapi/build.gradle.kts
```
jsync-swirlds, line 21: Something worth noting:
Sed is greedy by default, so this might match a lot more than we want, basically it will extend as far as it can without crossing to a new line.  That _shouldn't_ be an issue, unless someone puts a comment at end of line with a `"` in it, in which case this may break (that's why I restricted the match set in my suggestion).

Neeharika-Sompalli (approved):
LGTM! Thanks @tinker-michaelj

jsync-swirlds:
@tinker-michaelj Is this PR still active and relevant?

PR #382, opened 2024-06-25
chore: Block Stream protobuf changes for PBJ compiler

derektriley:
**Description**:
This PR makes a few changes to the block stream protobuf definitions to allow PBJ to compile them.

1. ```java_package``` and ```pbj.java_package``` need to be different packages so the compiled java classes in hedera-services don't clash. Same with the generated tests.
2. Fixed two javadoc errors in ```block_header.proto``` and ```state_changes.proto```
3. Moved enum definitons outside of message definition and changed their name's.

**Notes for reviewer**:
Suggest any additional changes or renaming. Thanks for your time.

**Checklist**

### DIFF
--- a/block/block_service.proto
+++ b/block/block_service.proto
@@ -62,7 +62,7 @@ package com.hedera.hapi.block;
  * limitations under the License.
  */
 
-option java_package = "com.hedera.hapi.block.stream";
+option java_package = "com.hederahashgraph.api.proto.java";
 // <<<pbj.java_package = "com.hedera.hapi.block.stream">>> This comment is special code for setting PBJ Compiler java package
 option java_multiple_files = true;
 
@@ -206,7 +206,7 @@ message PublishStreamResponse {
          * This code indicates the reason the stream ended.<br/>
          * This value MUST be set to a non-default value.
          */
-        ResponseCode status = 1;
+        EndOfStreamResponseCode status = 1;
 
         /**
          * The number of the last completed and _verified_ block.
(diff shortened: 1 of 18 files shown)


block/block_service.proto
```diff
@@ -62,7 +62,7 @@ package com.hedera.hapi.block;
  * limitations under the License.
  */
 
-option java_package = "com.hedera.hapi.block.stream";
+option java_package = "com.hederahashgraph.api.proto.java";
```
jsync-swirlds, line 65: I'd prefer to use `com.hedera.hapi.block.stream.protoc` for these, to avoid creating even more uses of a package name we do not want to retain.
More generally, the pattern would be `${package}.protoc`.

Hopefully we can remove these unnecessary and non-matching package directives sooner rather than later.

block/block_service.proto
```diff
@@ -220,45 +220,45 @@ message PublishStreamResponse {
...
     }
+}
 
+/**
+ * An enumeration indicating the status of this request.
+ *
+ * This enumeration describes the reason a block stream
+ * (sent via `writeBlockStream`) ended.
+ */
+enum EndOfStreamResponseCode {
```
jsync-swirlds, line 231: Probably best if we name this more like the other response codes in this file.
`PublishResponseCode` or `PublishStreamResponseCode` should work.

block/stream/output/consensus_service.proto
```diff
@@ -78,90 +78,90 @@ message SubmitMessageOutput {
...
+ *
+ * <hr style="margin: 0.2em 5em 0.2em 5em; height: 0.5em; border-style: solid none solid none; border-width: 2px;"/>
+ *
+ * The topic running hash is a 48-byte value that is the output
+ * of a SHA-384 digest with input data determined by the value of
+ * the `topic_running_hash_version` field.<br/>
+ * All new transactions SHALL use `topic_running_hash_version`
+ * `WITH_MESSAGE_DIGEST_AND_PAYER`.<br/>
+ */
+enum SubmitMessageOutputRunningHashVersion {
```
jsync-swirlds, line 102: This is general, it can just be `RunningHashVersion`, as it applies across the system, not just for one message.  I was planning to extract this to top level in the next update regardless.

block/stream/output/state_changes.proto
```diff
@@ -288,7 +288,7 @@ message SingletonUpdateChange {
         /**
          * A change to the exchange rates singleton.
          * <p>
-         * This change SHALL be present if the HBAR`<=>`USD exchange
+         * This change SHALL be present if the HBAR:USD exchange
```
jsync-swirlds, line 291: We could keep the original with entities and still work with javadoc (another PBJ limitation that it doesn't handle these; markdown javadoc _cannot_ come soon enough, in my opinion).
```suggestion
         * This change SHALL be present if the <tt>HBAR&lt;=&gt;USD</tt> exchange
```

block/stream/output/state_changes.proto
```diff
@@ -553,24 +553,24 @@ message MapChangeValue {
          * An updated set of staking information for all nodes in
          * the address book.
          */
-        proto.StakingNodeInfo staking_node_info_value = 33;
+        proto.StakingNodeInfo staking_node_info_value = 10;
```
jsync-swirlds, line 556: Thanks for renumbering these; I completely missed this one.

block/stream/block_header.proto
```diff
@@ -106,7 +106,7 @@ message BlockHeader {
      *         </ol>
      *       </li>
      *       <li>The second "internal" node SHALL be the parent of the
-     *           two "right-most" nodes.</li>
```
jsync-swirlds, line 109: Thanks for fixing this.

jsync-swirlds (commented):
A few updates I'd prefer to see.
If you prefer, I can make these changes after merge.

jsync-swirlds (changes requested):
Both recent merge commits are incorrect, and the result rewrites the whole destination branch in a way that breaks that branch and would break main, badly.
The merges need to be removed and the branch recovered before this PR can continue.

jsync-swirlds (approved):

block/stream/output/state_changes.proto
Nana-EC: Good catch

Nana-EC (approved):
Thanks, unfortunate that they are necessary but nice job capturing the changes

jsync-swirlds (approved):

PR #391, opened 2024-07-10
HIP-904: add airdrop claim protos

vtronkov:
**Description**:
Proto changes supporting TokenClaimAirdrop

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-services/issues/14201


Neeharika-Sompalli (commented):
LGTM! Except Joseph's comments. Also, do we expect any response codes to be added ?

vtronkov:
> LGTM! Except Joseph's comments. Also, do we expect any response codes to be added ?

Per this document I didn't find new response codes: https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/docs/design/services/token-service/airdrops/token-claim-airdrop.md

jsync-swirlds (approved):

edwin-greene (approved):

services/basic_types.proto
```diff
@@ -1745,3 +1750,32 @@ message PendingAirdropValue {
...
+    /**
+     * A list of one or more pending airdrop identifiers.
+     * <p>
+     * This transaction MUST be signed by the account identified by
+     * the `receiver_id` for each entry in this list.<br/>
+     * This list MUST contain between 1 and 10 entries, inclusive.<br/>
+     * This list MUST NOT have any duplicate entries.
+     */
+    repeated PendingAirdropId pending_airdrops = 1;
+}
```
jsync-swirlds, line 1781: The general pattern for Services is to add a new proto file for each transaction (e.g. `token_create`, `token_reject`, `token_update`, etc...).
This should _probably_ be in a file `token_claim_airdrop.proto`.

vtronkov: Fixed 👍

PR #393, opened 2024-07-10
Add protos for TSS Roster.

### LINKED ISSUE #392: Add proto for TSS Roster
As part of TSS Roster effort, a protobuf is required for a RosterEntry object that will be stored in the state, as well as updates to the state to store the roster entries.

As a part of this story, all protobuf models mentioned in the TSS-Roster design document need to be implemented. For a complete list, see https://github.com/hashgraph/hedera-services/pull/14045

kfa-aguda:
As part of [TSS Roster effort](https://github.com/hashgraph/hedera-services/pull/14045/files#top), a protobuf is required for a RosterEntry object that will be stored in the state. The state proto is also updated to include candidate and active roster hashes.

closes #392


jsync-swirlds:
To start, here are the recommendations for protobuf specification text, so you have that.
I'll comment on each block separately with suggested reword/restructure so you have direct examples as well.

## Recommendations for protobuf document "specification" comments
- Wrapping is 80 characters for top-level messages.
- Wrap everything else to 80 characters or 100 depending
   - Prefer to wrap lines to 80 characters
   - It is OK to go over 80, but less than 100, if it makes the
     text clearer and easier to read.
     - This SHOULD be rare, however.
   - If you have links or tags, wrapping at 100 characters for clarity
     is acceptable.
- Use blank lines (with `*` prefix) for line breaks for messages or
  enums, regardless of where defined.
   - Use `<p>` tags for _fields_, and _no_ "empty" lines (it breaks the tables).
- Use HTML tags where appropriate, but stick to markdown as much as possible.
- Use correct US-English spelling and grammar to the maximum extent possible.
   - Write proper and complete sentences, avoid partial sentences and
     unnecessary abbreviations.
   - Do not use partial words (e.g. instead of "spec" write "specification").
   - Explain all acronyms on first use in a document.
      - Use "Hedera Consensus Service(HCS)" the first time "HCS" is written
        but "HCS" may be used thereafter.
      - Extremely common acronyms (e.g. RAM or LASER) do not need
        to be expanded.
- First line of each document block is "what is this", and should be a single
  sentence.  Examples: "An identifier for an account." or "A `tokenTransfer`
  transaction request body.".
   - Avoid "this is a thing" type sentences, write "A thing" instead.
      - This applies for the _first sentence_ only, **not** elsewhere.
        The first sentence is a "what is this" answer, not full description
        and mostly just translates from "code" syntax to plain English.
- Include a line break (`<br/>`) after the first line, then a brief description,
  if appropriate.
   - Not every field needs a description, but _every_ `message` does.
   - Descriptions may be detailed, but should not include internal
     implementation details or requirements. The description _should_ be
     detailed enough to explain to an external contributor what the
     message or field contains, where it fits within the overall network
     system, and why it is needed.
      - This is a long-term goal; we have a long way to go to get to this
        level of description.
- After a paragraph break (a blank line for messages), one requirement per
  line. Use break tags at the end of each requirement except the last.
   - In field comments a `<p>` tag is _required_; blank lines break the
     automated documentation generator in **fields**.
- Be _prescriptive_ in requirements.  Do not _describe_ the behavior.
   - Declare what MUST be done for valid inputs, and what SHALL be present in
     valid outputs (or in state messages).
   - Declare what SHALL happen as a result of a successful, or failed,
     transaction.
   - The word `will` usually means `SHALL` or `MUST`.
   - The word `is` often means `SHALL be`.
   - Use both positive (`SHALL` or `MUST`) and negative (`SHALL NOT` or
     `MUST NOT`) requirements.
   - Avoid "perfect" or "imperfect" verb aspects and "wiggle" words
     (might, could, perhaps, etc...).
      - The specification items MAY, SHOULD, and RECOMMENDED serve to fill
        the _occasional_ need for less imperative requirements.
   - Also avoid "progressive" verbs (e.g. working, writing, going, deleting).
- Requirements should be short and focused.
   - Better to have three separate lines for three items than one longer
     line with three clauses.
   - Duplicating text across lines with slightly different requirements
     is not only acceptable, but often recommended.
   - Well written specification tends to be much more short and clipped
     sentences than "normal" prose. Some describe the flow as "staccato".
- Add another blank line after requirements for `message` documents only.
- Each message that represents a transaction body **must** document the
  Record Stream/Block Stream effects with a heading `### Record Stream Effects`.
- General notes go last, for `message` only with a heading
  `#### Additional Notes`.

platform/state/platform_state.proto
```diff
@@ -81,6 +81,25 @@ message PlatformState {
...
     proto.Timestamp last_frozen_time = 5;
 
+    /**
+     * The SHA-384 hash of a candidate roster.
+     * <p>
+     * This is the hash of the roster that is currently being considered
+     * for adoption.
+     * A Node SHALL NOT, ever, have more than one candidate roster
+     * at the same time.
+     */
```
jsync-swirlds, line 91:
```suggestion
    /**
     * A cryptographic hash of a roster.<br/>
     * This hash represents a candidate roster, which is a roster that
     * is not yet active, but is being considered.
     * <p>
     * A Node SHOULD NOT have more than one candidate roster at
     * any given moment.
     */
```

platform/state/platform_state.proto
```diff
@@ -81,6 +81,25 @@ message PlatformState {
...
+    bytes candidate_roster_hash = 6;
+
+    /**
+     * The SHA-384 hash of an active roster
+     * This is the hash of the roster that has already been adopted by the
+     * network.
+     * A Node SHALL NOT, ever, have more than one active roster
+     * at the same time.
+     */
```
jsync-swirlds, line 100:
```suggestion
    /**
     * A cryptographic hash of a roster.<br/>
     * This hash represents an active roster, which is a roster that
     * is adopted by the network and in current use.
     * <p>
     * The network SHALL NOT have more than one active roster at
     * any given moment.<br/>
     * This value MUST be a SHA-384 hash.
     */
```

services/state/roster/roster_entry.proto
```diff
@@ -0,0 +1,97 @@
+syntax = "proto3";
```
jsync-swirlds, line 1: It wouldn't hurt to include a file comment here, but it's not strictly required at the moment.

services/state/roster/roster_entry.proto
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * A single roster in the network state.
+ *
+ * Each roster entry in the roster list SHALL encapsulate the pieces of data
+ * required to manage node participation in the Threshold Signature Scheme (TSS)
+ *
+ * Roster nodeIDs SHALL NOT be _globally_ uniquely identified. A given nodeId
+ * is only valid within a single realm and shard combination, so the identifier
+ * for a network node SHALL only be unique within a single realm and shard
+ * combination.
```
jsync-swirlds, line 36:
```suggestion
 * Each roster entry in the roster list SHALL encapsulate the elements required
 * to manage node participation in the Threshold Signature Scheme (TSS).
```

services/state/roster/roster_entry.proto
```diff
@@ -0,0 +1,97 @@
...
+    uint64 node_id = 1;
+
+    /**
+     * A consensus weight.
+     * <p>
+     * Each node SHALL have a weight in consensus calculations.<br/>
+     * The consensus weight of a node SHALL be calculated based on the amount
+     * of HBAR staked to that node.<br/>
+     * Consensus SHALL be calculated based on agreement of greater than `2/3`
+     * of the total `weight` value of all nodes on the network.
```
jsync-swirlds, line 56: Perhaps worth noting that this _may_ not be correct (Edward is discussing on the TSS Notion page).

services/state/roster/roster_entry.proto
```diff
@@ -0,0 +1,97 @@
...
+     */
+    uint64 weight = 2;
+
+    /**
+     * The RSA public certificate used for signing gossip events.
+     * <p>
+     * This value SHALL be a certificate of a type permitted for gossip
+     * signatures.<br/>
+     * This value SHALL be the DER encoding of the certificate presented.<br/>
+     * This field is REQUIRED and MUST NOT be empty.
```
jsync-swirlds, line 66:
```suggestion
     * An RSA public certificate used for signing gossip events.
     * <p>
     * This value SHALL be a certificate of a type permitted for gossip
     * signatures.<br/>
     * This value SHALL be the DER encoding of the certificate presented.<br/>
     * This field is REQUIRED and MUST NOT be empty.
```

services/state/roster/roster_entry.proto
```diff
@@ -0,0 +1,97 @@
...
+     */
+    bytes gossip_ca_certificate = 3;
+
+    /**
+     * The long term elliptic-curve cryptography
+     * public key for the node used in TSS.
+     * <p>
+     * This value SHALL be the DER encoding of the presented elliptic-curve
+     * public key that supports TSS<br/>
+     * This field is REQUIRED and MUST NOT be empty.
```
jsync-swirlds, line 76:
```suggestion
     * An elliptic curve public key.<br/>
     * This contains the _long term_ public key for this node.
     * <p>
     * This value SHALL be the DER encoding of the presented elliptic curve
     * public key.<br/>
     * This field is REQUIRED and MUST NOT be empty.
```

services/state/roster/roster_entry.proto
kfa-aguda: The good news is we have a bit of time between now and when this is pushed. By that time, we would have had some more clarity.

services/state/roster/roster.proto
```diff
@@ -0,0 +1,126 @@
...
+ */
+message RosterMap {
+
+    /**
+     * A map of roster hashes to roster entries.
+     * The map will be stored in the state as a PBJ map.
+     * It will itself be managed as a Singleton state object.
+     */
+    map<string, RosterEntry> entries = 1;
+}
```
anthony-swirldslabs, line 126: This should probably be removed in favor of the latest design decision to use States API maps for storing the RosterMap.

platform/state/platform_state.proto
```diff
@@ -81,6 +81,28 @@ message PlatformState {
...
+     * A cryptographic hash of a roster.<br/>
+     * This hash represents an active roster, which is a roster that
+     * is adopted by the network and in current use.
+     * <p>
+     * The network SHALL NOT have more than one active roster at
+     * any given moment.<br/>
+     * This value MUST be a SHA-384 hash.
+     */
+    bytes active_roster_hash = 7;
+
```
anthony-swirldslabs, line 105: Does this belong to the PlatformState or a separate RosterState singleton per the design?

Also, I recall we wanted to replace the `active_roster_hash` with a list of round/roster_hash pairs.

services/state/roster/roster.proto
kfa-aguda: it needs a quick update post roster conversations. Should have that ready by end of day.

platform/state/platform_state.proto
kfa-aguda: I paused updating this PR until fully ratified in Roster Proposal. Now that that's been accepted, I'll pull those updates here today.

services/state/roster/roster.proto
```diff
@@ -0,0 +1,125 @@
...
+     * This list SHALL contain roster entries in natural order of ascending node ids.
+     * This list SHALL NOT be empty.<br/>
+     */
+    repeated RosterEntry rosters = 1;
+}
+
+/**
+ * A single roster entry in the network state.
+ *
+ * Each roster entry in SHALL encapsulate the elements required
```
anthony-swirldslabs, line 48:
```suggestion
 * Each roster entry SHALL encapsulate the elements required
```

services/state/roster/roster.proto
```diff
@@ -0,0 +1,125 @@
...
+ * to manage node participation in the Threshold Signature Scheme (TSS).<br/>
+ * All fields are REQUIRED.
```
anthony-swirldslabs, line 50: This contradicts the lines 92-93 just below. I suggest to remove this statement from here and instead use field-specific docs below about whether a field is required or not.

services/state/roster/roster.proto
anthony-swirldslabs: This RosterMap protobuf model still needs to go.

services/state/roster/roster_state.proto
```diff
@@ -0,0 +1,80 @@
...
+   * A Node SHALL NOT, ever, have more than one candidate roster
+   * at the same time.
+   */
+  byte candidate_roster_hash = 1;
+
+  /**
+   * A list of round numbers and roster hashes.
+   * The round number indicates the round in which the corresponding roster became active
+   * <p>
+   * This list SHALL be ordered by round numbers.<br/>
```
anthony-swirldslabs, line 53: In ascending or descending order? It might be worth specifying this here.

services/state/roster/roster.proto
kfa-aguda: how so, because it says it's optional on line 92? If yes, i don't think that's contradictory. Line 92/93 is just saying once it's set, you can't unset it.

services/state/roster/roster_state.proto
kfa-aguda: I think specifying ascending here is kind of tautological. There's nowhere in the hashgraph logic where anything related to round numbers can possibly be descending. I think it's fine to leave it.

services/state/roster/roster.proto
anthony-swirldslabs: Well, it is. Line 50:
> All fields are REQUIRED.

Line 92:
> This field is OPTIONAL

So either the definition of `all` doesn't include this particular field for some reason, or not all fields are actually required. Logically, there's a contradiction currently. You either want to provide an explicit definition for the word `all` as used in this context, or you want to remove either of the statements.

As for "unsetting" a field - PBJ models are immutable by definition. So if you're talking about an instance of this model in JVM, then this can be assumed and doesn't need mentioning. If you're talking about the lifecycle of the RosterEntry object over a span of multiple roster changes, then I don't believe that this particular aspect belongs to the specification of this protobuf model because it's not a property of this model, but rather a property of an algorithm that implements and supports the lifecycle of the roster as a whole, which is a very high-level concept compared to this tiny low-level model.

services/state/roster/roster_state.proto
anthony-swirldslabs: Why? One could imagine that an element at index zero could return the "active" or "latest" entry, which would imply a descending order. And this may make sense as fetching the element at index zero is the fastest operation because it doesn't require the knowledge about the size of your list.

So the actual order is not very clear from the specification. I suggest to specify it explicitly to avoid any confusion or misinterpretation.

services/state/roster/roster.proto
kfa-aguda: Gotcha. How about `All fields except tss_encryption_key are REQUIRED.. Is that better?

services/state/roster/roster_state.proto
kfa-aguda: applied your suggestion.

services/state/roster/roster.proto
```diff
@@ -3,6 +3,8 @@ syntax = "proto3";
 package com.hedera.hapi.node.state.roster;
 
 /*
+ * Hedera Network Services Protobuf
+ *
```
jsync-swirlds, line 7: We don't include these lines in new protobuf; they're only present in old files that followed an obsolete recommendation.

services/state/roster/roster.proto
```diff
@@ -26,39 +28,48 @@ option java_multiple_files = true;
...
 message RosterEntry {
 
     /**
      * A consensus node identifier.
      * <p>
-     * Node identifiers SHALL be unique _within_ a shard and realm,
-     * but a node SHALL NOT, ever, serve multiple shards or realms,
-     * therefore the node identifier MAY be repeated _between_ shards and realms.
+     * Node identifiers SHALL be unique _within_ a ledger,
+     * and MAY NOT be repeated _between_ shards and realms.
```
jsync-swirlds, line 58:
```suggestion
     * and MUST NOT be repeated _between_ shards and realms.
```

services/state/roster/roster.proto
```diff
@@ -26,39 +28,48 @@ option java_multiple_files = true;
...
     /**
      * A consensus weight.
      * <p>
-     * Each node SHALL have a weight in consensus calculations.<br/>
-     * The consensus weight of a node SHALL be calculated based on the amount
-     * of HBAR staked to that node.<br/>
-     * Consensus SHALL be calculated based on agreement of greater than `2/3`
-     * of the total `weight` value of all nodes on the network.
+     * Each node SHALL have a weight of zero or more in consensus calculations.<br/>
+     * The sum of the weights of all nodes in the roster SHALL form the total weight of the system,<br/>
```
jsync-swirlds, line 66: Generally should not include a break tag in the middle of a sentence.
```suggestion
     * The sum of the weights of all nodes in the roster SHALL form the total weight of the system,
```

services/state/roster/roster.proto
```diff
@@ -68,14 +79,20 @@ message RosterEntry {
     bytes gossip_ca_certificate = 3;
 
     /**
-     * The long term elliptic-curve cryptography
-     * public key for the node used in TSS.
+     * An ALT_BN128 elliptic curve public encryption key.<br/>
```
jsync-swirlds, line 82: Perhaps a little too specific here:
```suggestion
     * An elliptic curve public encryption key.<br/>
     * This is currently an ALT_BN128 key.<br/>
```
Also, we prefer not to have a paragraph break within the description for a field; so the tag on line 83 could be removed.

services/state/roster/roster.proto
```diff
@@ -68,14 +79,20 @@ message RosterEntry {
...
      * <p>
-     * This value SHALL be the DER encoding of the presented elliptic-curve
-     * public key that supports TSS<br/>
-     * This field is REQUIRED and MUST NOT be empty.
+     * The elliptic curve type may change in the future. For example, <br/>
```
jsync-swirlds, line 84:
```suggestion
     * The elliptic curve type may change in the future. For example,
```

services/state/roster/roster.proto
```diff
@@ -68,14 +79,20 @@ message RosterEntry {
...
+     * if the Ethereum ecosystem creates precompiles for BLS12_381,
+     * we may switch to that curve.
+     * <p>
+     * This value SHALL be specified according to EIP-196 and EIP-197 standards, <br/>
```
jsync-swirlds, line 88: It looks like you've added break tags when we're just breaking the line for wrapping.
We should only add a break when the javadoc or markdown should have a newline.
```suggestion
     * This value SHALL be specified according to EIP-196 and EIP-197 standards,
```

services/state/roster/roster.proto
```diff
@@ -68,14 +79,20 @@ message RosterEntry {
...
+     * See https://eips.ethereum.org/EIPS/eip-196#encoding and <br/>
```
jsync-swirlds, line 89:
```suggestion
     * See https://eips.ethereum.org/EIPS/eip-196#encoding and
```

services/state/roster/roster.proto
```diff
@@ -68,14 +79,20 @@ message RosterEntry {
...
+     * https://eips.ethereum.org/EIPS/eip-197#encoding
+     * <p>
```
jsync-swirlds, line 91: For _fields_ we should use line breaks between individual requirements, and a single paragraph break between the summary/description and the requirements.
```suggestion
     * https://eips.ethereum.org/EIPS/eip-197#encoding<br/>
```

services/state/roster/roster.proto
```diff
@@ -88,10 +105,6 @@ message RosterEntry {
      * then endpoints in this list MAY supply either IP address or FQDN, but
      * SHALL NOT supply both values for the same endpoint.<br/>
      * This list SHALL NOT be empty.<br/>
-     * This list SHALL NOT contain more than `10` entries.<br/>
-     * The first two entries in this list SHALL be the endpoints published to
-     * all consensus nodes.<br/>
-     * All other entries SHALL be reserved for future use.
```
jsync-swirlds: These lines are a requirement in the Node definition.  Are we expecting something different now?

services/state/roster/roster.proto
```diff
@@ -88,10 +105,6 @@ message RosterEntry {
...
      */
     repeated proto.ServiceEndpoint gossip_endpoint = 5;
-}
+}
```
jsync-swirlds, line 110: A newline is required at the end of the file.

services/state/roster/roster_state.proto
```diff
@@ -0,0 +1,79 @@
+syntax = "proto3";
+
+package com.hedera.hapi.node.state.roster;
+
+/*
+ * Hedera Network Services Protobuf
+ *
```
jsync-swirlds, line 7: These two lines should not be present.
Also, check for hidden zero-width characters if you copied this from an old proto file.

services/state/roster/roster_state.proto
```diff
@@ -0,0 +1,79 @@
...
+
+option java_package = "com.hederahashgraph.api.proto.java";
+// <<<pbj.java_package = "com.hedera.hapi.node.state.roster">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+/**
+ * The current state of platform rosters.<br/>
+ * This message stores a roster data for the platform in network state.
+ *
+ * The roster state SHALL encapsulate the incoming candidate roster's hash,<br/>
```
jsync-swirlds, line 33:
```suggestion
 * The roster state SHALL encapsulate the incoming candidate roster's hash,
```

services/state/roster/roster_state.proto
```diff
@@ -0,0 +1,79 @@
...
+ * and a list of pairs of round number and active roster hash.<br/>
+ * This data SHALL be used to track round numbers and the rosters used in determining the consensus.<br/>
+ */
+message RosterState {
+
+  /**
+   * The SHA-384 hash of a candidate roster.
+   * <p>
+   * This is the hash of the roster that is currently being considered
+   * for adoption.
```
jsync-swirlds, line 43:
```suggestion
   * for adoption.<br/>
```

services/state/roster/roster.proto
anthony-swirldslabs: That would sound perfect!

services/state/roster/roster_state.proto
```diff
@@ -0,0 +1,79 @@
...
+   * A Node SHALL NOT, ever, have more than one candidate roster
+   * at the same time.
+   */
+  bytes candidate_roster_hash = 1;
+
+  /**
+   * A list of round numbers and roster hashes.
```
jsync-swirlds, line 50:
```suggestion
   * A list of round numbers and roster hashes.<br/>
```

anthony-swirldslabs (approved):

jsync-swirlds (commented):
Some small text updates.
Mostly the slight overuse of the break (`<br/>`) tag.

services/state/roster/roster.proto
```diff
@@ -0,0 +1,110 @@
...
+     * signatures.<br/>
+     * This value SHALL be the DER encoding of the certificate presented.<br/>
+     * This field is REQUIRED and MUST NOT be empty.
+     */
+    bytes gossip_ca_certificate = 3;
+
+    /**
+     * An ALT_BN128 elliptic curve public encryption key.<br/>
+     * <p>
+     * The elliptic curve type may change in the future. For example, <br/>
```
jsync-swirlds, line 84:
```suggestion
     * An elliptic curve public encryption key.<br/>
     * This is currently an ALT_BN128 curve, but the elliptic curve
     * type may change in the future. For example,
```

services/state/roster/roster.proto
```diff
@@ -0,0 +1,110 @@
...
+     * if the Ethereum ecosystem creates precompiles for BLS12_381,
+     * we may switch to that curve.
+     * <p>
+     * This value SHALL be specified according to EIP-196 and EIP-197 standards, <br/>
+     * See https://eips.ethereum.org/EIPS/eip-196#encoding and <br/>
+     * https://eips.ethereum.org/EIPS/eip-197#encoding
+     * <p>
+     * This field is OPTIONAL (that is, it can initially be null)
+     * but once set, it can no longer be null.
```
jsync-swirlds, line 93: For _fields_ we should use line breaks between individual requirements, and a single paragraph break between the summary/description and the requirements.
```suggestion
     * This value SHALL be specified according to EIP-196 and EIP-197 standards,
     * See https://eips.ethereum.org/EIPS/eip-196#encoding and
     * https://eips.ethereum.org/EIPS/eip-197#encoding<br/>
     * This field is _initially_ OPTIONAL (i.e. it can be unset _when created_)
     * but once set, it is REQUIRED thereafter.
```

services/state/roster/roster.proto
kfa-aguda: >The first two entries in this list SHALL be the endpoints published to all consensus nodes
Richard pushed back on the correctness of this on the proposal PR so i removed it.
And as for the size, it's something that's defined in the code and I believe should be validated right where it's defined. The prototype definition itself has no facility to enforce it.

jsync-swirlds: Three points:

The specification should still be expressing such limitations; else how does a consumer of block stream (this is a public API via block stream) know what those limitations are?

Out protobuf definitions _may_ be extended with proto-validate or similar support when PBJ is able to handle that; these specifications are _hoped_ to support doing so with minimal need to re-read the full codebase.

If the requirement is _actually_ incorrect, then that should be discussed with services as well, and the change (including to specification in the node create and update transactions) added to any HIP related to adopting the Roster.

The primary purpose of specification is to enable consumers of state information to correctly consume the data, and also support any sufficiently motivated entity to create an independent, and correct, consensus node implementation.

Any code enforced limitations that are relevant _should_ be included.

services/state/roster/roster.proto
```diff
@@ -107,4 +104,5 @@ message RosterEntry {
      * This list SHALL NOT be empty.<br/>
      */
     repeated proto.ServiceEndpoint gossip_endpoint = 5;
+
 }
```
jsync-swirlds, line 108: Not a blank line before the end curly brace, we require a newline at EOF (github adds a warning symbol to show where that's missing).
The consistent presence of the newline helps some of the tooling to handle all files consistently.

kfa-aguda: sorry about that. I just checked and didn't get the newline warning in intellij. Guessing i have to enable it, i'll look for it. Updated now

services/state/roster/roster.proto
jsync-swirlds: 1. The specification can refer to configuration, but the critical element is that it is currently specified in the node create and node update transactions, so it's a HIP to change it there.  If the limitation is variable (I don't actually know the source of the limit), then we probably need a cross-team conversation to clarify that.
1. The line can be removed, if it's not accurate. I believe the original line came from a platform "this is what we currently do" requirement.  If that's changed, then, again, we probably need to make sure services is aware and that all of the associated documentation (including code) is updated.

anthony-swirldslabs (approved):

jsync-swirlds (approved):

services/state/roster/roster.proto
kfa-aguda: noted, I'll check with services.

Neeharika-Sompalli (approved):
LGTM as per the proposal! Thanks @kfa-aguda

kimbor (approved):
lgtm!

PR #397, opened 2024-07-17
feat: 14294 Add `AddressBook` to `PlatformState`

imalygin:
**Description**:

This PR adds two fields to `PlatformState` object. This is a temporary solution that should be sunsetted once TSS Roster (https://github.com/hashgraph/hedera-services/pull/14045) is implemented. In fact, the migration from `AddressBook` to `Roster` objects should be a part of this design. 

**Important**: The implementation of the `AddressBook` is not optimized and reflects exactly `com.swirlds.platform.system.address.AddressBook` and its components. This implementation  will be sunsetted **before** we launch block streams, and therefore there is no need in optimizing it. 

**Related issue(s)**:

Fixes # https://github.com/hashgraph/hedera-services/issues/14294 

**Notes for reviewer**:

**Checklist**


platform/state/platform_state.proto
```diff
@@ -195,3 +211,134 @@ message MinimumJudgeInfo {
...
+/**
+ * A consensus node identifier.<br/>
+ * This value uniquely identifies a single consensus node within the network.
+ */
+message NodeId {
+  /**
+   * A numeric identifier.
+   */
+  uint64 id = 1;
+}
```
jsync-swirlds, line 344: Missing newline at EOF.

platform/state/platform_state.proto
```diff
@@ -195,3 +211,134 @@ message MinimumJudgeInfo {
...
+   * A string that provides additional information about this consensus node.
+   */
+  string memo = 11;
+}
+
+/**
+ * A consensus node identifier.<br/>
+ * This value uniquely identifies a single consensus node within the network.
+ */
+message NodeId {
```
jsync-swirlds, line 339: Is this message necessary?  Could we just have node ID values be direct uint64 values (or even uint32 or uint16 given that we don't actually have that many nodes)?

platform/state/platform_state.proto
```diff
@@ -195,3 +211,134 @@ message MinimumJudgeInfo {
...
+   */
+  int64 round = 1;
+
+  /**
+   * The node ID of the next address that can be added.
+   * <p>
+   * The next node identifier assigned SHALL be equal to this value.<br/>
+   * All existing node identifiers SHALL be strictly less than this value.<br/>
+   */
+  NodeId next_node_id = 2;
```
jsync-swirlds, line 233: Is this needed?  I assume we're setting the address book information from `config.txt` rather than computing node identifiers separately.

platform/state/platform_state.proto
```diff
@@ -195,3 +211,134 @@ message MinimumJudgeInfo {
...
+   * The name that a consensus node uses to refer to themselves.
+   */
+  string self_name = 3;
+
+  /**
+   * A consensus weight.<br/>
+   * This value is the relative weight for this consensus node in the
+   * consensus voting algorithm.
+   */
+  uint64 weight = 4;
```
jsync-swirlds, line 277: Could this be uint32 or smaller?

platform/state/platform_state.proto
```diff
@@ -195,3 +211,134 @@ message MinimumJudgeInfo {
...
+message Address {
+  /**
+   * The ID of this member.
+   * <p>
+   * This identifier SHALL be agreed upon by all consensus nodes.
+   */
+  NodeId id = 1;
+
+  /**
+   * The nickname this consensus node uses to refer to another consensus node.
```
jsync-swirlds, line 263: Should this be "nickname other nodes use to refer to this node"?

platform/state/platform_state.proto
```diff
@@ -195,3 +211,134 @@ message MinimumJudgeInfo {
...
+}
+
+/**
+ * A single network address.
+ * This is one address in the network address book, including all required
+ * information to include that consensus node in the consensus gossip.
+ *
+ * ### Implementation Detail
+ * When `enableEventStreaming` is set to true, the memo field is REQUIRED
+ * and MUST be unique.
```
jsync-swirlds, line 252: What is memo used for that this requirement exists?
Also, could we just rename the field in this temporary structure to describe how we are _actually_ using it?

jsync-swirlds (commented):
A few questions.

platform/state/platform_state.proto
imalygin: Added

platform/state/platform_state.proto
imalygin: No, actually this nickname is a name that the current node uses to refer to the other consensus node that is in its `AddressBook`. 

That is, we have something like this for the node0

|| address || nickname ||
|0.0.4 | Swirlds |
|  0.0.5 | FIS |
| 0.0.6 | Wipro |

platform/state/platform_state.proto
imalygin: This comment comes from `com.swirlds.platform.system.address.AddressBook` javadoc. I took it at face value, but at the second glance I couldn't find any justification for this requirement, so I remove this comment

platform/state/platform_state.proto
```diff
@@ -195,3 +211,134 @@ message MinimumJudgeInfo {
...
+   * This value SHALL be between 1 and 65535.
+   */
+  uint32 port_external = 8;
+
+  /**
+   * The signing x509 certificate of the consensus node.
+   * <p>
+   * This SHALL provide the public key used for signing.
+   */
+  bytes sig_cert = 9;
```
kfa-aguda, line 317:
```suggestion
  bytes signing_certificate = 9;
```

platform/state/platform_state.proto
```diff
@@ -195,3 +211,134 @@ message MinimumJudgeInfo {
...
+
+  /**
+   * The agreement x509 certificate of the consensus node.
+   * <p>
+   * This SHALL be used for establishing TLS connections.
+   */
+  bytes agree_cert = 10;
```
kfa-aguda, line 324:
```suggestion
  bytes agreement_certificate = 10;
```

platform/state/platform_state.proto
jsync-swirlds: This doesn't say what you're describing because the context is not "the node reading config.txt"  The context is "the node described in this individual message".

"This consensus node" in context only has this one nickname, so there is no way it can map it's own nickname to other nodes (unless each node refers to _every_ other node by it's own nickname).

If this is temporary we can ignore the incongruity of this text, but it's definitely not saying what you're describing.

imalygin: This is how javadoc looks like for the field
```
    /** name this member uses to refer to that member */
    private String nickname;
```

I'm not sure if I understand how you'd like to rephrase it, given the fact that this nickname is the information about other node that this node have.

jsync-swirlds: The perspective is reversed in the protobuf from the java, so this should work well:
```suggestion
   * The nickname other consensus nodes will use to refer to this consensus node.
```

kfa-aguda (approved):

PR #401, opened 2024-07-30
chore: Rename payload to transaction

timo0:
Part of https://github.com/hashgraph/hedera-services/issues/14000

### DIFF
--- a/platform/event/event_descriptor.proto
+++ b/platform/event/event_descriptor.proto
@@ -54,7 +54,7 @@ message EventDescriptor {
    * The hash SHALL be a SHA-384 hash.<br/>
    * The hash SHALL have the following inputs, in the specified order:<br/>
    * 1. The bytes of the `EventCore` protobuf<br/>
-   * 2. The SHA-384 hash of each individual `EventPayload`, in the order the payloads appear in the `event_payloads` field of the `GossipEvent` protobuf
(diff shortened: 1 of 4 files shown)


platform/event/event_descriptor.proto
```diff
@@ -54,7 +54,7 @@ message EventDescriptor {
    * The hash SHALL be a SHA-384 hash.<br/>
    * The hash SHALL have the following inputs, in the specified order:<br/>
    * 1. The bytes of the `EventCore` protobuf<br/>
-   * 2. The SHA-384 hash of each individual `EventPayload`, in the order the payloads appear in the `event_payloads` field of the `GossipEvent` protobuf
+   * 2. The SHA-384 hash of each individual `EventTransaction`, in the order the transactions appear in the `event_transactions` field of the `GossipEvent` protobuf
```
jsync-swirlds, line 57: nit: This line is a bit long.  Ideally we'd wrap these comments at about 80 characters.

jsync-swirlds (approved):
Looks Good.  Only minor editing nitpicks (Which can be addressed elsewhere if desired).

lpetrovic05 (approved):

lpetrovic05 (changes requested):
Looks good, but must not be merged before the hedera repo gets updated, so blocking it

lpetrovic05 (approved):

PR #404, opened 2024-08-06
Add TransactionReceiptEntries to RecordCache

Neeharika-Sompalli:
As per https://github.com/hashgraph/hedera-services/issues/14541, added `TransactionReceiptEntries` object that will be stored in state. It is needed to reduce the size of receipt cache in state.

### DIFF
--- a/services/state/recordcache/recordcache.proto
+++ b/services/state/recordcache/recordcache.proto
@@ -24,6 +24,7 @@ package proto;
 
 import "basic_types.proto";
 import "transaction_record.proto";
+import "response_code.proto";
 
 option java_package = "com.hederahashgraph.api.proto.java";
 // <<<pbj.java_package = "com.hedera.hapi.node.state.recordcache">>> This comment is special code for setting PBJ Compiler java package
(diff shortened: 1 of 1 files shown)


services/state/recordcache/recordcache.proto
```diff
@@ -53,3 +54,34 @@ message TransactionRecordEntry {
...
+ * As a single transaction is handled a receipt is created. It is stored in state for a configured time
+ * limit (perhaps, for example, 3 minutes). During this time window, any client can query the node and get the
+ * receipt for the transaction. The TransactionReceiptEntry is the object stored in state with this information.
+ */
+message TransactionReceiptEntry {
+    /**
+    * The ID of the node that submitted the transaction to consensus. The ID is the ID of the node as known by the
+    * address book. Valid node IDs are in the range 0..2^63-1, inclusive.
+    */
+    int64 node_id = 1;
```
jsync-swirlds, line 67: This should be `uint64` to match what [the `Node` object in state](https://github.com/hashgraph/hedera-protobufs/blob/main/services/state/addressbook/node.proto#L46) uses.

Neeharika-Sompalli: Fixed. Thanks @jsync-swirlds

iwsimon (approved):
LGTM. Thanks @Neeharika-Sompalli

jsync-swirlds (approved):

PR #405, opened 2024-08-08
fix: support additional states

tinker-michaelj:
**Description**:
 - Add some fields to support additional states used.


block/stream/output/state_changes.proto
```diff
@@ -810,4 +854,4 @@ message QueuePushChange {
  * information beyond the state name common to all state changes.
  */
 message QueuePopChange {
-}
+}
```
jsync-swirlds, line 857: A newline is required at EOF (isn't enforced yet due to minimal CI support).

block/stream/output/state_changes.proto
```diff
@@ -635,7 +658,7 @@ message MapChangeKey {
         /**
          * A change to a virtual map keyed by File identifier.
          */
-        proto.FileID filed_id_key = 4;
+        proto.FileID file_id_key = 4;
```
jsync-swirlds, line 661: Thanks for fixing this.

block/stream/output/state_changes.proto
```diff
@@ -41,8 +41,10 @@ option java_multiple_files = true;
 import "google/protobuf/wrappers.proto";
 import "basic_types.proto";
 import "exchange_rate.proto";
+import "state/addressbook/node.proto";
 import "state/blockrecords/block_info.proto";
 import "state/blockrecords/running_hashes.proto";
+import "state/blockstream/block_stream_info.proto";
```
jsync-swirlds, line 47: Is this in another PR?
I cannot find this file.

jsync-swirlds (commented):
One trivial (but required) fix and a file I cannot find.

block/stream/output/state_changes.proto
tinker-michaelj: Ah yes, added.

block/stream/output/state_changes.proto
tinker-michaelj: Oops, didn't add! Should be there now.

jsync-swirlds (commented):
Some comments for the newly added file.

PR #407, opened 2024-08-09
feat: 14773 Created `PlatformVersion` message that adequately represents `PlatformState` version

imalygin:
**Description**:

This PR adds correct representation of the `PlatformState`  version that contains not only software version but HAPI version as well.

**Related issue(s)**:

Fixes # https://github.com/hashgraph/hedera-services/issues/14773

### DIFF
--- a/platform/state/platform_state.proto
+++ b/platform/state/platform_state.proto
@@ -347,7 +347,9 @@ message NodeId {
  */
 message PlatformVersion {
     /**
-     * Version of the HAPI specification that was used to serialize the state.
+     * An API version.<br/>
+     * This is the Hedera API specification supported by the release of
+     * the Hashgraph Platform SDK that produced this message.
      */
     proto.SemanticVersion hapi_version = 1;
     /**
--- a/platform/state/platform_state.proto
+++ b/platform/state/platform_state.proto
@@ -352,6 +352,7 @@ message PlatformVersion {
      * the Hashgraph Platform SDK that produced this message.
      */
     proto.SemanticVersion hapi_version = 1;
+
     /**
      * A software version for the Hashgraph Platform SDK.<br/>
      * This is the version of the software that produced the state containing
--- a/platform/state/platform_state.proto
+++ b/platform/state/platform_state.proto
@@ -50,7 +50,7 @@ message PlatformState {
      * <p>
      * This SHALL be the software version that created this state.
      */
-    proto.SemanticVersion creation_software_version = 1;
+    PlatformVersion creation_software_version = 1;
 
     /**
      * A number of non-ancient rounds.
@@ -117,7 +117,7 @@ message PlatformState {
      * If birth round migration is complete, this SHALL be the _first_ software
      * version that enabled birth round mode.
     */
-    proto.SemanticVersion first_version_in_birth_round_mode = 10003 [deprecated = true];
+    PlatformVersion first_version_in_birth_round_mode = 10003 [deprecated = true];
 
     /**
      * An address book for this round.
@@ -338,4 +338,18 @@ message NodeId {
    * A numeric identifier.
    */
   uint64 id = 1;
+}
+
+/**
+ * A version of the platform software.
+ */
+message PlatformVersion {
+    /**
+     * Version of the HAPI specification that was used to serialize the state.
+     */
+    proto.SemanticVersion hapi_version = 1;
+    /**
+     * The software version of the state
+     */
+    proto.SemanticVersion software_version = 2;
 }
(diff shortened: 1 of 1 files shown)


platform/state/platform_state.proto
```diff
@@ -339,3 +339,17 @@ message NodeId {
...
+message PlatformVersion {
+    /**
+     * Version of the HAPI specification that was used to serialize the state.
+     */
+    proto.SemanticVersion hapi_version = 1;
+    /**
+     * The software version of the state
+     */
+    proto.SemanticVersion software_version = 2;
+}
```
jsync-swirlds, line 355: Newline is required at EOF.

platform/state/platform_state.proto
```diff
@@ -339,3 +339,17 @@ message NodeId {
...
+
+/**
+ * A version of the platform software.
+ */
+message PlatformVersion {
+    /**
+     * Version of the HAPI specification that was used to serialize the state.
+     */
+    proto.SemanticVersion hapi_version = 1;
+    /**
```
jsync-swirlds, line 351: Should be a blank line between fields.

platform/state/platform_state.proto
```diff
@@ -339,3 +339,17 @@ message NodeId {
    */
   uint64 id = 1;
 }
+
+/**
+ * A version of the platform software.
```
jsync-swirlds, line 344:
```suggestion
 * A combined software version.<br/>
 * This message describes both the API version and the software version
 * supported by a release of the Hashgraph Platform SDK.
```

platform/state/platform_state.proto
```diff
@@ -339,3 +339,17 @@ message NodeId {
...
+ */
+message PlatformVersion {
+    /**
+     * Version of the HAPI specification that was used to serialize the state.
```
jsync-swirlds, line 348:
```suggestion
     * An API version.<br/>
     * This is the Hedera API specification supported by the release of
     * the Hashgraph Platform SDK that produced this message.
```

platform/state/platform_state.proto
```diff
@@ -339,3 +339,17 @@ message NodeId {
...
+     */
+    proto.SemanticVersion hapi_version = 1;
+    /**
+     * The software version of the state
```
jsync-swirlds, line 352:
```suggestion
     * A software version for the Hashgraph Platform SDK.<br/>
     * This is the version of the software that produced the state containing
     * this message.
```

jsync-swirlds (commented):
Mostly wording improvements.

platform/state/platform_state.proto
```diff
@@ -339,3 +339,17 @@ message NodeId {
...
 }
+
+/**
+ * A version of the platform software.
+ */
+message PlatformVersion {
+    /**
+     * Version of the HAPI specification that was used to serialize the state.
+     */
+    proto.SemanticVersion hapi_version = 1;
```
lpetrovic05, line 350: Conceptually, platform should not have any relation to HAPI. Should this be `app_version` & `platform_version`?
Btw, I think we need to sync on the future of software version, because events also contain a version inside

imalygin: >Conceptually, platform should not have any relation to HAPI.

Currently, the platform has HAPI dependency and, unless there is a major architectural change (which very well may be), it will remain this way. 

> Should this be app_version & platform_version?

That would introduce more confusion. I believe there is no notion of `platform_version` at the moment. 

>Btw, I think we need to sync on the future of software version, because events also contain a version inside

Yes. I agree with this. I wonder though if it can be a part of release/0.54. Also, I talked to @jasperpotts about this, and he mentioned that in the Block Node we took similar approach:

```
message BlockHeader {
    /**
     * A version of the HAPI specification.<br/>
     * This is the API version that was used to serialize the block.
     */
    proto.SemanticVersion hapi_proto_version = 1;

    /**
     * A version of the consensus node software.<br/>
     * This is the software version that executed the transactions
     * within this block.
     */
    proto.SemanticVersion software_version = 2;
```

platform/state/platform_state.proto
imalygin: Done

platform/state/platform_state.proto
imalygin: Done

imalygin:
After a conversation with @lpetrovic05 I decided to close this PR. 

The main reason for this PR was to provide adequate mapping from `com.hedera.node.app.version.HederaSoftwareVersion` to a PBJ object. It turns out that HAPI version doesn't need to be a part of the state at all and can be represented as a config parameter. Therefore, instead of having `PlatformVersion` object with two `SemanticVersion` objects, just a single `SemanticVersion` should be sufficient.

PR #408, opened 2024-08-12
Further block node updates

jsync-swirlds:
* Removed old "entity number" style messages from state changes
   * Replaced with recommended alternatives.
* Removed state change cause

Mirrored update for hedera-services is [PR 14827](https://github.com/hashgraph/hedera-services/pull/14827)

### DIFF
--- a/block/stream/output/state_changes.proto
+++ b/block/stream/output/state_changes.proto
@@ -45,7 +45,6 @@ import "state/addressbook/node.proto";
 import "state/blockrecords/block_info.proto";
 import "state/blockrecords/running_hashes.proto";
 import "state/blockstream/block_stream_info.proto";
-import "state/common.proto";
 import "state/congestion/congestion_level_starts.proto";
 import "state/consensus/topic.proto";
 import "state/contract/bytecode.proto";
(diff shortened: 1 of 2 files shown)


block/stream/output/state_changes.proto
```diff
@@ -63,45 +62,6 @@ import "state/token/token.proto";
...
-    STATE_CHANGE_CAUSE_END_OF_BLOCK = 2;
-
-    /**
-     * A set of data migration state changes.<br/>
-     * Data migration is typically necessary following a network upgrade to
-     * make required storage format updates and enable new features.
-     */
-    STATE_CHANGE_CAUSE_MIGRATION = 3;
-}
-
```
mattp-swirldslabs, line 104: Test to see if I can leave a comment

block/stream/output/state_changes.proto
```diff
@@ -63,45 +62,6 @@ import "state/token/token.proto";
...
 
-/**
- * The "cause" of a state change.
- *
- * What was the network doing that led to the change in state?<br/>
- * The most common event is, of course, a user transaction.
- * Other sources of state changes include end-of-block housekeeping changes,
- * data migration, and "system" transactions.
- */
-enum StateChangesCause {
```
mhess-swl, line 74: I'm sure there's a conversation I missed, but since I missed it–why are we removing `StateChangesCause`?

jsync-swirlds: It's unnecessary.  We don't really need the block node to know what caused a state change, we just need to record the state change and update the block-node copy of the state Merkle tree.

jsync-swirlds: Most of these changes are discussed on the HIP discussion page, if you're interested.

mhess-swl: What about state changes that don't come as a result of a specific transaction, e.g. the running hashes? I'm not sure we'll be able to pin the state changes currently with `CAUSE_END_OF_BLOCK` to anything concrete without the cause..

jsync-swirlds: The issue is that nobody really cares to know a cause.
It can be determined if an entity is _interested_ (I can describe how separately if needed), but it's not material to the block stream or block node.

mhess-swl (approved):

PR #410, opened 2024-08-20
Adding updates from hedera-services

jsync-swirlds:
* Transaction record entries replaced with transaction receipt entries in the block stream
* Platform state changes added to the block stream
* Additional response codes for airdrop and throttle failures added.


Neeharika-Sompalli (approved):
LGTM! Thanks @jsync-swirlds

PR #414, opened 2024-08-31
Topic Custom Fees

kimbor:
**Description**:
Create new message ConsensusCustomFee. 
Update ConsensusCreateTopic, ConsensusUpdateTopic, ConsensusTopicInfo to use it.

**Related issue(s)**:
Fixes #15144 in hedera-services

### DIFF
--- a/services/consensus_update_topic.proto
+++ b/services/consensus_update_topic.proto
@@ -28,6 +28,7 @@ option java_multiple_files = true;
 
 import "google/protobuf/wrappers.proto";
 import "basic_types.proto";
+import "custom_fees.proto";
 import "duration.proto";
 import "timestamp.proto";
 
@@ -87,4 +88,19 @@ message ConsensusUpdateTopicTransactionBody {
      * If unspecified, no change.
      */
     AccountID autoRenewAccount = 9;
+
+    /**
+      * Access control for update/delete of custom fees. Null if there is no key.
+      */
+    Key fee_schedule_key = 10;
+
+    /**
+      * If the transaction contains a signer from this list, no custom fees are applied.
+      */
+    repeated Key free_messages_key_list = 11;
+
+    /*
+     * The custom fee to be assessed during a message submission to this topic
+     */
+    repeated ConsensusCustomFee custom_fees = 12;
 }
--- a/services/custom_fees.proto
+++ b/services/custom_fees.proto
@@ -173,3 +173,15 @@ message AssessedCustomFee {
    */
   repeated AccountID effective_payer_account_id = 4;
 }
+
+message ConsensusCustomFee {
+  /**
+  * Fixed fee to be charged
+  */
+  FixedFee fixed_fee = 1;
+
+  /**
+  * The account to receive the custom fee
+  */
+  AccountID fee_collector_account_id = 2;
+}
--- a/services/state/consensus/topic.proto
+++ b/services/state/consensus/topic.proto
@@ -98,4 +98,16 @@ message Topic {
      * If present, enforces access control for message submission to the topic.
      */
     Key submit_key = 10;
+    /**
+     * If present, enforces access control for update/delete of this topic custom fees.
+     */
+    Key fee_schedule_key = 11;
+    /**
+      * If the transaction contains a signer from this list, no custom fees are applied.
+      */
+    repeated Key free_messages_key_list = 12;
+    /*
+     * The custom fee to be assessed during a message submission to this topic
+     */
+    repeated ConsensusCustomFee custom_fees = 13;
 }
--- a/services/state/consensus/topic.proto
+++ b/services/state/consensus/topic.proto
@@ -23,6 +23,7 @@ package proto;
  */
 
 import "basic_types.proto";
+import "custom_fees.proto";
 
 option java_package = "com.hederahashgraph.api.proto.java";
 // <<<pbj.java_package = "com.hedera.hapi.node.state.consensus">>> This comment is special code for setting PBJ Compiler java package
--- a/services/consensus_create_topic.proto
+++ b/services/consensus_create_topic.proto
@@ -27,6 +27,7 @@ option java_package = "com.hederahashgraph.api.proto.java";
 option java_multiple_files = true;
 
 import "basic_types.proto";
+import "custom_fees.proto";
 import "duration.proto";
 
 /**
@@ -70,4 +71,14 @@ message ConsensusCreateTopicTransactionBody {
      * If specified, there must be an adminKey and the autoRenewAccount must sign this transaction.
      */
     AccountID autoRenewAccount = 7;
+
+    /**
+      * Access control for update/delete of custom fees. Null if there is no key.
+      */
+    Key fee_schedule_key = 8;
+
+    /**
+      * The custom fee to be assessed during a message submission to this topic
+      */
+    repeated ConsensusCustomFee custom_fees = 9;
 }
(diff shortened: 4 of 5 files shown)


jsync-swirlds:
Which HIP is this change supporting?

services/consensus_update_topic.proto
```diff
@@ -87,4 +88,19 @@ message ConsensusUpdateTopicTransactionBody {
...
+    Key fee_schedule_key = 10;
+
+    /**
+      * If the transaction contains a signer from this list, no custom fees are applied.
+      */
+    repeated Key free_messages_key_list = 11;
+
+    /*
+     * The custom fee to be assessed during a message submission to this topic
+     */
```
jsync-swirlds, line 104: Text should specify behavior of the field and follow correct formatting:
```suggestion
    /**
     * A set of custom fee definitions.<br/>
     * These are fees to be assessed for each submit to this topic.
     * <p>
     * If this list is empty, the current set of custom fees
     * SHALL NOT change.<br/>
     * If this list is not empty, the current `fee_schedule_key` MUST sign
     * this transaction.<br/>
     * If this list is not empty, each fee defined in this set SHALL
     * be evaluated for each message submitted to this topic, and the
     * resultant total assessed fees SHALL be charged.<br/>
     * If this list is not empty, custom fees defined here SHALL be
     * charged _in addition to_ the base network and node fees.
     */
```
Note: I am guessing a bit as to behavior in this suggestion; please correct any errors above.

services/custom_fees.proto
```diff
@@ -173,3 +173,15 @@ message AssessedCustomFee {
...
+  /**
+  * Fixed fee to be charged
+  */
+  FixedFee fixed_fee = 1;
+
+  /**
+  * The account to receive the custom fee
+  */
+  AccountID fee_collector_account_id = 2;
+}
```
jsync-swirlds, line 187: 1. Why are we defining a whole new custom fee type here?  Would it not make more sense to use the existing custom fee structure?
1. There must be a newline at EOF.
1. Specification comments should specify behavior of the field and follow correct formatting
1. Messages must have specification comments that describe the use of that message and interactions with other systems (e.g. Block Stream), if any.
```suggestion
/**
 * A custom fee definition for a consensus topic.
 *
 * This fee definition is specific to an Hedera Consensus Service (HCS) topic
 * and SHOULD NOT be used in any other context.<br/>
 * All fields for this message are REQUIRED.<br/>
 * Only "fixed" fee definitions are supported because there is no basis for
 * a fractional fee on a consensus submit transaction.
 */
message ConsensusCustomFee {
    /**
     * A fixed custom fee.
     * <p>
     * The amount of HBAR or other token described by this `FixedFee` SHALL
     * be charged to the transction payer for each message submitted to a
     * topic that assigns this consensus custom fee.
     */
    FixedFee fixed_fee = 1;

    /**
     * A collection account identifier.
     * <p>
     * All amounts collected for this consensus custom fee SHALL be transferred
     * to the account identified by this field.
     */
    AccountID fee_collector_account_id = 2;
}

```
Note: I am guessing a bit as to behavior in this suggestion; please correct any errors above.

jsync-swirlds (changes requested):
There are some spacing and text issues, as well as a few questions about possibly missing fields.

ibankov:
> Which HIP is this change supporting?

https://hips.hedera.com/hip/hip-991

services/custom_fees.proto
Neurone: > 1) Why are we defining a whole new custom fee type here? Wouldn't it make more sense to use the existing custom fee structure?

We tried to recycle as much as possible, and considering the final schema we needed (a fixed fee + an account), we ended up with this definition.

There are two other messages that use FixedFee: CustomFee and RoyaltyFee. RoyaltyFee doesn't fit because of the data type, and CustomFee is too permissive (it accepts values we don't accept for this function).

But the HIP just suggests the proto definition. Do you see other ways we can define this?

jsync-swirlds: Most use of `CustomFee` applies point-specific restrictions.  We could allow `CustomFee` and specify that only a `FixedFee` is permitted for Consensus Topics.  This is not dissimilar to how different token types only allow certain types of `CustomFee` to apply.

We can also use the approach detailed here; it's just that we then have to modify a lot more of the system to handle special cases everywhere we might encounter a `ConsensusCustomFee` instead of just reusing all the existing custom fee logic.

I _believe_ it would be clearer, and easier to implement, if we use `CustomFee` for the Consensus Topic custom fees and just document that "This field MUST contain only `FixedFee` variants of `CustomFee`" on the custom fees field for `topicCreate` and `topicUpdate`.

Neurone: Personally, I still prefer a clean dedicated message, but if reusing `CustomFee` is consistent with what we're already doing elsewhere, it works for me.

However, it's important to note that with the same HIP, we're going to create a new allowance transaction type to avoid cross-dependencies between services, so using a dedicated fee message can be seen as consistent with that goal.

We need @netopyr's decision on this.

jsync-swirlds: I'll read that HIP for the new transaction type, thanks for surfacing that.

Neurone: Thanks to you! Please wait for the final version of the doc, the definition of the new tx is still missing from the current visible HIP. We will update the HIP text after we defined all the details.

services/state/consensus/topic.proto
```diff
@@ -98,4 +99,16 @@ message Topic {
...
+     */
+    Key fee_schedule_key = 11;
+    /**
+      * If the transaction contains a signer from this list, no custom fees are applied.
+      */
+    repeated Key free_messages_key_list = 12;
+    /*
+     * The custom fee to be assessed during a message submission to this topic
+     */
+    repeated ConsensusCustomFee custom_fees = 13;
```
jsync-swirlds, line 113: Text should specify behavior of the field and follow correct formatting:
```suggestion
    /**
     * Access control for update/delete of custom fees.
     * <p>
     * If this field is unset, the current custom fees CANNOT be changed.<br/>
     * If this field is set, that `Key` MUST sign any transaction to update
     * the custom fee schedule for this topic.
     */
    Key fee_schedule_key = 11;

    /**
     * A list of "privileged payer" keys.
     * <p>
     * If a submit transaction is signed by _any_ key from this list,
     * custom fees SHALL NOT be charged for that transaction.<br/>
     * This list SHALL _implicitly_ contain the topic `admin_key`,
     * the topic `submit_key`, and the topic `fee_schedule_key`, if any
     * of those keys is set.
     */
    repeated Key free_messages_key_list = 12;

    /**
     * A set of custom fee definitions.<br/>
     * These are fees to be assessed for each submit to this topic.
     * <p>
     * If this list is empty, the only fees charged for a submit to this
     * topic SHALL be the network and node fees.<br/>
     * If this list is not empty, each fee defined in this set SHALL
     * be evaluated for each message submitted to this topic, and the
     * resultant total assessed fees SHALL be charged.<br/>
     * If this list is not empty, custom fees defined here SHALL be
     * charged _in addition to_ the base network and node fees.
     */
    repeated ConsensusCustomFee custom_fees = 13;
```
Note: I am guessing a bit as to behavior in this suggestion; please correct any errors above.

kimbor:
Closing this PR in favor of https://github.com/hashgraph/hedera-services/pull/15350. The changes requested here have been incorporated into that PR.

jsync-swirlds:
> Closing this PR in favor of [hashgraph/hedera-services#15350](https://github.com/hashgraph/hedera-services/pull/15350). The changes requested here have been incorporated into that PR.

A PR will still be needed in this repo to keep the two copies in sync.

PR #415, opened 2024-09-04
chore: add needed `TransactionOutput` fields

tinker-michaelj:
**Description**:
 - Adds `TransactionOutput` fields for `SCHEDULE_CREATE`, `TOKEN_AIRDROP`, and `ETHEREUM_TRANSACTION` that are necessary to recover the full record stream from the block stream.

### DIFF
--- a/block/stream/output/schedule_service.proto
+++ b/block/stream/output/schedule_service.proto
@@ -65,6 +65,17 @@ import "basic_types.proto";
  */
 message CreateScheduleOutput {
     /**
+     * A schedule identifier.
+     * <p>
+     * If the status of the transaction is `SUCCESS`, this value SHALL be
+     * set to the identifier of the schedule that was created. If the
+     * transaction status is `IDENTICAL_SCHEDULE_ALREADY_CREATED`, this
+     * value SHALL be set to the identifier of the existing schedule that
+     * is identical to the one that was attempted to be created. For any
+     * other status, this value SHALL NOT be set.
+     */
+    proto.ScheduleID schedule_id = 1;
+    /**
      * A scheduled transaction identifier.
      * <p>
      * This value SHALL be the transaction identifier for the _scheduled_
--- a/block/stream/output/smart_contract_service.proto
+++ b/block/stream/output/smart_contract_service.proto
@@ -135,7 +135,27 @@ message EthereumOutput {
     /**
      * An ethereum hash value.
      * <p>
-     * This SHALL be a keccak256 hash of the ethereumData.
+     * This SHALL be a keccak256 hash of the ethereumData, after
+     * interpolating the calldata from the referenced calldata file
+     * (if set).
      */
     bytes ethereum_hash = 2;
+
+    oneof eth_result {
+        /**
+         * A result for an Ethereum Transaction executed as a call.
+         * <p>
+         * This field SHALL contain all of the data produced by the contract
+         * call transaction as well as basic accounting results.
+         */
+        proto.ContractFunctionResult ethereum_call_result = 3;
+
+        /**
+         * A result for an Ethereum Transaction executed as a create.
+         * <p>
+         * This field SHALL contain all of the data produced by the contract
+         * create transaction as well as basic accounting results.
+         */
+        proto.ContractFunctionResult ethereum_create_result = 4;
+    }
 }
--- a/block/stream/output/token_service.proto
+++ b/block/stream/output/token_service.proto
@@ -34,6 +34,9 @@ option java_package = "com.hedera.hapi.block.stream.output.protoc";
 // <<<pbj.java_package = "com.hedera.hapi.block.stream.output">>> This comment is special code for setting PBJ Compiler java package
 option java_multiple_files = true;
 
+import "custom_fees.proto";
+import "transaction_record.proto";
+
 /**
  * Block Stream data for a `createToken` transaction.
  *
@@ -161,3 +164,31 @@ message UnpauseTokenOutput {}
  * the original transaction.
  */
 message UpdateTokenNftsOutput {}
+
+/**
+ * Block Stream data for a `tokenAirdrop` transaction.
+ *
+ * This message SHALL NOT duplicate information already contained in
+ * the original transaction.
+ */
+message TokenAirdropOutput {
+  /**
+   * Custom fees assessed during a TokenAirdrop.
+   * <p>
+   * These fees SHALL be present in the full transfer list for the transaction.
+   */
+  repeated proto.AssessedCustomFee assessed_custom_fees = 1;
+
+  /**
+   * A list of pending token airdrops.
+   * Each pending airdrop represents a single requested transfer from a
+   * sending account to a recipient account. These pending transfers are
+   * issued unilaterally by the sending account, and MUST be claimed by the
+   * recipient account before the transfer MAY complete.
+   * A sender MAY cancel a pending airdrop before it is claimed.
+   * An airdrop transaction SHALL emit a pending airdrop when the recipient has no
+   * available automatic association slots available or when the recipient
+   * has set `receiver_sig_required`.
+   */
+  repeated proto.PendingAirdropRecord new_pending_airdrops = 2;
+}
--- a/services/state/blockstream/block_stream_info.proto
+++ b/services/state/blockstream/block_stream_info.proto
@@ -59,8 +59,8 @@ message BlockStreamInfo {
 
     /**
      * A consensus time for the current block.<br/>
-     * This is the _first_ consensus time in the current block, and
-     * is used to determine if this block was the first across an
+     * This is the consensus time of the first round in the current block,
+     * and is used to determine if this block was the first across an
      * important boundary in consensus time, such as UTC midnight.
      * This may also be used to purge entities expiring between the last
      * block time and this time.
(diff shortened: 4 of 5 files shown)


block/stream/output/schedule_service.proto
```diff
@@ -64,6 +64,17 @@ import "basic_types.proto";
...
 message CreateScheduleOutput {
+    /**
+     * A schedule identifier.
+     * <p>
+     * If the status of the transaction is `SUCCESS`, this value SHALL be
+     * set to the identifier of the schedule that was created. If the
+     * transaction status is `IDENTICAL_SCHEDULE_ALREADY_CREATED`, this
+     * value SHALL be set to the identifier of the existing schedule that
+     * is identical to the one that was attempted to be created. For any
+     * other status, this value SHALL NOT be set.
```
jsync-swirlds, line 75: This should have some line breaks for readability.
```suggestion
     * If the status of the transaction is `SUCCESS`, this value SHALL be
     * set to the identifier of the schedule that was created.<br/>
     * If the transaction status is `IDENTICAL_SCHEDULE_ALREADY_CREATED`,
     * this value SHALL be set to the identifier of an existing schedule
     * that is identical to the one that was requested.<br/>
     * For any other status, this value SHALL NOT be set.
```

block/stream/output/smart_contract_service.proto
```diff
@@ -135,7 +135,27 @@ message EthereumOutput {
     /**
      * An ethereum hash value.
      * <p>
-     * This SHALL be a keccak256 hash of the ethereumData.
+     * This SHALL be a keccak256 hash of the ethereumData, after
+     * interpolating the calldata from the referenced calldata file
+     * (if set).
```
jsync-swirlds, line 140:
```suggestion
     * This SHALL be a keccak256 hash of the ethereumData<br/>
     * This hash SHALL be computed strictly _after_ interpolating
     * the calldata from the calldata file (if set).
```

block/stream/output/smart_contract_service.proto
```diff
@@ -135,7 +135,27 @@ message EthereumOutput {
...
      */
     bytes ethereum_hash = 2;
+
+    oneof eth_result {
+        /**
+         * A result for an Ethereum Transaction executed as a call.
```
jsync-swirlds, line 146:
```suggestion
         * A result for an Ethereum _contract call_ transaction.
```

block/stream/output/smart_contract_service.proto
```diff
@@ -135,7 +135,27 @@ message EthereumOutput {
...
+         * <p>
+         * This field SHALL contain all of the data produced by the contract
+         * call transaction as well as basic accounting results.
+         */
+        proto.ContractFunctionResult ethereum_call_result = 3;
+
+        /**
+         * A result for an Ethereum Transaction executed as a create.
```
jsync-swirlds, line 154:
```suggestion
         * A result for an Ethereum _contract create_ transaction.
```

block/stream/output/token_service.proto
```diff
@@ -161,3 +164,31 @@ message UnpauseTokenOutput {}
...
+  /**
+   * A list of pending token airdrops.
+   * Each pending airdrop represents a single requested transfer from a
+   * sending account to a recipient account. These pending transfers are
+   * issued unilaterally by the sending account, and MUST be claimed by the
+   * recipient account before the transfer MAY complete.
+   * A sender MAY cancel a pending airdrop before it is claimed.
+   * An airdrop transaction SHALL emit a pending airdrop when the recipient has no
+   * available automatic association slots available or when the recipient
+   * has set `receiver_sig_required`.
```
jsync-swirlds, line 191: Line breaks and slight rewording for clarity and to match style.
```suggestion
   * Each entry in this list SHALL represent a single requested transfer.<br/>
   * Each entry SHALL specify a single sending account and a
   * single recipient account.<br/>
   * Each pending airdrop SHALL be requested unilaterally by the sending
   * account, and MUST be claimed by the recipient account before the
   * transfer can be completed.<br/>
   * A sender MAY cancel a pending airdrop before it is claimed.<br/>
   * An airdrop transaction SHALL emit a pending airdrop when the recipient
   * has no available automatic association slots available or when the
   * recipient has set `receiver_sig_required`.
```

block/stream/output/token_service.proto
```diff
@@ -161,3 +164,31 @@ message UnpauseTokenOutput {}
...
  */
 message UpdateTokenNftsOutput {}
+
+/**
+ * Block Stream data for a `tokenAirdrop` transaction.
+ *
+ * This message SHALL NOT duplicate information already contained in
+ * the original transaction.
+ */
+message TokenAirdropOutput {
```
jsync-swirlds, line 174: I do not think we need this.
The State Changes entries will list every created `PendingAirdrop`.
The transfer list will contain the custom fees.

block/stream/output/schedule_service.proto
```diff
@@ -64,6 +64,17 @@ import "basic_types.proto";
...
+     */
+    proto.ScheduleID schedule_id = 1;
```
jsync-swirlds, line 77: I do not believe this is needed.  The new `Schedule` that is created will be present as a StateChange and will contain the ID.

services/state/blockstream/block_stream_info.proto
```diff
@@ -72,6 +72,7 @@ message BlockStreamInfo {
      * This combines several trailing output block item hashes and
      * is used as a seed value for a pseudo-random number generator.<br/>
      * This is also requiried to implement the EVM `PREVRANDAO` opcode.
```
jsync-swirlds, line 74: Missing a line break here.
```suggestion
     * This is also requiried to implement the EVM `PREVRANDAO` opcode.<br/>
```

jsync-swirlds (commented):
I am not sure we actually need the new outputs.
We deliberately do _not_ include values that are present in some other block item in outputs, so created ID values (which must be present in the state changes) or messages added to state as part of a transaction (i.e. PendingAirdrop messages) are not included in transaction outputs, for example.

block/stream/output/schedule_service.proto
tinker-michaelj: In the case of an `IDENTICAL_SCHEDULE_ALREADY_CREATED`, the existing schedule's id is set in the receipt; but that existing schedule will not be changed in the transaction.

block/stream/output/token_service.proto
tinker-michaelj: Just as with `CRYPTO_TRANSFER`, it is quite onerous to expect block stream consumers to be able to infer what part of the `TokenTransferList` is due to assessed custom fees; and we need to this recover the record stream.

For a top-level token airdrop, it's definitely possible to identify the created airdrops; but for multiple child airdrop transactions it would get tricky.

But maybe we can cross that bridge when we come to it... 🤔

block/stream/output/schedule_service.proto
jsync-swirlds: That makes sense.
Worth noting that it is very important to avoid sending any data unless strictly necessary in BlockStream.  Every byte has a permanent ongoing cost to store (there is no purge or delete for block stream), so the less we write the less we end up committing to spend forever.

If we can avoid setting this (e.g. in every case _except_ duplicate), we should.

block/stream/output/token_service.proto
jsync-swirlds: All of the state changes for a particular transaction follow that transaction in the stream.  It's definitely possible to connect the dots, where needed.
There is a bit of mild obsession in the BlockStream design with minimizing data because of the "eternal" cost associated with block stream storage that is never truly offset by one-time transaction fees for the transaction that creates that cost.

jsync-swirlds: Side note, I think we probably _should not_ try to recreate a record stream from a block stream.  It's very likely to be somewhere between extremely hard and impossible, in my opinion.

Block stream is very deliberately a different structure so that we can dramatically reduce data volume, unify several different streams, provide state proof services from block nodes instead of consensus nodes, and efficiently prove the authenticity of each block.

block/stream/output/schedule_service.proto
tinker-michaelj: Ah yes that is very sensible, will make sure it works that way. 💯

block/stream/output/token_service.proto
tinker-michaelj: Makes sense; removed the `new_pending_airdrops` field from this message.

block/stream/output/schedule_service.proto
```diff
@@ -64,6 +64,17 @@ import "basic_types.proto";
...
+     * A schedule identifier.
+     * <p>
+     * If the status of the transaction is `SUCCESS`, this value SHALL be
+     * set to the identifier of the schedule that was created.<br/>
+     * If the transaction status is `IDENTICAL_SCHEDULE_ALREADY_CREATED`,
+     * this value SHALL be set to the identifier of an existing schedule
+     * that is identical to the one that was requested.<br/>
+     * For any other status, this value SHALL NOT be set.
+     */
+    proto.ScheduleID schedule_id = 1;
```
jsync-swirlds, line 77: Should this be renamed to `duplicate_schedule_id`?
Also, if we only set this for duplicate schedules, we may need to remove the first specification sentence in the document comment above.

jsync-swirlds (commented):
Looks good, one small question remains.

PR #420, opened 2024-09-30
feat: sync up services repo with proto for release 55

povolev15:
sync up services repo with proto for release 55

**Related issue(s)**:

Fixes #15607

### DIFF
--- a/block/stream/output/schedule_service.proto
+++ b/block/stream/output/schedule_service.proto
@@ -65,6 +65,17 @@ import "basic_types.proto";
  */
 message CreateScheduleOutput {
     /**
+     * A schedule identifier.
+     * <p>
+     * If the status of the transaction is `SUCCESS`, this value SHALL be
+     * set to the identifier of the schedule that was created. If the
+     * transaction status is `IDENTICAL_SCHEDULE_ALREADY_CREATED`, this
+     * value SHALL be set to the identifier of the existing schedule that
+     * is identical to the one that was attempted to be created. For any
+     * other status, this value SHALL NOT be set.
+     */
+    proto.ScheduleID schedule_id = 1;
+    /**
      * A scheduled transaction identifier.
      * <p>
      * This value SHALL be the transaction identifier for the _scheduled_
--- a/block/stream/output/smart_contract_service.proto
+++ b/block/stream/output/smart_contract_service.proto
@@ -132,10 +132,23 @@ message EthereumOutput {
      */
     repeated proto.TransactionSidecarRecord sidecars = 1;
 
-    /**
-     * An ethereum hash value.
-     * <p>
-     * This SHALL be a keccak256 hash of the ethereumData.
-     */
     bytes ethereum_hash = 2;
+
+    oneof eth_result {
+        /**
+         * A result for an Ethereum Transaction executed as a call.
+         * <p>
+         * This field SHALL contain all of the data produced by the contract
+         * call transaction as well as basic accounting results.
+         */
+        proto.ContractFunctionResult ethereum_call_result = 3;
+
+        /**
+         * A result for an Ethereum Transaction executed as a create.
+         * <p>
+         * This field SHALL contain all of the data produced by the contract
+         * create transaction as well as basic accounting results.
+         */
+        proto.ContractFunctionResult ethereum_create_result = 4;
+    }
 }
--- a/block/stream/output/token_service.proto
+++ b/block/stream/output/token_service.proto
@@ -34,6 +34,9 @@ option java_package = "com.hedera.hapi.block.stream.output.protoc";
 // <<<pbj.java_package = "com.hedera.hapi.block.stream.output">>> This comment is special code for setting PBJ Compiler java package
 option java_multiple_files = true;
 
+import "custom_fees.proto";
+import "transaction_record.proto";
+
 /**
  * Block Stream data for a `createToken` transaction.
  *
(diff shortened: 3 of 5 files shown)


block/stream/output/schedule_service.proto
```diff
@@ -64,6 +64,17 @@ import "basic_types.proto";
...
+     * A schedule identifier.
+     * <p>
+     * If the status of the transaction is `SUCCESS`, this value SHALL be
+     * set to the identifier of the schedule that was created. If the
+     * transaction status is `IDENTICAL_SCHEDULE_ALREADY_CREATED`, this
+     * value SHALL be set to the identifier of the existing schedule that
+     * is identical to the one that was attempted to be created. For any
+     * other status, this value SHALL NOT be set.
+     */
+    proto.ScheduleID schedule_id = 1;
```
jsync-swirlds, line 77: This field is unnecessary, and was expressly removed from the block stream HIP.
The schedule that is created will be present in the State Changes, including the ID.

block/stream/output/smart_contract_service.proto
```diff
@@ -132,10 +132,23 @@ message EthereumOutput {
      */
     repeated proto.TransactionSidecarRecord sidecars = 1;
 
-    /**
-     * An ethereum hash value.
-     * <p>
-     * This SHALL be a keccak256 hash of the ethereumData.
-     */
```
jsync-swirlds, line 139: Why was this removed?

block/stream/output/smart_contract_service.proto
```diff
@@ -132,10 +132,23 @@ message EthereumOutput {
...
+        proto.ContractFunctionResult ethereum_call_result = 3;
+
+        /**
+         * A result for an Ethereum Transaction executed as a create.
+         * <p>
+         * This field SHALL contain all of the data produced by the contract
+         * create transaction as well as basic accounting results.
+         */
+        proto.ContractFunctionResult ethereum_create_result = 4;
+    }
```
jsync-swirlds, line 153: We already have the `CallContractOutput` and `CreateContractOutput` items, why are we adding the same data again here? A "canonical" Block Stream would simply follow this `EthereumOutput` with one of those items if necessary.

block/stream/output/token_service.proto
```diff
@@ -34,6 +34,9 @@ option java_package = "com.hedera.hapi.block.stream.output.protoc";
 // <<<pbj.java_package = "com.hedera.hapi.block.stream.output">>> This comment is special code for setting PBJ Compiler java package
 option java_multiple_files = true;
 
+import "custom_fees.proto";
+import "transaction_record.proto";
```
jsync-swirlds, line 38: This import line is unnecessary and incorrect.

services/state/blockstream/block_stream_info.proto
```diff
@@ -48,7 +48,9 @@ option java_multiple_files = true;
  * This value MUST be transmitted in the "state changes" section of
  * _each_ block, but MUST be updated at the beginning of the _next_ block.<br/>
  * This value SHALL contain the block hash up to, and including, the
- * immediately prior completed block.
+ * immediately prior completed block.<br/>
+ * The state change to update this singleton MUST be the last "output"
```
jsync-swirlds, line 52: `"output"` can be dropped, it must be the last block item in the block; period.  The next item of any type _must_ be the `BlockProof`.

services/state/blockstream/block_stream_info.proto
```diff
@@ -72,6 +74,7 @@ message BlockStreamInfo {
      * This combines several trailing output block item hashes and
      * is used as a seed value for a pseudo-random number generator.<br/>
      * This is also requiried to implement the EVM `PREVRANDAO` opcode.
```
jsync-swirlds, line 76: Missing `<br/>` at the end of this line.

services/state/blockstream/block_stream_info.proto
```diff
@@ -83,6 +86,49 @@ message BlockStreamInfo {
...
+    uint32 num_preceding_output_items = 7;
+
+    /**
+     * A concatenation of SHA2-384 hash values.<br/>
+     * This is the "rightmost" values of the "output" subtree.
+     * <p>
+     * The subtree containing these hashes SHALL be constructed from all "output"
+     * `BlockItem`s in this block that _precede_ the update to this singleton.
+     */
+    repeated bytes rightmost_preceding_output_tree_hashes = 8;
```
jsync-swirlds, line 125: The "output" tree is strictly binary, I would expect that this only ever needs to be a single hash, but perhaps I'm missing something in what this is supposed to be.

Also, wouldn't "preceding" items be "left" of the current item?

jsync-swirlds (changes requested):
There are some items added here that shouldn't be present, and some that _probably_ shouldn't be added.
Also, some of the specification text is a little off.

block/stream/output/schedule_service.proto
tinker-michaelj: We discussed this before @jsync-swirlds , as per the comment there isn't a corresponding `StateChanges` item when the `ScheduleCreate` result is `IDENTICAL_SCHEDULE_ALREADY_CREATED`.

block/stream/output/smart_contract_service.proto
tinker-michaelj: Ah that is an interesting idea @jsync-swirlds  and would work perfectly well. Created https://github.com/hashgraph/hedera-services/issues/15619 and self-assigned.

block/stream/output/schedule_service.proto
jsync-swirlds: I had forgotten.  Thanks for the reminder.

services/state/blockstream/block_stream_info.proto
Neeharika-Sompalli: These are all the right hashes of the binary tree needed to compute indirect block proof.

PR #421, opened 2024-10-10
Modifications to Block Stream based on recent work

jsync-swirlds:
### Modify block service definition to allow repeated block items
* Modified `PublishStreamRequest` to permit repeated block items.
* Added a `SubscribeStreamResponseSet` wrapper
   * This works around that `oneof` does not permit `repeated` fields.
* Modified `SubscribeStreamResponseSet` to have a `repeated` `BlockItem` list.

### Make the gRPC service modular
* Modify gRPC service design.
   * Replace single service with four specific services for modularity

Resolves https://github.com/hashgraph/hedera-block-node/issues/259

Note: A mirror PR will be added to hedera-services once this PR is complete.

### DIFF
--- a/block/block_service.proto
+++ b/block/block_service.proto
@@ -58,12 +58,135 @@ import "stream/block_item.proto";
  * ensure all data was received and processed correctly.
  */
 message PublishStreamRequest {
+    oneof request {
+        /**
+         * A stream item containing one or more `BlockItem`s.
+         * <p>
+         * The full stream SHALL consist of many `block_items` messages
+         * followed by a single `status` message.
+         */
+        BlockItemSet block_items = 1;
+
+        /**
+         * An item sent when a stream is ended by the publisher.
+         */
+        EndOfStream status = 2;
+    }
+
     /**
-     * An ordered list of items written to the block stream.
+     * A message sent to end a stream.
+     *
+     * This request message SHALL be sent from a Publisher to a block
+     * node system when a `publishBlockStream` stream ends.<br/>
+     * This message SHALL be sent _at most_ once for each `publishBlockStream`
+     * call.<br/>
+     * The source system SHALL cease sending block items after sending
+     * this item.
+     */
+    message EndOfStream {
+        /**
+         * A status code.
+         * <p>
+         * This code indicates the reason the stream ended.<br/>
+         * This value MUST be set to a non-default value.
+         */
+        PublishStreamEndCode status = 1;
+
+        /**
+         * The number of the earliest available block.
+         * <p>
+         * Publishers MUST provide the earliest available block still
+         * present in local cache.
+         */
+        uint64 earliest_block_number = 2;
+
+        /**
+         * The number of the last available block.
+         * <p>
+         * Publishers MUST provide the most recent available block.
+         */
+        uint64 latest_block_number = 3;
+    }
+}
+
+/**
+ * An enumeration indicating why a publisher ended a stream.
+ *
+ * This enumeration describes the reason a block stream
+ * (sent via `publishBlockStream`) was ended by the publisher.
+ */
+enum PublishStreamEndCode {
+    /**
+     * An "unset value" flag, this value SHALL NOT be used.<br/>
+     * This status indicates the server software failed to set a
+     * status, and SHALL be considered a software defect.
+     */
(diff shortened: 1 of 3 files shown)


block/block_service.proto
```diff
@@ -539,15 +517,41 @@ message SubscribeStreamResponse {
...
+    /**
+     * An ordered list of `BlockItem`s.<br/>
+     * This list supports sending block items to subscribers in batches
+     * for greater channel efficiency.
+     * <p>
+     * The full response SHALL consist of many of these collections
+     * followed by a single `status` message.
+     */
+    repeated com.hedera.hapi.block.stream.BlockItem block_items = 1;
+}
```
AlfredoG87, line 553: As is my understanding we are only going to send a "Response" once a whole block is received. Am I right?
The why we need to send the whole block_items back as part of the Response, it would not be more effective to have a response object that is only a block number instead of streaming back all the items again?

jsync-swirlds: We are not guaranteed to send only whole blocks back, as the low-latency unverified stream requires data to flow through immediately.

AlfredoG87: I see, I was mixing the clients 😅 
Thank you!

derektriley (dismissed):
LGTM.

petreze (dismissed):

jsync-swirlds:
Rebased to main, and modified the publish request structure based on protocol discussions.
Please re-review.

AlfredoG87 (approved):

petreze (approved):

tinker-michaelj (approved):
Makes sense to me, tyvm @jsync-swirlds!

block/block_service.proto
```diff
@@ -71,23 +38,157 @@ import "stream/block.proto";
...
+ * If a `BlockHeader` is present in a request, it MUST be the first `BlockItem`
+ * in the `block_items` list.<br/>
  * The block node SHALL append each `BlockItem` to an internal structure
  * to reconstruct full blocks.<br/>
  * The block node MUST verify the block proof for each block before sending a
  * response message acknowledging that block.<br/>
- * Each Block MUST end with a single `BlockStateProof` block item.<br/>
- * The block node MUST verify the Block using the `BlockStateProof` to
- * ensure all data was received and processed correctly.<br/>
+ * Each Block MUST end with a single `BlockProof` block item.<br/>
```
mhess-swl, line 54: This is true because we will only buffer a block without a proof on the node, but won't send it to a block node, correct?

block/block_service.proto
```diff
@@ -71,23 +38,157 @@ import "stream/block.proto";
...
+ * If a `BlockProof` is present in a request, it MUST be the last `BlockItem`
```
mhess-swl, line 55: Does this mean that only one full block can be sent in a request? I don't think that's what this means, but the verbiage implies a maximum of one `BlockProof`

block/block_service.proto
jsync-swirlds: The block data will be sent in batches to the block node as data is available.  What we require here is that the Block Proof _MUST_ arrive _before_ the Block Header for the _next_ block.
The Consensus Node should _not_ buffer any more than necessary for efficiency and correct ordering.

block/block_service.proto
jsync-swirlds: A request should almost never be a full block.  A single streamed request (the API is a stream of request messages) is 0 or more `BlockItem`s.  What we require here is that _if_ a `BlockProof` is present in the batch, it _must_ be the _last_ item.  This is so that we _never_ include more than one block in a single batch.  The Block Node has to process this data _without parsing_ to the maximum extent possible, so we require that no single batch of block items will _ever_ contain data from more than one block.
A block _must_ begin with a BlockHeader, and _must_ end with a BlockProof.  By requiring that a Block Header is _first_ in a batch, and a Block Proof is _last_ in a (usually different) batch, we ensure no single batch contains items from 2 or more blocks.

block/block_service.proto
jsync-swirlds: Note, the protocol/handshake between a _publisher_ (e.g. a consensus node) and a _block node_ is defined [here](https://github.com/hashgraph/hedera-block-node/blob/main/docs/Protocol/publishBlockStream.md).

PR #423, opened 2024-10-16
fix: broken import on `services/transaction_body.proto`

### LINKED ISSUE #422: Bug: Wrong import on `services/transaction_body.proto`
### Description

When using the latest version (`main` branch) of the protobuf on the `block node` the generation process fails with the following error:

```
Import "auxiliary/tss/tss_message.proto" in proto file "/Users/user/Projects/hedera-block-node/stream/build/hedera-protobufs/services/transaction_body.proto" can not be found in src files.

```

### Steps to reproduce

Try to Generate using `./gradlew :stream:assemble`  on the block node repo.

### Additional context

image

issue was introduced in this [PR](https://github.com/hashgraph/hedera-protobufs/commit/4c8fa91d1f625a5098dbc28e9dd74de6c9328790)

### Hedera network

other

### Version

v0.21.0-SNAPSHOT

### Operating system

None

AlfredoG87:
**Description**:
fix path to tss_message and tss_vote proto files from services/transaction_body.proto

**Related issue(s)**:

Fixes #422 

**Notes for reviewer**:

**Checklist**


AlfredoG87:
@thomas-swirlds-labs could you check if this is the correct way to fix this bug? I understand that it was also possible to create a `auxiliary/tss` folder.

thomas-swirlds-labs:
> @thomas-swirlds-labs could you check if this is the correct way to fix this bug? I understand that it was also possible to create a `auxiliary/tss` folder.

@AlfredoG87 - The best way would be to create a `auxilliary/tss` folder. This will sync it up correctly with Services. Then `transaction_body` will have to be updated to reflect the same. Good catch, Thanks!

jsync-swirlds (approved):

AlfredoG87:
> 
> @thomas-swirlds-labs could you check if this is the correct way to fix this bug? I understand that it was also possible to create a auxiliary/tss folder.
> 
> @AlfredoG87 - The best way would be to create a auxilliary/tss folder. This will sync it up correctly with Services. Then transaction_body will have to be updated to reflect the same. Good catch, Thanks!
> 

@thomas-swirlds-labs, re-do the change as you suggested and tested it and works :)

thomas-swirlds-labs (approved):
LGTM. Thanks! @AlfredoG87 🎉

PR #429, opened 2024-10-24
chore: Copied recent protobuf changes from hedera-services

jsync-swirlds:
* Copied recent changes
* Fixed several small errors (incorrect indentation, malformed break tags, etc...)
* Applied temporary workaround for PBJ bug 280

### DIFF
--- a/block/block_service.proto
+++ b/block/block_service.proto
@@ -70,43 +70,45 @@ message PublishStreamRequest {
         /**
          * An item sent when a stream is ended by the publisher.
          */
-        EndOfStream status = 2;
+        // @workaround(PBJ #280) REMOVED pending fix for PBJ bug 280
+        //EndOfStream status = 2;
     }
 
-    /**
-     * A message sent to end a stream.
-     *
-     * This request message SHALL be sent from a Publisher to a block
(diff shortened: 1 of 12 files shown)


block/block_service.proto
```diff
@@ -70,43 +70,45 @@ message PublishStreamRequest {
         /**
          * An item sent when a stream is ended by the publisher.
          */
-        EndOfStream status = 2;
+        // @workaround(PBJ #280) REMOVED pending fix for PBJ bug 280
```
Neeharika-Sompalli, line 73: We will enable this in later PR ?

Neeharika-Sompalli (approved):
Have one question. LGTM otherwise. Thanks @jsync-swirlds

block/block_service.proto
jsync-swirlds: Yes, once the PBJ bug (as noted) is fixed (it's been waiting on resources since June), we'll add this back in.  Hopefully within a month or two (it's not a trivial bug, it requires some pretty large improvements to PBJ's code generation).

kimbor (approved):

PR #431, opened 2024-10-30
Sync hedera-services pr16144 changes.

iwsimon:
**Description**:

This is to sync hedera-services PR 16144 changes. node_update.proto is not included here as there are changes from pr16131 not synced yet.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/block/stream/output/state_changes.proto
+++ b/block/stream/output/state_changes.proto
@@ -354,6 +354,16 @@ enum StateIdentifier {
     STATE_ID_ROSTERS = 28;
 
     /**
+    * A state identifier for scheduled transaction expiration.
+    */
+    STATE_ID_SCHEDULE_IDS_BY_EXPIRY = 29;
+
+    /**
+     * A state identifier for scheduled transaction deduplication.
+     */
+    STATE_ID_SCHEDULE_ID_BY_EQUALITY = 30;
+
+    /**
      * A state identifier for the round receipts queue.
      */
     STATE_ID_TRANSACTION_RECEIPTS_QUEUE = 126;
@@ -795,6 +805,16 @@ message MapChangeValue {
          * A roster value.
          */
         com.hedera.hapi.node.state.roster.Roster roster_value = 16;
+
+        /**
+         * A list of scheduled ids.
+         */
+        proto.ScheduleIdList schedule_id_list_value = 17;
+
+        /**
+         * A scheduled id value.
+         */
+        proto.ScheduleID schedule_id_value = 18;
     }
 }
 
--- a/services/state/schedule/schedule.proto
+++ b/services/state/schedule/schedule.proto
@@ -141,8 +141,22 @@ message Schedule {
  * simplified hash code.
  */
 message ScheduleList {
-  /**
-   * a list of schedules, in no particular order.
-   */
-  repeated Schedule schedules = 1;
+    /**
+     * a list of schedules, in no particular order.
+     */
+    repeated Schedule schedules = 1;
+}
+
+/**
+ * A message for storing a list of schedule identifiers in state.<br/>
+ * This is used to store lists of `ScheduleID` values.
+ * One example is all schedules that expire at a particular time.
+ */
+message ScheduleIdList {
+    /**
+     * A list of schedule identifiers, in no particular order.
+     * <p>
+     * While the order is not _specified_, it MUST be deterministic.
+     */
+    repeated ScheduleID schedule_ids = 1;
 }


services/state/schedule/schedule.proto
```diff
@@ -141,8 +141,22 @@ message Schedule {
  * simplified hash code.
  */
 message ScheduleList {
-  /**
-   * a list of schedules, in no particular order.
-   */
-  repeated Schedule schedules = 1;
+    /**
+     * a list of schedules, in no particular order.
+     */
```
jsync-swirlds, line 146:
```suggestion
    /**
     * A list of schedules, in no particular order.
     * <p>
     * While the order is not _specified_, it MUST be deterministic.
     */
```

block/stream/output/state_changes.proto
```diff
@@ -795,6 +805,16 @@ message MapChangeValue {
          * A roster value.
          */
         com.hedera.hapi.node.state.roster.Roster roster_value = 16;
+
+        /**
+         * A list of scheduled ids.
```
jsync-swirlds, line 810:
```suggestion
         * A list of schedule identifiers.
```

block/stream/output/state_changes.proto
```diff
@@ -795,6 +805,16 @@ message MapChangeValue {
...
+         */
+        proto.ScheduleIdList schedule_id_list_value = 17;
```
jsync-swirlds, line 812: I _believe_ this can **replace** field number 8, as we do not store lists of full schedules any longer.

block/stream/output/state_changes.proto
```diff
@@ -795,6 +805,16 @@ message MapChangeValue {
...
+
+        /**
+         * A scheduled id value.
```
jsync-swirlds, line 815:
```suggestion
         * A schedule identifier value.
```

block/stream/output/state_changes.proto
```diff
@@ -353,6 +353,16 @@ enum StateIdentifier {
...
+    /**
+    * A state identifier for scheduled transaction expiration.
+    */
+    STATE_ID_SCHEDULE_IDS_BY_EXPIRY = 29;
+
+    /**
+     * A state identifier for scheduled transaction deduplication.
+     */
+    STATE_ID_SCHEDULE_ID_BY_EQUALITY = 30;
+
```
jsync-swirlds, line 365: These should **replace** enumerations 15 and 16.  Block streams aren't published yet, so we can still make that change.

jsync-swirlds (commented):
A few updates we should make (apologies if I missed these in the original PR).
This _might_ be better held until the noted updates for `#16131` are ready as well.

block/stream/output/state_changes.proto
iwsimon: Those values are used in the code for migration to work. Should not replace them.

block/stream/output/state_changes.proto
iwsimon: Updated

block/stream/output/state_changes.proto
iwsimon: Still used as above

block/stream/output/state_changes.proto
iwsimon: Updated.

services/state/schedule/schedule.proto
iwsimon: Updated.

block/stream/output/state_changes.proto
jsync-swirlds: That's odd, this enumeration is specific to the block stream and _should not_ be used in state.

block/stream/output/state_changes.proto
jsync-swirlds: Similar, this is a block stream item, and shouldn't be part of any upgrade migration...

kimbor (approved):

block/stream/output/state_changes.proto
iwsimon: There are many code already use those values. The code won't even compile if we remove those values. This PR is to just sync the changes from hedera-services hapi directory to hedera-protobufs. We had our internal services meeting is to make code changes in the hedera-services and sync the changes to hedera-protobufs.

jsync-swirlds: I get that, it's more that we _should_ be able to still rewrite/update these enumerations until block streams is live.
Using block stream definitions in state is going to create all manner of problems eventually, because blocks streams need to change _independent_ of the consensus node.

iwsimon: That's a big topic to discuss with Schorodinger Team who works on Block Steams. This PR is just to sync my changes from hedera-services PR 16144.

jsync-swirlds: Sure, the proto can be changed in a later PR to clean up unnecessary items and adjust the numbering accordingly.
It would be helpful if you would initiate any necessary discussions when time permits.

iwsimon: I will mention this at our (Blockchain Wizards team) next daily this Thur. (Skipped tomorrow because of another meeting). Services has two teams on different daily.

PR #449, opened 2024-11-08
Sync 16495 protobuf change

iwsimon:
**Description**:

Sync protobuf changes from PR 16495.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/services/node_update.proto
+++ b/services/node_update.proto
@@ -161,27 +161,4 @@ message NodeUpdateTransactionBody {
      * If set, this field MUST NOT be set to an empty `KeyList`.
      */
     proto.Key admin_key = 8;
-
-    /**
-     * An ALT_BN128 elliptic curve public encryption key.<br/>
-     * This is controlled by the node operator and specific to this
-     * node's TSS operations.
-     * <p>
-     * The elliptic curve type MAY change in the future. For example,
-     * if the Ethereum ecosystem creates precompiles for BLS12_381,
-     * we may switch to that curve.<br/>
-     * This value SHALL be specified according to EIP-196 and EIP-197 standards.
-     * See [EIP-196](https://eips.ethereum.org/EIPS/eip-196#encoding) and
-     * (EIP-197](https://eips.ethereum.org/EIPS/eip-197#encoding)<br/>
-     * This field is OPTIONAL (that is, it can initially be null),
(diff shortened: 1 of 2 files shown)


services/node_update.proto
```diff
@@ -162,4 +161,27 @@ message NodeUpdateTransactionBody {
...
+     * but once set, it MUST NOT be null.<br/>
+     * If this field is set:
+     * <ul>
+     *   <li>This field MUST contain the bytes of a standard ALT_BN128 key value.</li>
+     *   <li>This key MUST only be used for node TSS operations.</li>
+     *   <li>This key MUST be unique in all future rosters.</li>
+     *   <li>This key MUST be used to encrypt all of this node's future TSS messages.</li>
+     * </ul>
+     */
+    bytes tss_encryption_key = 9;
```
jsync-swirlds, line 186: This seems possibly extraneous (differs from corresponding services PR).
Just checking that it's intended?

iwsimon: Yes, this was @mhess-swl 's change. He has not sync his change from services to here yet.

jsync-swirlds (approved):

services/node_update.proto
mhess-swl: FYI I haven't synced this portion yet because we don't yet have formal requirements from the cryptography team for this field

steven-sheehy, line 186: In which HIP is this defined? In what phase will this be delivered? Mirror node does not support this field. This change needs to go through TSC and techcom. It might be too late to update HIP-869 so might require a new HIP.

steven-sheehy (changes requested):

services/node_update.proto
iwsimon: This is not for HIP-869. @mhess-swl, can you please comment here why you made this change on develop branch?

jsync-swirlds: Might be better to remove this field from this PR, then.

jsync-swirlds (approved):

services/node_update.proto
iwsimon: Removed this section.

kimbor (approved):
lgtm!

steven-sheehy (approved):

PR #451, opened 2024-11-15
Removed incorrect specification for node id in receipt

jsync-swirlds:
Very small update to fix some documentation-only errors in the protocol buffer specification for Node update/delete and transaction receipt.

### DIFF
--- a/services/node_delete.proto
+++ b/services/node_delete.proto
@@ -36,9 +36,8 @@ option java_multiple_files = true;
  * - A deleted address book node SHALL be removed entirely from network state.
  * - A deleted address book node identifier SHALL NOT be reused.
  *
- * ### Record Stream Effects
- * Upon completion the "deleted" `node_id` SHALL be in the transaction
- * receipt.
+ * ### Block Stream Effects
+ * None.
  */
 message NodeDeleteTransactionBody {
     /**
--- a/services/node_update.proto
+++ b/services/node_update.proto
@@ -41,9 +41,8 @@ import "basic_types.proto";
  *   configuration during the next `freeze` transaction with the field
  *   `freeze_type` set to `PREPARE_UPGRADE`.
  *
- * ### Record Stream Effects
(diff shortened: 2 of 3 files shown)


services/node_delete.proto
```diff
@@ -36,9 +36,8 @@ option java_multiple_files = true;
  * - A deleted address book node SHALL be removed entirely from network state.
  * - A deleted address book node identifier SHALL NOT be reused.
  *
- * ### Record Stream Effects
- * Upon completion the "deleted" `node_id` SHALL be in the transaction
- * receipt.
+ * ### Block Stream Effects
+ * None.
```
Neeharika-Sompalli, line 40: There will be StateChanges recorded here

services/node_update.proto
```diff
@@ -41,9 +41,8 @@ import "basic_types.proto";
  *   configuration during the next `freeze` transaction with the field
  *   `freeze_type` set to `PREPARE_UPGRADE`.
  *
- * ### Record Stream Effects
- * Upon completion the `node_id` for the updated entry SHALL be in the
- * transaction receipt.
+ * ### Block Stream Effects
+ * None.
```
Neeharika-Sompalli, line 45: same here

services/node_delete.proto
jsync-swirlds: I haven't been specifically calling out state changes generally, as those are universal.

Do you find it valuable to state that explicitly?

Neeharika-Sompalli: I wasn't sure what are being added here. Since it says None, I wanted to mention there are some block stream effects.

jsync-swirlds: Nothing being added, just removing specification for items that are _not_ in the block stream (the node ID is only published in receipt for the create transaction).

Neeharika-Sompalli (approved):
LGTM!

PR #460, opened 2024-12-18
chore: Record File Items update based on conversion experience.

jsync-swirlds:
* Added `RecordFileSignature` message for record file signatures
* Modified sidecar file to use `SidecarFile` from streams v6.

### DIFF
--- a/block/stream/record_file_item.proto
+++ b/block/stream/record_file_item.proto
@@ -37,6 +37,7 @@ option java_multiple_files = true;
 
 import "basic_types.proto";
 import "timestamp.proto";
+import "sidecar_file.proto";
 
 /**
  * A Block Item for record files.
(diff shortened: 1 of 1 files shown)


jasperpotts (approved):

block/stream/record_file_item.proto
```diff
@@ -82,11 +83,29 @@ message RecordFileItem {
...
+     * This is the RSA signature of the node on the SHA384 hash of
+     * the record file
+     */
+    bytes signatures_bytes = 1;
+
+    /**
+     * A unique node identifier.<br/>
+     * This is the node id of the consensus node that created this signature.
+     */
+    int32 node_id = 2;
```
xin-hedera, line 110: `NodeAddress.nodeId` is of type int64, so for consistency, better to stick to the same type here?

jsync-swirlds: Not necessary, as node IDs are only 0-31 and once we turn on block streams, this will never be used thereafter.
So this value will only ever be much much smaller than int32.

xin-hedera: sounds good

xin-hedera (approved):
LGTM

PR #466, opened 2025-01-24
chore: Sync protobufs with hedera-services

derektriley:
**Description**:
This PR syncs all changes from hedera-services for v0.59.0-alpha.0

**Notes for reviewer**:

**Checklist**


block/stream/output/consensus_service.proto
```diff
@@ -75,7 +77,15 @@ message DeleteTopicOutput {}
  * The actual topic running hash SHALL be present in a `StateChanges` block
  * item, and is not duplicated here.
  */
-message SubmitMessageOutput {}
+message SubmitMessageOutput {
+
+    /**
+     * Custom fees assessed during a SubmitMessage.
+     * <p>
+     * These fees SHALL be present in the full transfer list for the transaction.
```
jsync-swirlds, line 85: Generally it is best to have fees in one location; here _or_ transfer_list, but not both, if we can manage that.

Are we committed to adding these to transfer_list, or could we perhaps keep them separate?

services/custom_fees.proto
```diff
@@ -336,3 +336,87 @@ message AssessedCustomFee {
      */
     repeated AccountID effective_payer_account_id = 4;
 }
+
+/**
+ * A custom fee definition for a consensus topic.
+ * <p>
```
jsync-swirlds, line 342: Nitpick:  The `<p>` is not needed on messages; only on fields.

block/stream/output/consensus_service.proto
xin-hedera: I think the fees here are for information only, they are not in addition to transfer_list or token_transfer_lists. they serve the purpose that it's easy for readers to understand in all the transfers as a result of the transaction's execution, what amount of which token is assessed custom fee.

jsync-swirlds: Actually, after re-reading the HIP discussions in the Block Stream HIP, we've been discussing removing `transfer_list` from the `TransactionResult` entirely.  Certainly any value in a Service Output message _should not_ be duplicated in the `transfer_list`.

We do not want to duplicate data in the block stream anywhere, as the cost of storing that duplication _forever_ exceeds any value offered to downstream software reading the stream.

services/state/entity/entity_counts.proto
```diff
@@ -0,0 +1,89 @@
...
+  uint64 num_schedules = 11;
+  /**
+   * The number of contract storage slots in the network.
+   */
+  uint64 num_contract_storage_slots = 12;
+  /**
+   * The number of contract bytecodes in the network.
+   */
+  uint64 num_contract_bytecodes = 13;
+}
```
jsync-swirlds, line 89: missing newline at EOF.

services/state/entity/entity_counts.proto
```diff
@@ -0,0 +1,89 @@
...
+   */
+  uint64 num_schedules = 11;
+  /**
+   * The number of contract storage slots in the network.
+   */
+  uint64 num_contract_storage_slots = 12;
+  /**
+   * The number of contract bytecodes in the network.
+   */
+  uint64 num_contract_bytecodes = 13;
```
jsync-swirlds, line 88: This doesn't meet the style guidelines.  Indentation is 4 spaces, there should be a blank line between entries, the descriptions are anemic, and the message definition offers no specification.
A reader would have no idea why this exists or what it's purpose is, and there is no indication of what must, or must not, be filled in; nor what limits exist.

services/state/entity/entity_counts.proto
```diff
@@ -0,0 +1,89 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This proto file contains primitive value messages.
+ * These are intended only for situations where the entire value to be stored in state is a single
+ * primitive.  These should never be used as components of another message; use the protobuf
+ * type instead.
+ */
```
jsync-swirlds, line 25: This reads like a random comment leftover from something else.

services/state/entity/entity_counts.proto
```diff
@@ -0,0 +1,89 @@
+syntax = "proto3";
```
jsync-swirlds, line 1: We're missing the file header comment here.

jsync-swirlds (dismissed):
The new state files are not acceptable.
* entity_counts.proto
* hints_types.proto
* history_types.proto
* tss_encryption_keys.proto

These files are all in incorrect style (missing newlines, 2 space indentation, etc...) and have half-baked descriptions instead of specification (a reversion to the bad practices that required a recent multi-month rework of the whole repo to fix).

All of those files need extensive work to properly specify the new  messages and fields, as well as fix the style guideline violations.

Those files should not have passed code review in the source repo.

services/state/entity/entity_counts.proto
Neeharika-Sompalli: Updated. Thanks @jsync-swirlds. 
It will be helpful to provide reviews on `hedera-services` PRs on source repo, so the sync PRs will not be blocked.

services/state/entity/entity_counts.proto
Neeharika-Sompalli: deleted. Thanks

services/state/entity/entity_counts.proto
Neeharika-Sompalli: Fixed. Thanks!

services/state/entity/entity_counts.proto
Neeharika-Sompalli: Fixed. Thanks!

services/transaction.proto
```diff
@@ -84,3 +151,462 @@ message Transaction {
...
+ * content of the transaction, while the `SignedTransaction` includes a
+ * signature map for signatures authenticating that byte array, and that is
+ * serialized and transmitted wrapped in a `Transaction` message.<br/>
+ * The bulk of this message is a `oneof` block which offers the option for
+ * any one of the transaction messages for the network.
+ * This message also includes several additional fields to specify
+ * various parameters required to process a transaction.
+ */
+message TransactionBody {
+    // <<<pbj.java_package = "com.hedera.hapi.node.transaction">>> This comment is special code for setting PBJ Compiler java package
```
jsync-swirlds, line 169: What happened here?

jsync-swirlds (commented):
This still isn't compiling (looks like a merge issue).
Can we fix that, at least, before merging this in?

jsync-swirlds:
> > The new state files are not acceptable.
> > 
> > * entity_counts.proto
> > * hints_types.proto
> > * history_types.proto
> > * tss_encryption_keys.proto
> > 
> > These files are all in incorrect style (missing newlines, 2 space indentation, etc...) and have half-baked descriptions instead of specification (a reversion to the bad practices that required a recent multi-month rework of the whole repo to fix).
> > All of those files need extensive work to properly specify the new messages and fields, as well as fix the style guideline violations.
> > Those files should not have passed code review in the source repo.
> 
> @jsync-swirlds Thanks for the review. These protobufs will not take in effect until 0.60. We can open PR to address documentation comments in `hedera-services` and port those changes back to `hedera-protobufs` (and re-tag if needed). This will unblock tagging for 0.59.
> 
> Also it will be helpful to have reviews on the PRs getting opened in `hedera-services` repo, so we will not need to block these sync PRs with multiple changes.

Understood.  We can merge this once it compiles correctly (doesn't at the moment).
It would definitely be good for the regular review process in services to catch the style and specification issue much earlier.

services/transaction.proto
derektriley: These changes were intentionally in https://github.com/hashgraph/hedera-services/pull/17402/files

services/transaction.proto
```diff
@@ -31,11 +31,77 @@ package proto;
  */
 
 option java_package = "com.hederahashgraph.api.proto.java";
-// <<<pbj.java_package = "com.hedera.hapi.node.base">>> This comment is special code for setting PBJ Compiler java package
```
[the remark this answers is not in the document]
derektriley: It compiles properly in hedera-services see https://github.com/hashgraph/hedera-services/pull/17402/

services/state/entity/entity_counts.proto
jsync-swirlds: Services makes enough changes that one person cannot likely keep up, also I don't usually know a PR exists if not added as a reviewer.
Generally, these things are something _everyone_ reviewing the code changes should be checking (style, compliance with HIP expectations, etc...).

services/transaction.proto
jsync-swirlds: This particular comment doesn't belong here, it belongs above.

services/transaction.proto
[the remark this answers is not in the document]
jsync-swirlds: pbj-integration-tests is failing when I run on this branch.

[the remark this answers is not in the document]
jsync-swirlds: I got it working, but it's definitely not _correct_.  There are multiple parsing errors and some important warnings that this isn't compiling the way it's probably intended.

It can merge if there is an important deadline, but we need to fix these issues _soon_.

Also, Services CI should _probably_ at least consider treating PBJ warnings as errors, much as java warnings are errors in the java compilation.

Neeharika-Sompalli (approved):

PR #467, opened 2025-01-31
fix: Adjust Block Stream to ensure a newly created account is identifiable

jsync-swirlds:
* Added a created account ID to the CreateAccountOutput.
* Added CreateAccountOutput to the TransactionOutput oneOf.

### DIFF
--- a/block/stream/output/transaction_output.proto
+++ b/block/stream/output/transaction_output.proto
@@ -57,9 +57,6 @@ import "stream/output/consensus_service.proto";
  *
  * <!--
  * Reserved definitions:
- * import "stream/output/consensus_service.proto";
- *    SubmitMessageOutput submit_message;
- *
  * import "stream/smart_contract_service.proto";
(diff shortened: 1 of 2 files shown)


block/stream/output/transaction_output.proto
```diff
@@ -161,5 +161,11 @@ message TransactionOutput {
          * Output from a consensus submit message transaction.
          */
         SubmitMessageOutput submit_message = 9;
+
+        /**
+         * Output from a transaction that includes the creation of a new
+         * CryptoService Account.
+         */
+        CreateAccountOutput account_create = 10;
     }
```
edwin-greene, line 170: Looks like `CreateAccountOutput` can be removed from the reserved messages list on line 83. `SubmitMessageOutput` may also be removed from line 61.

jsync-swirlds: Will do.

edwin-greene (approved):
Looks good to me

PR #469, opened 2025-02-26
Remove Previous Block Hash from Header

### LINKED ISSUE #468: Remove previous_block_hash from Block Stream BlockHeader
### Problem

Having  `previous_block_hash` in the Block Stream `BlockHeader` causes a timing problem that prevents optimal parallel thread pipelining in the consensus node. To get the block hash we have to have finished hashing the state of the previous block. This can take as long as it takes to execute all transactions in the next block. Because of that we do not want to hold up creating the block header until that hash is ready. 

### Solution

We already also have the hash of the previous block in `BlockProof.previous_block_root_hash` so we can just remove `BlockHeader.previous_block_hash`. This should then remove the synchronization requirement from services and allow more optimal parallel thread pipelining. The detailed comment for explaining the hash should be moved to `BlockProof.previous_block_root_hash` so we don't loose it.

thomas-swirlds-labs:
**Description**:
The previousBlockHash is no longer required in the Block Stream BlockHeader because it causes a timing problem that prevents optimal parallel thread pipelining in the consensus node and we do not want to hold up creating the block header until that hash is ready. It will continue to be available in the Block Proof however.

**Related issue(s)**:

Fixes #468

### DIFF
--- a/block/stream/block_proof.proto
+++ b/block/stream/block_proof.proto
@@ -98,12 +98,62 @@ message BlockProof {
     uint64 block = 1;
 
     /**
-     * A merkle root hash of the previous block.
+     * A block root hash for the previous block.
      * <p>
-     * This MUST contain a hash of the "block" merkle tree root for the
-     * previous block.<br/>
-     * The hash algorithm used MUST match the algorithm declared in the
-     * block header _for that block_.
+     * This value MUST match the block merkle tree root hash of the previous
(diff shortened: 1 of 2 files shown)


jsync-swirlds (approved):

mhess-swl (approved):

edwin-greene (approved):
Looks good to me

PR #471, opened 2025-03-14
Block content proof api

AlfredoG87:
**Description**:
Adds the necessary messages for a `BlockContentProof` Service:

**block_service.proto**
- message BlockContentProofRequest
- enum BlockContentProofResponseCode
- message BlockContentProofResponse
- service BlockContentProofService

**block_proof.proto**
- added message BlockContentProof
- removed field sibling_hashes from BlockProof  (I don't think it will ever be used.

**Related issue(s)**:

Fixes [#651](https://github.com/hiero-ledger/hiero-block-node/issues/651)

**Notes for reviewer**:

**Checklist**


block/stream/block_proof.proto
jsync-swirlds, line 1: Why are we changing this file?
We should _not_ change the BlockProof.

Block _content_ proof is completely separate.

block/block_service.proto
```diff
@@ -959,3 +960,131 @@ service BlockStreamService {
...
+/**
+ * A request for a proof of a block's content.
+ *
+ * This message is used to retrieve a proof that validates the content of a specific
+ * block item in a given block. The request MUST include the block number and MAY
+ * include either the full block item or its hash.
+ */
+ message BlockContentProofRequest {
+    /**
+     * The number of the block for which the block proof is requested.
```
jsync-swirlds, line 973: We are not providing block proofs, but block _content_ proofs, which are different.

block/block_service.proto
```diff
@@ -959,3 +960,131 @@ service BlockStreamService {
...
+     * <p>
+     * This value identifies the block that contains the block item to be verified.
```
jsync-swirlds, line 975: This needs to be specification, not description.

The [style guidelines](https://github.com/hashgraph/hedera-protobufs/blob/main/documents/Specification-Format-Style-Guidelines.md) for this codebase offer detailed guidance and examples.

block/block_service.proto
```diff
@@ -959,3 +960,131 @@ service BlockStreamService {
...
+        com.hedera.hapi.block.stream.BlockItem block_item = 1;
+
+        /**
+         * A hash of the block item for which the proof is requested.
+         * <p>
+         * This field allows the client to request a proof by providing a hash rather than
+         * the complete block item.
+         */
+        bytes block_item_hash = 2;
+    }
```
jsync-swirlds, line 997: We also need to know where, within the block, the item is.
This might be an index or a merkle path. I suspect the latter is more compatible with state proofs, but the former is easier to implement, both for client and service.

AlfredoG87: With the `block_item` we can calculate its hash, with the `block_item_hash` we can find the item index in either of the `input` or `output` trees.

Asking the index to the client, means they also would need to specify in what `sub merkleTree` the item is, and then the index.

Also, if they know the merkle path, they might as well calculate the proof themselves, merkle path is the output or result that the client is expecting from this request.

More info on Design Doc Review and Discussion (coming very soon)

jsync-swirlds: Index is just order, they only need the block to calculate that (1-n in stream order).
A Merkle Path does require calculating the merkle tree, which is why I noted it's easier to use an index.

I would expect more than just a merkle path for a block content proof. If that's all it requires, then it's sufficiently trivial I wonder why we don't just add block content proofs to the client SDK and they can be calculated locally (in any quantity needed) once the block is received.

AlfredoG87: Yes, I agree that given a block and a block_item, you can get the index. but again, that is not the index we care, we care about the index in their respective merkle tree (inputs or outputs) this creates many complications for users/clients, and they are paying us for the convenience of generating a proof, something that they could have done alone (if they have the whole block).

this is the BlockContentProof:
```
 message BlockContentProof {
    uint64 block = 1;
    com.hedera.hapi.block.stream.BlockItem block_item = 2;
    bytes block_item_hash = 3;
    repeated MerkleSiblingHash sibling_hashes = 4;
    bytes block_root_hash = 5;
    bytes block_signature = 6;
}
```

and everything this possible to calculate offline in any sdk given they have the block.
However they are paying for the convenience. that is why I advocate to keep the API as simple and flexible as possible.

block/stream/block_proof.proto
AlfredoG87: I will be changing things around files. expect a massive change. 
This was always expected being a draft

jsync-swirlds: Block Proof needs the sibling hash if it is a "cascade" proof, that is a proof of a prior block using the signature of a later block.
Consensus does not guarantee that every block will be signed directly, so it requires that capability.

jsync-swirlds: We also do not need to reuse the same object.
We can have a sibling hashes message _that is separate_ in the block node API without reusing the block stream.

It is very important to isolate API elements from the block stream.  The block stream is defined for consensus and affects the whole network.  Block Node API elements are specific to the block node, and have a much smaller radius of impact.

AlfredoG87: This is enlightening, Thank you!

I finally understand how "that" feature will work :)

block/block_service.proto
AlfredoG87: I am not sure about this... we need to discuss further with the rest of the team.
What kind of BlockItems could be identical? if the state update is different the block_item will be different, if not, then there should not be duplication, and even if there is, they should be idempotent, so proving either is valid for a user.

jsync-swirlds: Some of the outputs can be identical. I'm thinking of smart contract outputs in particular, as the same output can be created multiple times during a contract execution and produce identical output entries from child transactions.  The state change items, specifically, might not be repeated (or might even be elided entirely), but the contract outputs won't be.
There is no guarantee that a single state changes set will not include multiple identical changes.  A change set might include setting a value, executing a child transaction, then setting that value back, as an example.  That _might_ result in no value in the state changes, but it also might include both updates.

To my larger point, we _can_, always, uniquely specify the item to be proven with its index (which is cheap and easy for the caller to provide).
If we don't _require_ that index, then the user can leave it out and we can search through the block (although that should be a _more expensive_ call, as it requires more work). If the caller _does_ specify the index, however, then we have a faster seek and can differentiate between otherwise identical items.

AlfredoG87: Will try to do some research and take your points into consideration, maybe there is some value on requesting the `index` as an optional parameter.

block/block_proof_service.proto
```diff
@@ -0,0 +1,186 @@
...
+ * document are to be interpreted as described in
+ * [RFC2119](https://www.ietf.org/rfc/rfc2119) and clarified in
+ * [RFC8174](https://www.ietf.org/rfc/rfc8174).
+ */
+syntax = "proto3";
+
+package com.hedera.hapi.block;
+
+/*
+ * Copyright (C) 2025 Hedera Hashgraph, LLC
```
jsync-swirlds, line 17: Hmmmm, I thought we had already replaced these with SPDX in this repository.

That _was_ done in Services, so it looks like a sync is a bit overdue.

block/block_proof_service.proto
```diff
@@ -0,0 +1,186 @@
...
+import "stream/block_item.proto";
+import "stream/block_proof.proto";
+
+option java_package = "com.hedera.hapi.block.protoc";
+// <<<pbj.java_package = "com.hedera.hapi.block">>> This comment is special code for setting PBJ Compiler java package
+option java_multiple_files = true;
+
+/**
+ * A cryptographic proof of a block item within a block.
+ * <p>
```
jsync-swirlds, line 41: The `<p>` tag is _not_ needed for messages; just a blank line.

The tag is needed for _fields_ only because of a limitation in the way markdown processes tables.

block/block_proof_service.proto
```diff
@@ -0,0 +1,186 @@
...
+ * This message contains the block item to be verified, its Merkle tree sibling
+ * hashes, and a digital signature that attests to the authenticity of the proof.
+ */
```
jsync-swirlds, line 44: The second paragraph is description, not requirements, so it belongs _before_ the first paragraph break separated from the summary line by a line break `<br/>`.

Here is an example of what this should look like.
```suggestion
/**
 * A cryptographic proof of a block item within a block.<br/>
 * This message contains the block item to be verified, its Merkle tree
 * sibling hashes, and a digital signature that attests to the authenticity
 * of the proof.
 *
 * All fields of this message are REQUIRED.<br/>
 * The `block_item` field SHALL be identical to the `block_item`
 * requested.<br/>
 * The `block_signature` SHALL be the aggregate signature of the hiero
 * network that produced the block.<br/>
 * The recipient of this message SHOULD verify that the hash of the
 * `block_item`, when combined properly with the `sibling_hashes`,
 * produces a block hash that matches the value signed by the hiero
 * network, as represented by the `block_signature` field.
 */
```

block/block_proof_service.proto
```diff
@@ -0,0 +1,186 @@
...
+   * <p>
+   * This field holds the complete block item data that is subject to verification.
+   */
+  com.hedera.hapi.block.stream.BlockItem block_item = 1;
+
+  /**
+   * A list of sibling hashes forming the Merkle proof.
+   * <p>
+   * These hashes are used to reconstruct the Merkle root, ensuring the block item
+   * is included in the corresponding Merkle tree.
```
jsync-swirlds, line 57: This is description, not specifcation.
There are many examples throughout the block stream of proper specification.

block/block_proof_service.proto
```diff
@@ -0,0 +1,186 @@
...
+   */
+  uint64 block_number = 1;
+
+  /**
+   * One of several methods to specify the block item.
+   * <p>
+   * Exactly one of the following fields MUST be set:
+   * <br/>- `block_item`: The full block item data.
+   * <br/>- `block_item_hash`: The hash of the block item.
+   * <br/>- `block_item_index`: The index of the block item within the block.
```
jsync-swirlds, line 88: Why did you add line breaks; that prevents markdown from producing the correct output.
```suggestion
   * - `block_item` - The full block item data.
   * - `block_item_hash` - The hash of the block item.
   * - `block_item_index` - The index of the block item within the block.
```

Side note: The `oneof` enforces this, so it's not necessary to specify here.

block/block_proof_service.proto
AlfredoG87: Yes, agreed.

block/block_proof_service.proto
```diff
@@ -0,0 +1,186 @@
...
+ * This message contains the block item to be verified, its Merkle tree sibling
+ * hashes, and a digital signature that attests to the authenticity of the proof.
+ */
+message BlockContentProof {
+  /**
+   * The block item that is proved.
+   * <p>
+   * This field holds the complete block item data that is subject to verification.
+   */
+  com.hedera.hapi.block.stream.BlockItem block_item = 1;
```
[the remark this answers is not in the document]
AlfredoG87: I've set indentation to 4 spaces, is there a reason for this?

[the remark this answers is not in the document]
jsync-swirlds: Our standard for proto files is 4 spaces.

I believe it was chosen to be more similar to the consensus-node java code.

block/block_proof_service.proto
AlfredoG87: I am having a really hard time emulating your voice and tone for writing "specification".

To me, comments are for description, and is meant to be understood by a human.
Again, to me Specification is the actual API without comments, and is meant to be understood by a compiler/machine.

In my personal, humble opinion, is just a style of writing, and also very personally, I find harder to even understand and read that style of writing, comments should be concise and descriptive, and when overused, they obscure (which is exactly what happens to me with your style)

I acknowledge that this API already has this [guidelines](https://github.com/hashgraph/hedera-protobufs/blob/main/documents/Specification-Format-Style-Guidelines.md) as required, but too me is really hard to match this style of writing, "Formal", "Proper" english or simply a given style.

However I will have to ask your help to help me complete or change these comments if you want them your way, since I don't think I will be able to make the changes to your satisfaction.

Happy to help with anything else.

jsync-swirlds: The guidelines are there because these are not "description" comments.
These entire files, with all comments, are specification of an API, and the documentary comments are not "descriptions", but javadoc-style API specification that is automatically processed to produce published documentation.

I know you're probably used to writing description, even in javadoc, but that's just bad habits.

There are literally thousands of examples.

The simple approach is as follows:
Write a one-sentence "A <thing>" to tell the reader what this message or field is. Follow with a line break (`<br/>`).
Write, if _necessary_, a sentence or two describing this particular thing, this is the description you are used to writing.
Add a paragraph break (blank line for message, or `<p>` for fields).
Write one sentence for each requirement.  Use the RFC keywords as needed.

You probably won't ever match my style and tone, and I wouldn't expect you to, but at least you can absolutely match the pattern and guidelines.  You just need to choose to do so.

AlfredoG87: I will circle back on this again on monday... 👍 

Thank you for the short-simpler approach cheat sheet :)

AlfredoG87:
No longer needed, closing in favor of https://github.com/hiero-ledger/hiero-block-node/pull/891

